PR #79, opened 2015-02-25
first attempt, created pugin file

msrlab:
first try to create a face_detection plugin, much work stil to do
basic plugin works without error, but does not detect faces


msrlab:
Sorry, this pull request was not intended and due to my inexperience with github.

PR #92, opened 2015-02-28
Refactoring Plugin_List and fixing some typos

toshikurauchi:
Inheriting from list may not be a good idea, as it exposes methods that don't make sense. For instance it is possible to call append instead of add. I changed it to a list attribute and implemented __iter__ so the rest of the code is still the same.

Also, the method clean was skipping some plugins when removing the ones that were not alive. I just changed the implementation to use list comprehension.

Example of problem that may happen when removing elements inside for loop:
```
a = range(10)
for x in a:
     a.remove(x)
print a # [1, 3, 5, 7, 9]
```

### DIFF
--- a/pupil_src/shared_modules/plugin.py
+++ b/pupil_src/shared_modules/plugin.py
@@ -15,7 +15,7 @@ class Plugin(object):
     """docstring for Plugin
     plugin is a base class
     it has all interfaces that will be called
-    instances of this class ususally get added to a plugins list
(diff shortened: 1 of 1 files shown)


mkassner:
I merged this. It makes a lot of sense. Also the bug is a good catch! I did change the implementation a bit ti reduce calls as clean() gets called a lot. (see follow on commit.)

PR #141, opened 2015-04-21
Sending calibration info to outside app (Unity)

ryscet:
Hello,
I try sending the location and status of the marker to another app, then using it there to recreate the behavior of the marker. If this happens then the calibration will work on any screen the other app is displayed at. 
Best,
Richard

### DIFF
--- a/pupil_src/shared_modules/calibration_routines/screen_marker_calibration.py
+++ b/pupil_src/shared_modules/calibration_routines/screen_marker_calibration.py
@@ -330,7 +330,7 @@ class Screen_Marker_Calibration(Calibration_Plugin):
                 if self.active_site == 10:
                     self.stop()
                     return
-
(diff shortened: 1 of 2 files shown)


mkassner:
Looks like this part is done. Now you just need a ZMQ client app with (OpenGL) Canvas to draw the marker. 

A simple python client example can be found here:
https://github.com/pupil-labs/pupil-helpers/blob/master/pupil_server/zmq_filter_message.py

I m not familiar with Unity but there is a ZMQ binding that should work: https://github.com/zeromq/netmq


Hint: You don't have to care about coordinate system orientation. The calibration routine does not depend on it being consistent.

mkassner:
As this is very application specific I would say we don't merge it. If other users request this, we can add it to master.

PR #146, opened 2015-04-23
update file_methods to serialize using JSON rather than pickle

beasfall:
Hi - First, thanks for all your hard work on this project! I'm enjoying starting to dig in and get things working.

While exploring, I was trying to figure out what was stored in settings files. To do that more easily I switched them to serialize to plain-text JSON rather than pickled files. This also makes it easy to hand-edit them. I was curious if there's any setting data that isn't easily serialized to JSON like binary data?

notes:
* This change will continue to read pickled files, but will only write JSON files.
* I deleted `load_object` because it didn't seem to be used. `save_object` seems to be used in only one location; I converted it to JSON for consistency.
* the files names are updated to include a `.json` extension

I'm new to contributing to open-source projects, so feel free to give feedback on my pull request process :)

### DIFF
--- a/pupil_src/shared_modules/file_methods.py
+++ b/pupil_src/shared_modules/file_methods.py
@@ -10,30 +10,37 @@
 
 import cPickle as pickle
 import os 
+import json
 import logging
 logger = logging.getLogger(__name__)
 
 class Persistent_Dict(dict):
-	"""a dict class that uses pickle to save inself to file"""
+	"""a dict class that uses json to save itself to file. backward-compatible with pickled objects."""
 	def __init__(self, file_path):
 		super(Persistent_Dict, self).__init__()
 		self.file_path = os.path.expanduser(file_path)
-		try:
+
+		# pickled files
+		if os.path.exists(self.file_path):
 			with open(self.file_path,'rb') as fh:
-				try:
-					self.update(pickle.load(fh))
-				except: #KeyError,EOFError 
-					logger.warning("Session settings file '%s'could not be read. Will overwrite on exit."%self.file_path)
-		except IOError:
-			logger.debug("Session settings file '%s' not found. Will make new one on exit."%self.file_path)
+				self.update(pickle.load(fh))
+		elif os.path.exists(self.file_path+'.json'):
+			with open(self.file_path+'.json') as fh:
+				self.update(json.load(fh))
+		else:
+			logger.debug("Session settings file '{}' not found. Will make new one on exit.".format(self.file_path))
 		
 
 	def save(self):
-		d = {}
-		d.update(self)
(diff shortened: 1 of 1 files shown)


cpicanco:
Good for Sublime Text users, hum?

mkassner:
Hi, thanks for the pull request! The code looks good:
- clear code
- backwards compatible
- json is part of python std lib so we are not adding extra dependencies
- human readability is nice to have for debugging!

I found a few things that need to be solved:

We use this also to save marker cache data (100k+ small numpy arrays that describe detected markers ). its part of the offline marker detector which saves a file called `square_marker_cache`. 
I found that json cannot serialise numpy arrays. 

I also did some benchmarks and found that while both are about equally fast on reads json was 10 times slower on writes for these cache files. Reads on these files take up to .5 sec. Write of 5 sec are not acceptable.

I d also like to open the discussion to wether human readability is useful in this context:
- Its allows users to create settings that the software would not create (setting some var from type string to int). This can lead to unpredictable behaviour. 
- In its current form the settings file is really just a cache to remember the state of some aspects of the app. Where is the benefit of having it human readable?
- Opening a settings file in a python interactive session with 2 lines of code and changes the values there is not as easy as just using a text editor but still easy. 

What do you think?

beasfall:
Hi, thanks for the feedback!

Some more context: my goal was to have app configuration defaults in source control (in a personal repo) so I could consistently reproduce certain states. I was planning on adding an environment variable to optionally load the config from a different directory. In particular, I'd like to be able to externally track things like:

* pupil perimeter config
* region of interest
* camera settings like focus and exposure state

> In its current form the settings file is really just a cache to remember the state of some aspects of the app. Where is the benefit of having it human readable?

That's interesting. You're right that there is a distinction between an app state cache and user settings. But in its current form I was going to just let the window sizes and menu states come along for the ride :)

I just noticed that some settings I expected to see are not there: in particular camera settings like exposure and focus. Is there a technical or functional reason for this?

Regarding numpy arrays, I'll look into that pending discussion on the above.

Regarding performance, agree that json read/writes should not occur anywhere they may introduce a bottleneck. That said, the [`ujson`](https://pypi.python.org/pypi/ujson/) library would introduce a dependency but has performance that meets or beats cPickle.

mkassner:
hi,

Could you just do source control on the pickeld file? It would not be pretty but it does the job.

yes camera settings could and should be added. This should check the camera model before applying saved settings.  I will do that monday.

If we introduce another dependency, its added features must be convincing :-) Since we also need to bundle the software extra dependencies introduce a bigger bundle size and more maintenance.

beasfall:
True, the pickled files can be saved - though diffs aren't easily visible in change logs.

Agreed, better not to introduce the dependency in this case.

PR #148, opened 2015-05-12
Enabled seeking bar on retina displays

smurakami:
Hi, I'm using pupil-pro in my lab. It really helps me.

I'm using pupil on MacBook Pro with retina display. But it seems that the seeking bar of pupil-player doesn't work on retina display. So I did some fixation on `SeekBar` class.

I'm new to contributing to open-source projects, so feel free to give feedback on my pull request process :)


mkassner:
Merged!

PR #149, opened 2015-05-13
Naive implementation for Binocular_Gaze_Mapper

toshikurauchi:
I have implemented a simple binocular mapping that computes one mapping function for each eye and averages them to get the estimated gaze position.

Maybe you can use this one until you have a more robust implementation.

A few remarks:
- I added an extra column to the point cloud containing the eye id for each row
- I decided to add an extra flag in get_map_from_cloud to indicate whether it is binocular or not
- get_map_from_cloud works as before for monocular data, but in the binocular case it returns a list of mapping functions instead of a mapping function
- I didn't implement a way to tell the difference between points from each eye in show_calibration 

If you feel something should be implemented in another way, or you don't like some of the decisions I made, just let me know and I can try to fix it :)


mkassner:
Hi,

sorry for not replying any earlier. We actually did a very similiar implementation in the end. We will add another more sophisticated version in the near future.

PR #168, opened 2015-07-18
current frame feedback on dragging trim marks

cpicanco:
On Trim marks. Allow one to see what is the current frame when in dragging mode.

### DIFF
--- a/pupil_src/player/trim_marks.py
+++ b/pupil_src/player/trim_marks.py
@@ -10,6 +10,8 @@
 
 from OpenGL.GL import *
 from pyglui.cygl.utils import RGBA,draw_points,draw_polyline
+from pyglui.pyfontstash import fontstash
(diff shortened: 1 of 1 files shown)


mkassner:
this is a nice feature.  but we should use pyfontstash to render this info. I can help with implementation.

cpicanco:
Good! I tried to make the frame number follow the x cursor (similar to youtube video previews), but since some recordings I had have 10.000 frames I found a little bit difficult to adjust the size with the opencv stuff. But for me a fixed position is working, what do you think?

cpicanco:
I am having difficulties trying to implement the gl stuff to auto-resize the font and keep its aspect ratio. There is an example in the pupil code already that I was missing?

mkassner:
I think I have a good solution for this using the new logger visuals. I'll close this for now.

PR #169, opened 2015-07-18
typos (recording)

cpicanco:
Player On drop message typo.


mkassner:
Thanks for the typofix!

PR #173, opened 2015-07-19
Update main.py

cpicanco:
missing on close callback


mkassner:
We removed the callback because in player it serves no purpose. glfwsetwindowshouldclose is set by glfw automatically.

cpicanco:
It is crashing on ṕress the exit. So, I think the on_close is should be removed.

mkassner:
We don't need the callback. But there was still something missing. I just fixed it. Player should no longer crash when you close it: d5fdb03b68535dc0b1a636089da4b776a878cc2a

PR #175, opened 2015-07-21
Update offline_marker_detector.py

cpicanco:
I am assuming the maximum is 1.0


cpicanco:
Nothing about this?? I would need this to avoid drawing conflicts with #217

mkassner:
thats for the fix!

Issue #172, opened 2015-07-19
Pupil Player - help wanted to implement trim sections


cpicanco:
Hi everyone,

I am:
- ~~creating a new system plugin~~;
- updating trim marks to support sections;
- making a new separate normal plugin to create new sections (player/segmentation.py) and persistence;

Features I would like to have when finished:

**internals**
- [x] define section/sections `[[in-frame, out-frame], [.., ..] .. [.., ..]]`;
- [x] define focus `trim_marks.focus(section_index)` as int;
  - assign `sections` to the `trim_marks.sections` var and then focusing it;
- [x] one should edit a section by:
  1. after focus, assigning a frame_index to `trim_marks.in_mark` or to `trim_marks.out_mark` 
  2. assigning to `trim_marks.set(section)`; use `trim_marks.set` ~~to create a new~~ as a shortcut to edit a section;
- existing plugins that export stuff should be able to do it using the current trim sections;
  - [x] offline marker detector (export all sections by default);
  - [x] dispersion/duration fixation detector  (export all sections by default); 
  - export launcher;
    - [x] export all sections;
    - [x] should not conflict with the export range (export current selected section);
    - [ ] update progress stuff;
- persistence;
  - [ ] load cached sections somehow (trim_marks.py)
  - [x] load cached sections using the init_dict (segmentation.py);
  - [x] save current section;
  - [ ] the sections used to export Surfaces/Markers data may differ from the one used to export videos, so the ability to save multiple sections (sections set) and a selector with the available saved sections would be nice.

**gui**
- [x] trim marks plugin should render the sections on its bar;
- [x] one can focus a section by clicking on its centre;
- [x] one can edit a section by drag and drop its marks; 
- [ ] see the export progress on the seek bar would be cool;

---

I've chosen to not use the [trim_marks.py](https://github.com/cpicanco/pupil/blob/upstream_master/pupil_src/player/trim_marks.py) for creating sections. I am using the [segmentation.py](https://github.com/cpicanco/pupil/blob/upstream_master/pupil_src/player/segmentation.py#L103) for persistence, to create and export them to the trim_marks.py.
- [x] one can create a new event by pressing `v`; press `, (comma)` to undo; repeat a frame(event) is not allowed;
- [x] then one can auto trim/generate sections:
  - by chaining the events (start event as in, end event as out, middle events as in/out);
  - using them as in-out pairs (in(1, 3, 5...) out (2, 4, 6...), if odd the last single is ignored);
[...]

PR #177, opened 2015-07-22
Trim sections

cpicanco:
#172


cpicanco:
@mkassner , this patch allows to check the interface. I did not include the segmentation plugin in the main file though. I did not include yet changes in the export launcher, exporter, off line marker detector, dispertion/duration fixation detector that should be done in order to use the sections feature to export stuff (what I think is the main purpose of all this). Let me know if I should include theses changes all together in this branch . I am not sure how the batch exporter could fit into. Any comments will be appreciated.

cpicanco:
Anything about that @mkassner ?

PR #179, opened 2015-07-24
fix slice((start > 0), stop, step) heatmap bug

cpicanco:
when `section.start > 0`, frame_idx (L257) does not reflect the frame index producing buggy heatmaps;
The `gaze_on_srf_in_section` method is already using `frame_idx+=section.start`, so I just copy and paste it.


mkassner:
thanks!

PR #206, opened 2015-09-25
Update file_capture.py

elmorg:
Was getting an error when loading the player, and this solved it.


mkassner:
I thought I had already fixed that. Weird...

Thank you!

PR #217, opened 2015-10-13
Trim Sections

cpicanco:
#177
#172


cpicanco:
Hi @mkassner , please check this out when possible. It is ready. I have a question about the thread implementation of the current exporter.py. It would allow an "export all sections" button with just an external `for`, for example:

```
# pseudocode

for section in Sections:
   self.add_export()
```

Or would you not recommend such a thing?

cpicanco:
@mkassner , @willpatera  please, a feedback for this until the next milestone would be great!

Issue #231, opened 2015-10-19
Be consistent about the location and label of the close button on plugin


cpicanco:
I think they should read `Close` and be on top of the list.

mkassner:
AGREED!

if you create a PR I ll merge immediately :-).

PR #233, opened 2015-10-21
Updating close button locations

cpicanco:
#231 
First commit updates vis_circle.py

### DIFF
--- a/pupil_src/player/batch_exporter.py
+++ b/pupil_src/player/batch_exporter.py
@@ -74,6 +74,7 @@ class Batch_Exporter(Plugin):
 
     def _update_gui(self):
         self.menu.elements[:] = []
+        self.menu.append(ui.Button('Close',self.unset_alive))
         self.menu.append(ui.Text_Input('source_dir',self,label='Recording Source Directory',setter=self.set_src_dir))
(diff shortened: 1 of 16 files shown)


mkassner:
For the plugins with info text. Should we consider having the close button as the first ui element below the info text?

cpicanco:
Honestly, I think the close button is so basic that it should be included in the plugin bar.  Right now, when you open a plugin, the close button will always be at sight, because plugins will load in a "minimized" state. Closing and opening a plugin becomes more agile this way, with a fixed position and with the close button always at sight.

willpatera:
I think for now the close button should just be the first UI element, followed by info text, this way the pattern will always be the same. 

For the next milestone (v7.0), I agree with @cpicanco -- making the close button as a simple `X` as part of the plugin title bar would be nice (see sketch with 2 options below). 

pyglui_close_menu-01

Option 1 - keep collapse button. 
Option 2 - remove collapse button -- only click on title of plugin to collapse the menu. 

I will migrate this to [pyglui repo](https://github.com/pupil-labs/pyglui) -- as a feature request as it is most appropriate in that context.

cpicanco:
So, I think this can be merged! :smile:

mkassner:
done!

PR #243, opened 2015-11-11
Hardware is not open source

cpicanco:
It does not diminished the quality of the hardware.
You already explained why the frames are not open source in the wiki.
You did not explain why cameras are not.


mkassner:
Thank for this. Yes the first sentence is misleading. I revised it.

PR #245, opened 2015-11-16
allow easy to override surface initialization

cpicanco:
Moving surface initialization code to a separated method. I am having some problems trying to override the __init__ directly with multiple inheritance. This should make it easier to do it.


mkassner:
Seems good. I ll merge and rename the fn to 'load_surface_definitions_from_file'.

cpicanco:
Good!

PR #255, opened 2015-12-12
avoid binocular player error

cpicanco:
introduced at [193c463fde22b33687494d58080722a97f5688fd]


mkassner:
thanks for spotting this. I fixed it like so: 0932991a3899028656b5c72eeaf3397c894cbe38

cpicanco:
Much better :)

PR #257, opened 2015-12-17
license review

cpicanco:
[04eaff9] Deleting repetition
[a7ce196] remove CC BY-NC-SA and add LGPL3 name and link


mkassner:
Thanks for spotting this!

PR #275, opened 2015-12-27
happy 2016!

cpicanco:
Distributed under the terms of the GNU Lesser General Public License (LGPL v3.0) **License**.

 Distributed under the terms of the GNU Lesser General Public License (LGPL v3.0) **license**.

### DIFF
--- a/license/update_license_header.py
+++ b/license/update_license_header.py
@@ -3,7 +3,7 @@
  Pupil - eye tracking platform
  Copyright (C) 2012-2015  Pupil Labs
 
- Distributed under the terms of the GNU Lesser General Public License (LGPL v3.0) License.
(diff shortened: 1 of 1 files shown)


willpatera:
Hi @cpicanco thanks for spotting this. The second `License` should just be removed instead of lowercase. Writing license two times is redundant. The line should read:

```
Distributed under the terms of the GNU Lesser General Public License (LGPL v3.0).
```

We will also need to update the header dates Dec 31st/Jan 1st from `2012-2015` to `2012-2016`. 

Changes are only required on L17:L19 for the `license_txt` variable. Other headers will be updated when the script is run. 

If you make the changes prior to Dec 31st, we can merge the pull request. I will then run the script and update headers.

willpatera:
Thanks :smile:

PR #304, opened 2016-01-29
Update license.txt

cpicanco:
The update_license_headers script will not update the license.txt file. Right now I am just doing it manually.


willpatera:
Sorry for being so slow on this. We should actually update the script to include this file.

PR #316, opened 2016-02-06
heatmaps transparent background and colormap alternatives

cpicanco:
Blue backgroud is not good ~~(i.e., inventing data is not good)~~(transparency may be useful for better eye inspection). Now the blue reflects >0 in the color scale. Also adding some useful comments..
heatmap_screen_1454768954 47

### DIFF
--- a/pupil_src/shared_modules/offline_marker_detector.py
+++ b/pupil_src/shared_modules/offline_marker_detector.py
@@ -88,6 +88,8 @@ class Offline_Marker_Detector(Marker_Detector):
 
         self.img_shape = None
         self.img = None
+        
+        self.heatmap_colormap = "JET"
 
     def load_surface_definitions_from_file(self):
         self.surface_definitions = Persistent_Dict(os.path.join(self.g_pool.rec_dir,'surface_definitions'))
@@ -119,14 +121,24 @@ class Offline_Marker_Detector(Marker_Detector):
         if self.add_button:
             self.g_pool.quickbar.remove(self.add_button)
             self.add_button = None
-
+            
+    def set_mode(self, value):
+        self.mode = value
+        self.update_gui_markers()
+        
     def update_gui_markers(self):
         pass
         self.menu.elements[:] = []
         self.menu.append(ui.Button('Close',self.close))
         self.menu.append(ui.Info_Text('The offline marker tracker will look for markers in the entire video. By default it uses surfaces defined in capture. You can change and add more surfaces here.'))
         self.menu.append(ui.Info_Text('Please note: Unlike the real-time marker detector the offline marker detector works with a fixed min_marker_perimeter of 20.'))
-        self.menu.append(ui.Selector('mode',self,label='Mode',selection=["Show Markers and Frames","Show marker IDs", "Surface edit mode","Show Heatmaps","Show Metrics"] ))
+        self.menu.append(ui.Selector('mode',self,setter=self.set_mode,label='Mode',selection=["Show Markers and Frames","Show marker IDs", "Surface edit mode","Show Heatmaps","Show Metrics"] ))
+
+        if self.mode == 'Show Heatmaps':
+            self.menu.append(ui.Info_Text('Heatmap Settings'))
+            # self.menu.append(ui.Slider('heatmap_blur_gradation',self,min=0.01,step=0.01,max=1.0,label='Blur Gradation'))
+            self.menu.append(ui.Selector('heatmap_colormap',self,label='Color Map',selection=['AUTUMN','BONE', 'JET', 'WINTER', 'RAINBOW', 'OCEAN', 'SUMMER', 'SPRING', 'COOL', 'HSV', 'PINK', 'HOT']))
+
         self.menu.append(ui.Info_Text('To see heatmap or surface metrics visualizations, click (re)-calculate gaze distributions. Set "X size" and "Y size" for each surface to see heatmap visualizations.'))
         self.menu.append(ui.Button("(Re)-calculate gaze distributions", self.recalculate))
         self.menu.append(ui.Button("Export gaze and surface data", self.save_surface_statsics_to_file))
@@ -171,6 +183,7 @@ class Offline_Marker_Detector(Marker_Detector):
         # calc heatmaps
         for s in self.surfaces:
             if s.defined:
+                s.heatmap_colormap = "JET"
                 s.generate_heatmap(section)
 
         # calc distirbution accross all surfaces.
--- a/pupil_src/shared_modules/offline_reference_surface.py
+++ b/pupil_src/shared_modules/offline_reference_surface.py
@@ -271,24 +271,26 @@ class Offline_Reference_Surface(Reference_Surface):
                                             normed=False,
                                             weights=None)
 
-
+        # numpy.histogram2d does not follow the Cartesian convention
         hist = np.rot90(hist)
 
         #smoothing..
         hist = cv2.GaussianBlur(hist, (filter_size,filter_size),std_dev)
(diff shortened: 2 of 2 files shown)


willpatera:
Hi @cpicanco - good point. I think we could consider adding a few more color maps as options in the plugin that can be user selected in the GUI.

cpicanco:
I was actually exploring the colormaps and code some stuff. I think would be better to push it as another pull request. 

But let me ask, I was thinking about using the Mode Selector (Marker Detector) to show Mode options in demand. For example, selecting the `Show Heatmaps` option would show something like this after the Mode Selector:
```
if not self.mode == 'Show Heatmaps':
    self.menu.append(ui.Info_Text('Heatmap Configs'))
    self.menu.append(ui.Slider('heatmap_blur_gradation',self,min=0.01,step=0.01,max=1.0,label='Blur Gradation'))
    self.menu.append(ui.Selector('heatmap_colormap',self,label='Color Map',selection=['AUTUMN','BONE', 'JET', 'WINTER', 'RAINBOW', 'OCEAN', 'SUMMER', 'SPRING', 'COOL', 'HSV', 'PINK', 'HOT']))

```

cpicanco:
Well, just changed my mind, lets use this brach :laughing:

cpicanco:
For quick reference:
Some visualizations:
http://www.learnopencv.com/applycolormap-for-pseudocoloring-in-opencv-c-python/

OpenCv docs
http://docs.opencv.org/2.4/modules/contrib/doc/facerec/colormaps.html

cpicanco:
Well, matplotlib seems to be doing a great jog on the colormap topic:

http://matplotlib.org/style_changes.html

cpicanco:
Also, the colormaps themselves were released as CC0, public domain.

https://raw.githubusercontent.com/BIDS/colormap/master/colormaps.py

cpicanco:
An alternative using the matplotlib colormaps would be:
```
        # colormapping
        cv2_colormaps = ['magma', 'inferno', 'plasma', 'viridis']

        if self.heatmap_colormap == 'magma': 
            v = np.asarray(magma)
        elif self.heatmap_colormap == 'inferno': 
            v = np.asarray(inferno)
        elif self.heatmap_colormap == 'plasma': 
            v = np.asarray(plasma)
        elif self.heatmap_colormap == 'viridis': 
            v = np.asarray(viridis)

        v *= 255
        v = v.astype(np.uint8)

        RLUT = v[:,0]
        GLUT = v[:,1]
        BLUT = v[:,2] 
        
        self.heatmap = np.zeros((y_size,x_size,4), np.uint8)
        self.heatmap[:,:,0] = cv2.LUT(hist, BLUT)
        self.heatmap[:,:,1] = cv2.LUT(hist, GLUT)
        self.heatmap[:,:,2] = cv2.LUT(hist, RLUT)

        # alpha blend/transparency
        c_alpha = hist
        c_alpha[c_alpha>0] = 150

        self.heatmap[:, :, 3] = c_alpha
```

cpicanco:
Sample from viridis:
heatmap_screen_1455490763 56

cpicanco:
Correspondent from opencv jet
heatmap_screen_1455490763 56

cpicanco:
The mean difference is that jet is not "perceptually uniform" (lot of blue, for example)
http://bids.github.io/colormap/

cpicanco:
An option for Pupil is creating its own set of "perceptually uniform" colormaps... no standards in my sight, the matplotlib guys created a tool for this. Need to do some research, but I was thinking about fixing the actual jet (assuming cv implementation is not uniform as well) a perceptually_uniform_jet does not sounds like a dumb idea right now...

willpatera:
@cpicanco -- I would like to reopen this PR. 

I just re-read the article you linked on [perceptually uniform](http://bids.github.io/colormap/) color maps. And now fully realize the importance of making "perceptually uniform" color maps.

I don't think we should add all the possible color maps available in OpenCV. Instead we should add "perceptually uniform" maps in addition to `jet`. In my opinion, I would like to see the following color maps be available options to be selected in the GUI:
  - Jet
  - Magma
  - Inferno
  - Plasma
  - Viridis
  - Grayscale (although not explicitly listed in the article, I think an explicit grayscale map would be useful in many situations)

I also think we might want to consider making one of the "perceptually uniform" color  maps the default (either `viridis` or `magma`). What do you think @cpicanco @mkassner? 

@cpicanco - I am still not sure I understand why you want to the heatmap to have a "transparent" background. It appears to me that you are removing the lowest level of the heatmap gradient? In my understanding this does not seem like consistent with the logic of the heatmap. Can you please link to some concise reference literature for your proposed representation?

cpicanco:
Hi @willpatera , that is right, the lowest level (**IFF** frequency equals zero) now is not painted. I will not be able to link to a reference (right now), but let explain the rationale:

In short, no gaze response, no ink. So, when frequency equals zero, nothing should be painted. This approach should let one be more precise while inspecting eye movements dispersion. No data is lost, no data is invented. As the surface boundaries is equivalent to the heatmap boundaries, the bi-dimensional limits of the data is also preserved. Additionally, it allows higher contrasts when overlapping the heatmap image on top of the surface image.

willpatera:
Hi @cpicanco - thanks for the further explanation. 

I totally understand the argument for higher contrast when overlapping the heatmap on top of another image surface based on your example use-case (and on an aesthetic level I do like the transparent bkg heatmaps). 

However, I am still not sure I understand how correlating the lowest level of a heatmap where frequency is `0` with the lowest level of the heatmap spectrum is considered _inventing_ data.   

What about a switch that you can toggle on/off in the plugin to enable/disable this method of representation where `0` is transparent.

cpicanco:
Ok, I got it. Sorry about the *inventing*, It does not apply strictly considering the correlation between the colormap and the frequency (even though Pupil does not present the scale). But I am inclined to say that the transparent **is** the appropriate "color" for `0`, even for perceptually uniform colormaps, when ploting eye tracking data. Of course, I will need more than that to convince you.

I trigger does not sounds bad. Maybe there is some application for a colored background that I am not aware of.

papr:
How about, instead of simply cutting off frequency zero, weighting the transparency with the histogram? The result looks a bit "floaty" though: [Google Drive](https://drive.google.com/file/d/0B_Y6kGV4Sb5bMTd6NjFOTlQyMEk/view?usp=sharing)

cpicanco:
@papr , an elegant floaty dress/recording may result from data without flickering (missing surfaces), so with excelent marker tracking, and over flat real surfaces. What do you mean by weighting?

cpicanco:
Also, sorry about taking too long to update this branch. I still think it may be useful for better eye inspection.

papr:
```diff
--- a/pupil_src/shared_modules/offline_reference_surface.py
+++ b/pupil_src/shared_modules/offline_reference_surface.py
@@ -235,7 +235,7 @@ class Offline_Reference_Surface(Reference_Surface):
         c_map = cv2.applyColorMap(hist, cv2.COLORMAP_JET)

         self.heatmap[:,:,:3] = c_map
-        self.heatmap[:,:,3] = 125
+        self.heatmap[:,:,3] = hist
         self.heatmap_texture = Named_Texture()
         self.heatmap_texture.update_from_ndarray(self.heatmap)```

cpicanco:
@papr , for look and feel it may be interest. However, I am guessing it would brake perceptual uniformity.

papr:
Oh, it does with high probability. It was just an idea I wanted to try out.

mkassner:
Any ideas here? Should we proceed?

mkassner:
I ll close this.

PR #326, opened 2016-02-19
Pupil Server JSON format

zenithlight:
Hi, I needed to make use of the data that Pupil provides for my application. But the data format it gives is pretty cumbersome and nonstandard. I changed the Pupil Server so it returns data in JSON format instead, with the advantage that other ZMQ implementations can filter by topic (`pupil`, `gaze`, etc.). I also added the option to switch back to the old style of sending data if needed.

### DIFF
--- a/pupil_src/shared_modules/pupil_server.py
+++ b/pupil_src/shared_modules/pupil_server.py
@@ -8,18 +8,15 @@
 ----------------------------------------------------------------------------------~(*)
 '''
 
-from plugin import Plugin
-
-import numpy as np
-from pyglui import ui
 import zmq
+import json
+import numpy as np
 
-
+from plugin import Plugin
+from pyglui import ui
 
 import logging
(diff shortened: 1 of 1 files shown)


mkassner:
This is a great idea! I say we extend this even further: 

Take the whole events dict:
```
for key, value in events:
    send a multipart_msg with topic: key, payload value

this way we send all events (including custom and future ones) and the subscriber gets to decide whats useful.

I would also say we get rid of the exclude list, it was implemented for performance reasons but events is not a lot of data. 

Also lets not make this backward compatible rather update the docs.

mkassner:
I implemented this as told above: bf05ae03407b279525e4e05e1d1457c6aa1c9ca1
The helper scripts have also been upgraded: https://github.com/pupil-labs/pupil-helpers/commit/b6eb986bb66d219328d27fe2753257940c1f7cc1 (getting rid of 50 lines of parsing code :-) )

PR #364, opened 2016-04-18
Improvements to the Marker Tracker

mkassner:
This PR contains the merges of my work on the marker tracker.

1. It adds lens distortion to the transform model.
2. It adds the option to add and remove markers from reference surfaces.
3. It contains bugfixes of the existing code
4. It renames `marker_detector` to `surface tracker` I think thats a much more appropriate name.

@willpatera, please test and lets me know what you think.

### DIFF
--- a/pupil_src/capture/eye.py
+++ b/pupil_src/capture/eye.py
@@ -204,8 +204,8 @@ def eye(pupil_queue, timebase, pipe_to_world, is_alive_flag, user_dir, version, 
         g_pool.flip = session_settings.get('flip',False)
         g_pool.display_mode = session_settings.get('display_mode','camera_image')
         g_pool.display_mode_info_text = {'camera_image': "Raw eye camera image. This uses the least amount of CPU power",
-                                    'roi': "Click and drag on the blue circles to adjust the region of interest. The region should be a small as possible but big enough to capture to pupil in its movements",
-                                    'algorithm': "Algorithm display mode overlays a visualization of the pupil detection parameters on top of the eye video. Adjust parameters with in the Pupil Detection menu below."}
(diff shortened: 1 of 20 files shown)


willpatera:
@mkassner - I tested and the improvements are great. I think we should merge this PR. 

Some notes on future enhancements. Primarily for the add/remove markers feature:
  - add/remove markers
    - Global add/remove markers from dropdown menu. Current behavior requires no change. All surfaces visible in the scene can be manipulated.
    - Per-surface add/remove markers button. Expected/desired behavior: **only** the current surface should be editable. This is really necessary, because when working with multiple surfaces defined within a similar area, it can become challenging to determine which marker belongs to the surface and makes editing a challenge.
    - Visual feedback
      - Green dot for on (current behavior - requires no changes)
      - Red dot for off (current behavior - requires no changes) 
      - Other color dot (white? or no dot?) for not detected? (when the marker is not robustly detected the dot flashes between red and green, making it difficult for a user to understand the current mode of the marker. 
    - Surface boundary update
      - After removing or adding a marker I would expect the surface boundary to recalculate to the convex hull of the new set markers. Does this make sense?

mkassner:
@willpatera 

thanks for the review. 

I think your notes show that the current implmentation is misleading:

The marker add/remove option is always only for a single surface (all surfaces would not make sense.) This is why there is a button for this per each surface. If you use the global drop down it will simply use the last edited surface.

The marker dot will not flash between reg/green if the marker is not robustly detected. This will only happen if the marker's id changes because of false detection. (This condition we cannot check for.) Please make sure to not set min marker perimeter too low.

I don't think changing the surface boundary when adding markers is useful as the shape of the markers and the boundary of the surface are not related and the user should not make that false assumption anyways.

mkassner:
Alright I reworked the GUI to make this much clearer.

PR #366, opened 2016-04-29
Improve circle marker detector

mkassner:
This PR improves the circle marker detector we use for calibration:

- The detector now analyses the contour hierarchy to find clusters of nested contours.
- The detected clusters are filtered for ellipse contour fit. 
- The detected clusters are filtered against eccentricity.

With these changes the detector is more discriminative and allows us to use smaller and simpler markers like the one below:

screen shot 2016-04-29 at 11 39 52

The simple marker above printed with a diameter of 40mm can be used at distances of 2m with current Pupil hardware. The old marker is still supported and will continue working.


willpatera:
I just finished testing. It is very stable - I suggest that we merge it. It would be really nice to update the screen marker now that we have this :smile:

mkassner:
I made one performance improvement. Now I will merge it.

PR #368, opened 2016-05-03
Capture robustification

mkassner:
The PR improves capture robustness and failed capture feedback.

- The pupil video_backend now tries to restart the capture device after all attempts to read from the device have failed.
- The world process with fall back to a `fake capture` if all attempts to restart and recover communications with the camera have failed. This behaviour is better feedback to the user than a shutdown of the capture program.
- The world process will stop recording when the fallback to `fake capture` happens.

This PR requires an updated of [pyuvc](https://github.com/pupil-labs/pyuvc) to v0.6

### DIFF
--- a/pupil_src/capture/recorder.py
+++ b/pupil_src/capture/recorder.py
@@ -163,9 +163,10 @@ class Recorder(Plugin):
 
     def toggle(self, _=None):
         if self.running:
-            self.stop()
+            self.notify_all( {'subject':'should_stop_recording','network_propagate':True} )
         else:
-            self.start()
+            self.notify_all( {'subject':'should_start_recording','session_name':self.session_name,'network_propagate':True} )
+
 
     def on_notify(self,notification):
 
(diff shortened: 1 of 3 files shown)


mkassner:
@willpatera  please review and test.

cpicanco:
I have really appreciated this step. Not tested yet, but conceptually seems something to keep in mind... It was about time to handle exceptions like this and prevent as much as possible ugly crashes (closing everything without any prompt). Usually this kind of crash have occured as a "first impression" of a new user. We should care about first impressions. Excelent work @mkassner !

willpatera:
@mkassner -- this is excellent.

I tested this on my machine and it works very well. My quick test:
  - Started Pupil Capture with the headset plugged
  - Unplugged the USB while Pupil Capture was running and all streams successfully fallback to `Fake Capture`
  - I plug USB back into the computer and can successfully select and restart cameras from the GUI

Next step is to provide more informative (user friendly) feedback in `Fake Capture` similar to text feedback in Pupil Player launch screen. 

@cpicanco -- yes, a good first impression is very important.

Merge it 👍

mkassner:
Done

PR #372, opened 2016-05-05
Add Pupil Capture version and user system info to `world.log`

willpatera:
Log information about Pupil Capture version and user system in `world.log`

This information is essential for bug reporting and will save time for debugging community member issues if system information and pupil capture version can be communicated by simply sharing the world.log file.

This change will add two log messages in the `world.log`. See example output below:

```
World Process: 2016-05-05 14:46:53,931 - world - INFO - Pupil Capture Version: 0.7.5.18
World Process: 2016-05-05 14:46:53,932 - world - INFO - System Info - User: wrp, Platform: Darwin, Machine: Wills-MacBook-Air.local, Release: 15.4.0, Version: Darwin Kernel Version 15.4.0: Fri Feb 26 22:08:05 PST 2016; root:xnu-3248.40.184~3/RELEASE_X86_64
```

@mkasser - please have a look at this.

### DIFF
--- a/pupil_src/capture/recorder.py
+++ b/pupil_src/capture/recorder.py
@@ -17,6 +17,7 @@ from time import strftime,localtime,time,gmtime
 from shutil import copy2
 from audio import Audio_Input_Dict
 from file_methods import save_object
+from methods import get_system_info
 from av_writer import JPEG_Writer, AV_Writer, Audio_Capture
 from calibration_routines.camera_intrinsics_estimation import load_camera_calibration
 #logging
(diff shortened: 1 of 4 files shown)


willpatera:
Before merging this I want to move the `log_system_info` function to shared file like `methods.py` as this exact functionality is also used in `recorder.py`

willpatera:
@mkassner and all - please test and make revisions as you see necessary.

willpatera:
@mkassner - I reviewed your changes - looks good. I am writing some thoughts below for future consideration with some code snippets. I have not committed changes, because these changes would require more changes to ensure backwards compatibility in Pupil Player:

### `get_sys_info`
Consider returning a dict instead of a string so that we can use keys for headers in meta info files.

```python
def get_system_info():
    try:
        if platform.system() == "Windows":
            username = os.environ["USERNAME"]
            sysname, nodename, release, version, machine, _ = platform.uname()
        else:
            username = getpass.getuser()
            sysname, nodename, release, version, machine = os.uname()
    except Exception as e:
        logger.error(e)
        username = 'unknown'
        sysname, nodename, release, version, machine = sys.platform,'unknown','unknown','unknown','unknown'

    return {"User":username, "Platform":sysname, "Machine":nodename, "Release":release, "Version":version}
```

### `info.csv` 
I am aware that a change in `info.csv` format would require an update to Player's version check, but maybe worth considering a more standard `csv` format using commas instead of tab delimiters with a clearly defined header (again this is not critical to performance, but perhaps worth considering for the future).  

In recorder this would look like this:

```python
with open(self.meta_info_path, 'wb') as csvfile:
    csv_writer = csv.writer(csvfile, delimiter=",")
    
    system_info = get_system_info()
    header = ['Duration Time','World Camera Frames','World Camera Resolution','Capture Software Version']+system_info.keys()
    values = [self.get_rec_time_str(),
                str(self.frame_count),
                str(self.g_pool.capture.frame_size[0])+"x"+str(self.g_pool.capture.frame_size[1]),
                self.g_pool.version]+system_info.values()
    
    csv_writer.writerow(header)
    csv_writer.writerow(values)
```

PR #373, opened 2016-05-06
Log History - GUI window that displays last 50 logger messages

willpatera:
Proposing `log_history` plugin for Pupil Capture motivated by UX testing feedback from @nathakits, @DouglasMackay and conversations with other members in the Pupil community about the brevity of the log messages.

screen shot 2016-05-06 at 11 12 45

This is a simple (convenience) plugin that shows the last 50 log messages as `Info_Text` within a separate GUI window (`Scrolling_Menu`). The most recent log message appears at the top. This plugin is motivated by feedback from UX tests and provides a way to review log messages in the case that they were not seen when they were rendered by `Log_Display`. Full logs are written to files, so this just provides a glimpse into the past instead of a full record.

Other enhancements to this plugin could include a timestamp for the log messages so that log messages are prepended with the timestamp -- but I was trying to keep this plugin very simple. 

Let me know what you think.

### DIFF
--- a/pupil_src/capture/world.py
+++ b/pupil_src/capture/world.py
@@ -85,6 +85,7 @@ def world(pupil_queue,timebase,lauchner_pipe,eye_pipes,eyes_are_alive,user_dir,v
     from log_display import Log_Display
     from annotations import Annotation_Capture
     from pupil_remote import Pupil_Remote
+    from log_history import Log_History
     # create logger for the context of this function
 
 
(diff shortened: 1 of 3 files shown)


cpicanco:
A button to open the folder containing the full logs seems a better choice than the textual description (with no path at all).

mkassner:
Alright I made some more changes to the plugin implementation.

mkassner:
@cpicanco  the button open is a bit overkill but a path to the log files is useful. I just added that.

willpatera:
@mkassner - I will check out your changes tomorrow -- thanks!

willpatera:
@cpicanco - thanks for the suggestion but I agree that a button to open system files is a bit too much for this little helper plugin. @mkassner - I tested your refactor - looks good 😸  -- I would vote to merge it.

mkassner:
I'll add one more improvement later than we can merge.

mkassner:
@willpatera  @cpicanco  With b671a74 the log now also shows previous messages and the formatting is unified.

willpatera:
@mkassner - these changes are very useful.

cpicanco:
Very useful, see the previous messages is imperative!

PR #377, opened 2016-05-09
Ipc logging

mkassner:
This PR improves the logging experience:

- It gives all processes a name so the logging record processName attribute become meaningful to the user.
- It uses a custom handler in the eye process to send pickled log records to the world process via ZMQ PUB/SUB on `tcp://127.0.0.1:502020`
- It adds a log collection thread in the world process that received the eye log records and handles them accordingly.
- All logs now show up in the world process.
- All logs are now written to `capture.log` instead of `world/eye/0/1.log`.

### DIFF
--- a/pupil_src/capture/eye.py
+++ b/pupil_src/capture/eye.py
@@ -37,31 +37,31 @@ def eye(pupil_queue, timebase, pipe_to_world, is_alive_flag, user_dir, version, 
     """
     is_alive = Is_Alive_Manager(is_alive_flag)
     with is_alive:
+        #logging setup: We stream all log records to the world process.
         import logging
         import zmq
-        from zmq.log.handlers import PUBHandler
         ctx = zmq.Context()
         pub = ctx.socket(zmq.PUB)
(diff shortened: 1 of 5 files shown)


mkassner:
@willpatera  please review and test.

willpatera:
@mkassner - this is nice. 

Some questions:
  - Can you explain `sleep(.1)` in `stop_eye_process` - this enables the eye process a bit of time to receive the exit message and exit gracefully?
  - Can you explain the specification of this address and port?

mkassner:
On `sleep(.1)`:
The sleep is inside a loop that is conditioned on the status of the eye process. It will stay in the loop until the eye process is shut down. We need to waste some time in the loop, thus the sleep. 
Wy waiting? If the eye process where to shut down after the world process not all log messages would be received.

On Address and port:
127.0.0.1 is localhost, nothing special. The port is a high arbitrary number `502020` thats not in this list: https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers

mkassner:
I just tested on Ubuntu 14.04 works there as well.

PR #385, opened 2016-05-27
ZMQ IPC

mkassner:
This PR replaces multiprocessing pipes and queues with a unified IPC backbone. 

From now on all exchanged data (pupil data, control messages, log messages) flow though this IPC backbone. All messaging has been unified and is now sent using the ZMQ PUBSUB pattern.

## Motivation
If Pupil Capture is very similar to Pupil Service we can share a lot of codebase between the two apps. The IPC backbone is very handy for Pupil Capture and will be crucial for Pupil Service.

## Message Format
Pupil messages are defined as multi-frame zmq messages with frame 0 being the topic and frame 1 being a json encoded dict that carries the payload. More message payload type may be added in the future. zmq_tools.py implements the format and augments zmq.Sockets accordingly. The message topics we are using as of now are `logging` , `pupil` , `gaze` , `notify` , `delayed_notify` . Message topics are often refined with postfixes like: `pupil.0` , `logging.error` . Pupil notifications have auto-created topics: `notify.[notification subject]` .Users may add more topics as long as their names do NOT start with the names mentioned. (As this would break the filter logic.)

## IPC Backbone
The launcher process now hosts a zmq proxy thread that handles all message routing, in this PR we refer to this as the IPC backbone.  World and eyes connect to it at the `ipc_sub_url` to receive and `url_pub_url` to send messages. The launcher process now also handles delay_notifications and logging.

## Tapping into the IPC from outside
Any external actor (on the same box or remote) with knowledge of the `ipc_sub_url` and `ipc_pub_url` (shown in the world GUI) can now tap into the IPC backbone and subscribe to all (or parts) of the message stream. It can also send messages that will be received by all subscribers in the app.

## Deprecation of Pupil Server:
Since this IPC backbone is basically Pupil Server on steroids we have no need for it any more in Pupil Capture and it has been removed from the app.

## Gaze Mapping:
All gaze mappers have been restructured to handle temporal correlation of pupil data explicitly. Two base classes for mono and binocular mappers have been added. This commit has not completed the work needed for binocular gaze mappers.

## Plugins:
Will continue working as before. No Plugin API changes except delayed notifications. See `plugin.py` for details.
Plugin notifications are now published on the IPC Backbone.

## New Dependencies:
`pip install msgpack-python`

@willpatera I still have to test this on Linux. Lets discuss the changes.


mkassner:
Alright. I tested it on Linux. It works.

I also bundled it and the bundle works on my dev machine as well.

mkassner:
I made more changes.  I think it makes most sense to use pupil remote as an intermediary for talking to Pupil Capture and all subscriptions to the Messages Bus (IPC Backbone) for read only. 

zmq_tools.py has a small script at the bottom that shows how to talk to pupil capture remotely.

PR #394, opened 2016-06-15
[pupil_remote] Return more precise float representation of timestamp

Tobias-Fischer:
Hi,
I noticed that the timestamp returned by `pupil_remote` is cut off after 2 decimal points due to the conversion to `string`. Using this change, 5 decimal points are used instead.

### DIFF
--- a/pupil_src/shared_modules/pupil_remote.py
+++ b/pupil_src/shared_modules/pupil_remote.py
@@ -145,7 +145,7 @@ class Pupil_Remote(Plugin):
                 self.g_pool.timebase.value = raw_time-target
                 response = 'Timesync successful.'
         elif msg[0] == 't':
-            response = str(self.g_pool.capture.get_timestamp())
+            response = '%.5f' % self.g_pool.capture.get_timestamp()
         else:
             response = 'Unknown command.'
         socket.send(response)


papr:
This is important if one wants to use this feature for time synchronisation.

Tobias-Fischer:
@papr Yes, sorry, I should have mentioned that I changed this exactly because of this reason.

papr:
I made a small test.
```
import zmq
ctx = zmq.Context()
ip = 'localhost'
port = 50020
requester.connect('tcp://%s:%s'%(ip,port))

for t in range(5):
    requester.send('t')
    t = requester.recv()
    print 'time:',t
```

Results:
```
time: 85450.767003
time: 85450.7672712
time: 85450.7674539
time: 85450.767634
time: 85450.7678496
```

My conclusion is that `str()` cuts off trailing zeros. An other small test confirmed that. `print str(0.0100), str(0.0101)` results in `0.01 0.0101`. My guess is that in your case, the timestamp had trailing zeros after the second digit. Therefore, I will close this pull request.

Tobias-Fischer:
Indeed, my bad. Sorry for that, it's all working fine. Thanks for checking!

Tobias-Fischer:
Actually, the problem was somewhere else. See my new PR for more info :).

PR #397, opened 2016-06-15
Pupil remote timestamp fixed

Tobias-Fischer:
Improved version of https://github.com/pupil-labs/pupil/pull/394
The issue is in the string representation of a float. For the example by @papr, all works fine in the current implementation. However, ROS is using timestamps which are much bigger than the ones typically used in pupil (by `zmq`).
The following scripts shows where the current implementation fails:
```
import zmq
ctx = zmq.Context()
ip = 'localhost'
port = 50020
requester.connect('tcp://%s:%s'%(ip,port))

for t in range(5):
    requester.send('t')
    t = requester.recv()
    print 'time:',t

requester.send('T 1466007067.2524871826')
resp = requester.recv()
print 'resp:',resp

for t in range(5):
    requester.send('t')
    t = requester.recv()
    print 'time:',t
```

Result:
```
tobias@deepthought:~/catkin_ws/src/pupil_ros/scripts$ python test.py 
time: 1466007403.55
time: 1466007403.55
time: 1466007403.55
time: 1466007403.55
time: 1466007403.55
resp: Timesync successful.
time: 1466007067.25
time: 1466007067.25
time: 1466007067.25
time: 1466007067.25
time: 1466007067.25
```

In short, `1466007067.2524871826` is cut off by the `str()` method as it is quite a large float.

This PR fixes this. After applying the patch, the result is as follows:
```
tobias@deepthought:~/catkin_ws/src/pupil_ros/scripts$ python test.py 
time: 1222020.140440977
time: 1222020.141337291
time: 1222020.141556358
time: 1222020.141666636
time: 1222020.141808348
resp: Timesync successful.
time: 1466007067.2527401
time: 1466007067.2528384
time: 1466007067.253045
time: 1466007067.2531295
time: 1466007067.2532117
```

Thanks,
Tobi


mkassner:
perfect. Thanks for this!

Tobias-Fischer:
Thanks for merging :)

gmierz:
Hello,

I have been looking around for the precision of the time stamps and I am wondering what the precision is now? I am posting here since this seems to be the most recent change to the time stamps that I could find.

Thanks!

cpicanco:
The epoch (time to start couting) is the startup of the computer (if you or PUPIL did not change it somehow). The precision floating point is in seconds. The accuracy is ~ -5ms (a litle bit of delay). The granularity I don't know the exact value, but on linux it depends on clock_monotonic function. Please someone correct me if necessary.

mkassner:
I have to correct @cpicanco here:

- The `unit` is seconds but the `precision` is much higher 
- The accuracy is that of clock monotonic on all platforms 
- We now correctly convert the float64 to a string and thus get the full precision of float64
- The added delay through transport is better that 5ms more like 1ms for wifi and in the usec range for localhost and wired

gmierz:
Great, thanks for the quick answers! One other thing is that float64 has 15 significant digits which means that the timestamp is guaranteed to be precise to at least ~10 usec.

cpicanco:
- `unit` : seconds
- `precision`: full float64 precision, i.e. `10 μs`.
- `accuracy`:
   - if wifi it is  `~1 ms`
   - if wired or localhost it is in the range of `μs`
- `granularity`:
   - it is machine specific (depends on clock_monotonic on linux). It is constrained by the processor cycles and software.
   - In my machine (2 GHz processor), the result comes from `clock_gettime(CLOCK_MONOTONIC, &time_record)` function on linux. This function delivers a record with `nanosec`, 1 GHz, granularity. Then PUPIL software does some math and delivers a float64.

gmierz:
Excellent breakdown! Can it be placed in the timestamp section of the pupil data wiki so that it will be easier to find for other people?

cpicanco:
Fell free to update the wiki @gmierz , it is open!

gmierz:
Great, will do. Thanks again for the help!

Also, based on pupil capture the max configurable sampling rate allowed is 120Hz. Not sure if it can go higher than that but based on the latency of the camera/processing, 120Hz makes sense.

cpicanco:
I suggest that you update the following section:

https://github.com/pupil-labs/pupil/wiki/Data-Format#timestamps

cpicanco:
I would like to note that PUPIL Capture is assychroneous. So, what is the max sampling rate allowed by the software itself (not only the cameras)? Consider the following problem:

PUPIL already supports two 120hz cameras running in parallel. Would PUPIL support an additional custom device running at 30hz?

gmierz:
In my experiments, one thing that I noticed was that you are not guaranteed to have a 120Hz sampling rate if you select 120Hz in the pupil capture settings. If you are running both eye readers, they seem to max out at just under 60Hz interestingly enough. If you put all the data together though, there is a 120Hz sampling rate.

If we bring in a third camera, based on my calculations, it would bring the maximum sampling rate up to just under 60Hz based on the latency alone and assuming they all have the same latency. I'm not sure if the calculation that I used here is correct though.

PR #402, opened 2016-06-21
zmq_tools extentions

papr:
These are changes I made during rewriting `Pupil Interface`.

Introduces `ZMQ_Actor` which unifies actor initialization and shutdown. Each actor has its own zmq socket. All actors inherit `ZMQ_Actor`. `Msg_Dispatcher` inherits `Msg_Streamer` but uses a `PUSH` instead of a `PUB` socket.

`Requester` acts as `Pupil Remote` client.

@mkassner Please review.


mkassner:
I think simplicity out-weights the gains of inheritance in this case.

Issue #414, opened 2016-06-29
Keep OS X alive - pupil capture


hmvandermeulen:
Would it be possible to prevent OS X from going to sleep mode while recording instead of changing the OS power settings manually for the recording? It could have saved me from data loss unfortunately. Other software has the ability to keep the system alive during certain processes (eg. full screen video)

papr:
Hey @hmvandermeulen, you can try `Caffeine` (https://itunes.apple.com/en/app/caffeine/id411246225?mt=12) as a temporal solution.

mkassner:
@hmvandermeulen we will add a way of doing this. In the meantime I recommend changin power settings. BTW: if you need to close the lid use: https://github.com/integralpro/nosleep/releases

papr:
With 8857db976, we support disabling idle sleep on Mac OS X 10.11 and higher.

PR #415, opened 2016-06-29
Problem: OS X might start sleeping during recording

papr:
See feature request pupil-labs/pupil#414

Solution: Use built-in `caffeinate` utility to prevent this behaviour.

State:
- [x] Support OS X
- [x] Start/stop `caffeinate`
- [ ] Support other platforms


mkassner:
I don't think we should use `caffeinate `. Neither is it required to write a plugin. 

1) If we want Pupil Capture to prevent idle sleep, it should prevent it when started not just during recording. 
2) The proper way of doing this for mac is documented here: https://developer.apple.com/library/mac/qa/qa1340/_index.html

papr:
Use of `caffeinate` as discussed. @mkassner Please review.

papr:
Ready to merge.

papr:
Fixes #415

PR #416, opened 2016-06-29
Problem: Pupil Sync is too complex

papr:
Solution: Split discovery and time sync features into different plugins.

- [x] Discovery implementation
- [ ] Discovery UI
- [ ] Time Sync implementation
- [ ] Time Sync UI

**`discovery.py` documentation:**
- Creates `Pyre` node with following headers:
    - "pub_address"
    - "sub_address"
    - "app_type"
- Joins "pupil-discovery" group by default.
- Shouts all notifications with the key "discovery.shout"
    - Notifications will be serialized with msgpack
- Whispers all notifications with the key "discovery.whisper"
    - Notifications will be serialized with msgpack
- Publishes all discovered notifications to the IPC backbone
    - Drops notifications which cannot be unpacked with msgpack
        or do not have "subject" key
    - Adds "discovery.peer" key with dictionary containing information:
        - "uuid": peer uuid
        - "name": peer name
        - "arrival_timestamp": time of arrival
        - "type": "SHOUT" or "WHISPER"
        - ["group": source group] if "type" equals "SHOUT"

### DIFF
--- a/deployment/bundle.sh
+++ b/deployment/bundle.sh
@@ -1,0 +1,15 @@
+rm *.dmg
+rm *.deb
+cd deploy_capture
+./bundle.sh
+mv *.deb ../
+mv *.dmg ../
+cd ../deploy_service
+./bundle.sh
+mv *.deb ../
+mv *.dmg ../
+cd ../deploy_player
+./bundle.sh
+mv *.deb ../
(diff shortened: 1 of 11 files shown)


papr:
Fixed in #434

PR #419, opened 2016-07-01
all modification of my work is described on the Pupil community

IroneP:
Hi,

I'm clement Picq and during my internship in LIRIS laboratory in Lyon (France), I've worked with Pupil.
The main goal of our project was to be able to interact with an open environment simply by looking at it. To do so, we use your marker detection to located the exact 3D camera pos in the environment and thus, interact with marker simply by looking at them during a few frames. We have implemented some basics augmented reality methods using OpenGL. We also create a different way to calibrate the camera using chessboard (it was more easy for us to do so and it seems that chessboard calibration give better results) and with that, create method to rectify lens distortion (we also add method to correct gaze position and markers points in an undistord image). We try to keep all our change in different files. We create 3 differents files : 
-world_interaction : The main file we have create to do our work
-chessboard calibration : the calibration with the chessboard
-drawing_utils : for all function that use OpenGL. We allso have to re-write the 'apply_gl_texture' method because the texture was applied in z=0 and thus, enabled us to draw in 3D.
-some change in method.py and square_marker_detect : Note that using all the point detected with the marker and not only the 4 angle will give the possibility to use SolvePNPRansac method to detect extrinsic parameters with better precision.

You can contact me at : clement.picq@etu.univ-lyon1.fr   if you have any questions or comments to me.

Have a nice day,
Clement

### DIFF
--- a/pupil_src/capture/world.py
+++ b/pupil_src/capture/world.py
@@ -84,6 +84,7 @@ def world(pupil_queue,timebase,lauchner_pipe,eye_pipes,eyes_are_alive,user_dir,v
     from marker_detector import Marker_Detector
     from log_display import Log_Display
     from annotations import Annotation_Capture
+    from world_interaction import World_Interaction
     # create logger for the context of this function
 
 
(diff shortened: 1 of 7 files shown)


mkassner:
Thank you for this contribution. I really appreciate the work! However, this will not apply to the broader use cases. So we will close this PR.

Issue #421, opened 2016-07-01
Can't start pupilcapture: Exception: ZMQ connection failed


PolynomialDivision:
Hi,
when I'm trying to start pupil (from source or deb) this error appears:
    
    Traceback (most recent call last):
      File "main.py", line 258, in <module>
        launcher()
      File "main.py", line 201, in launcher
        cmd_sub = zmq_tools.Msg_Receiver(zmq_ctx,ipc_sub_url,topics=topics )
      File "/home/nick/pupil/pupil_src/shared_modules/zmq_tools.py", line 48, in __init__
        raise Exception("ZMQ connection failed")
    Exception: ZMQ connection failed

So I looked at the status from the Msg_Receiver and I got this
    {'endpoint': 'tcp://127.0.0.1:5722', 'event': 2, 'value': 115}
    {'endpoint': 'tcp://127.0.0.1:5722', 'event': 128, 'value': 12}

So 128 is

    zmq.EVENT_CLOSED

this means

    The ZMQ_EVENT_CLOSED event triggers when a connection's underlying descriptor has been closed.

Adding the line

     if status['event'] == zmq.EVENT_CONNECTED or status['event'] == zmq.EVENT_CLOSED:

allows me to start pupil capture with this error:

      File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
        self.run()
      File "/usr/lib/python2.7/threading.py", line 754, in run
        self.__target(*self.__args, **self.__kwargs)
      File "main.py", line 110, in ipc_backbone
        xsub.bind(in_url)
      File "zmq/backend/cython/socket.pyx", line 487, in zmq.backend.cython.socket.Socket.bind (zmq/backend/cython/socket.c:5156)
      File "zmq/backend/cython/checkrc.pxd", line 25, in zmq.backend.cython.checkrc._check_rc (zmq/backend/cython/socket.c:7535)
        raise ZMQError(errno)
    ZMQError: Address already in use

PolynomialDivision:
I found the problem.
A fix for me was to assign the ports manually:
For example my setup was:

    ipc_sub_port = 5001
    ipc_pub_port = 5002
    ipc_push_port = 5003

mkassner:
The error in zmq tools is because for some reason random port assignment is failing. I will try a different approach to solve this problem.

mkassner:
I can reproduce this problem at all. Could you give some info on your setup? Hardware/ OS?

mkassner:
[...]

PR #422, opened 2016-07-01
Fix the random port initialization #421

PolynomialDivision:
It causes erros when you try to do this in the launcher.
Issue: #421


mkassner:
This will create problems for the bundle. We will need to find a different solution.

mkassner:
what happens if you just add a second of sleep after closing the test socket? Does that help?

mkassner:
fixed in a different way.

PR #428, opened 2016-07-04
Add ability to send individual refdata.

PolynomialDivision:
Instead of sending a list in the end you can send inidividual refdata and lists.


mkassner:
Could you give a bit of background for this PR? 

During implementation we had considered this option but thought that sending a single datum inside a list would be the effective equivalent while being more precise from an implementation and structure standpoint.

PolynomialDivision:
There is no good background for this PR. ^^
I was sending a single datum and wondered why I had problems.
So I looked in the code and thought it would be a nice feature.

But I think you are right when you say that only sending lists is more structured.
So I will just close this PR and rewrite my code.

PR #441, opened 2016-07-18
Fix notify_all

PolynomialDivision:
There is no self.notify_all in finish_calibration.
So use g_pool.active_calibration_plugin.notify_all instead.


mkassner:
Thanks!

PR #442, opened 2016-07-18
Add new gaze_mapper for hmd

PolynomialDivision:
The standard gaze_mapper hmd calibration is using is not that exact.
The gaze point jumps around.

To solve this the new gaze_mapper uses the confidence as an score for both eyes and then calculates weighted average for x,y of the gaze.

### DIFF
--- a/pupil_src/shared_modules/calibration_routines/__init__.py
+++ b/pupil_src/shared_modules/calibration_routines/__init__.py
@@ -17,7 +17,7 @@ from camera_intrinsics_estimation import Camera_Intrinsics_Estimation
 from adjust_calibration import Adjust_Calibration
 from accuracy_test import Accuracy_Test
 from hmd_calibration import HMD_Calibration
-from gaze_mappers import Dummy_Gaze_Mapper, Monocular_Gaze_Mapper, Binocular_Gaze_Mapper,Vector_Gaze_Mapper,Binocular_Vector_Gaze_Mapper,Dual_Monocular_Gaze_Mapper
+from gaze_mappers import Dummy_Gaze_Mapper, Monocular_Gaze_Mapper, Binocular_Gaze_Mapper,Vector_Gaze_Mapper,Binocular_Vector_Gaze_Mapper,Dual_Monocular_Gaze_Mapper, Binocular_HMD_Scored_Gaze_Mapper
 
 calibration_plugins =  [Screen_Marker_Calibration,
                         Manual_Marker_Calibration,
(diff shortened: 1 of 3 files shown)


mkassner:
Hi,

I think there is a misunderstanding as to how the normal hmd gaze mapper works. It maps each eye individually to a separate space (think HMD left and HMD right side display.)  You should use the `id` field to filter the right side. 

If you want to create a binocular hmd mapper you can do this though averaging but I would recommend doing something similar to what is done in the binocular mapper instead.

Even better would be using a 3d model but this gets more complicated.

PolynomialDivision:
Thanks for your feedback! :)

I think I want to create a binocular hmd mapper... That's why I inherited Binocular_Gaze_Mapper_Base from the gaze mapper class.

I tried several thinks. For example just call finish_calibration with the pupil data and the ref data. A Problem is to give the "HMD-Size" to this function.
It always said that the data is to bad to create a model...

I would like to use a 3d model but I don't know if I can handle this. Could you give me some advice?

Do you mean when u tell me that you would recommend something similiar the the binocular mapper that making a weighted average is not a good idea? So I should just use a normal average?
What I do is that I score the x and y component with the confidence...

So instead of

```
gaze_point_eye0 = self.map_fns[p0['id']](p0['norm_pos'])
score_eye0 = p0['confidence']
gaze_point_eye1 = self.map_fns[p1['id']](p1['norm_pos'])
score_eye1 = p1['confidence']
score_total = score_eye0 + score_eye1
gaze_point = (gaze_point_eye0[0] * score_eye0 + gaze_point_eye1[0] * score_eye1)/score_total , (gaze_point_eye0[1] * score_eye0 + gaze_point_eye1[1] * score_eye1)/score_total
```
u would just recommend to do this

```
gaze_point_eye0 = self.map_fn_fallback[0](p0['norm_pos'])
gaze_point_eye1 = self.map_fn_fallback[1](p1['norm_pos'])
gaze_point = (gaze_point_eye0[0] + gaze_point_eye1[0])/2. , (gaze_point_eye0[1] + gaze_point_eye1[1])/2.
```

mkassner:
hi,

sorry for the late reply. Is this still an issue? We can implement a binocular hmd gaze mapper using the 3d model.

PolynomialDivision:
I would like to see a binocular hmd gaze mapper using the 3d modelin in a new version. :)

I'm having some accuracy problems with my implementation of the mapper. I think this would solve my problems.

mkassner:
We are working on this right now. See https://github.com/pupil-labs/pupil/tree/hmd_calibration_3d

PR #454, opened 2016-08-29
Fix initial action menu label. Increase pyglui requirement to v0.9

papr:


mkassner:
Perfect. @papr  please merge at will.

PR #456, opened 2016-08-30
`should_call_ui_functions` implementation.

### LINKED ISSUE #455: How to run Capture in the background
How can I run capture process in the background? I don't want to edit any of the capture settings, all I want is to run pupil capture and start recording. While recording, I don't want the user nor myself to edit any of the capture settings - Basically, I just want to run and record.

If it's not possible to just run it in the background, how can I stop displaying the frames stream for both world and eye cameras?

papr:
Only call UI functions if window is *visible and not iconified*. Make Plugin `gl_display()` call part of UI functions. Fixes #455

### DIFF
--- a/pupil_src/capture/eye.py
+++ b/pupil_src/capture/eye.py
@@ -130,10 +130,14 @@ def eye(timebase, is_alive_flag, ipc_pub_url, ipc_sub_url,ipc_push_url, user_dir
         g_pool.app = 'capture'
         g_pool.timebase = timebase
 
+        def should_call_ui_functions(window):
+            # GLFW_VISIBLE == 0x00020004
+            visible = glfw.glfwGetWindowAttrib(window, 0x00020004)
+            return visible and not g_pool.iconified
 
         # Callback functions
         def on_resize(window,w, h):
-            if not g_pool.iconified:
+            if should_call_ui_functions(window):
                 active_window = glfw.glfwGetCurrentContext()
                 glfw.glfwMakeContextCurrent(window)
                 g_pool.gui.update_window(w,h)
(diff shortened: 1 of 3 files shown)


pupil_src/capture/eye.py
```diff
@@ -130,10 +130,14 @@ def eye(timebase, is_alive_flag, ipc_pub_url, ipc_sub_url,ipc_push_url, user_dir
         g_pool.app = 'capture'
         g_pool.timebase = timebase
 
+        def should_call_ui_functions(window):
+            # GLFW_VISIBLE == 0x00020004
+            visible = glfw.glfwGetWindowAttrib(window, 0x00020004)
```
mkassner: I think the constant is exposed like so: glfw.VISIBLE

pupil_src/capture/eye.py
```diff
@@ -130,10 +130,14 @@ def eye(timebase, is_alive_flag, ipc_pub_url, ipc_sub_url,ipc_push_url, user_dir
         g_pool.app = 'capture'
         g_pool.timebase = timebase
 
+        def should_call_ui_functions(window):
```
mkassner: just call this fn is_visible(window)

pupil_src/capture/world.py
```diff
@@ -149,9 +149,14 @@ def get_timestamp():
     default_plugins = [('Log_Display',{}),('Dummy_Gaze_Mapper',{}),('Display_Recent_Gaze',{}), ('Screen_Marker_Calibration',{}),('Recorder',{}),('Pupil_Remote',{})]
 
 
+    def should_call_ui_functions(window):
```
mkassner: import from gl_utils

pupil_src/capture/eye.py
papr: `AttributeError: 'module' object has no attribute 'VISIBLE'`

pupil_src/capture/world.py
papr: gl_utils do not have access to `g_pool.iconified`. Should we implement a poll (call `glfwGetWindowAttrib `) on the `iconified` attribute as well?

papr: We should do that anyway, because if we minimise the main window it would set `g_pool.iconified = True` which affects `eye` behaviour. This should not be the case.

mkassner:
@papr : If this works on your machine. Please merge.

PR #458, opened 2016-09-02
Update fixation_detector.py docs

cpicanco:
Please, see diffs.

### DIFF
--- a/pupil_src/shared_modules/fixation_detector.py
+++ b/pupil_src/shared_modules/fixation_detector.py
@@ -30,8 +30,14 @@ class Fixation_Detector(Plugin):
 
 class Fixation_Detector_Dispersion_Duration(Fixation_Detector):
     '''
-    This plugin classifies fixations and saccades by measuring dispersion and duration of gaze points
+    This plugin classifies fixations and saccades by measuring dispersion and duration of gaze points.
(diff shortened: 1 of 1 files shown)


papr:
The old fixation detector did not calculate the dispersion correctly. It used the distance to the mean  instead of maximal distance between all gaze points.

#460 solves this problem but is in need of references as well.

cpicanco:
Salvucci & Goldberg (2000) reported 5 algorithms for fixation detection. When you say "not correctly", did you have some of them in mind?

papr:
The source code currently says that a dispersion threshold fixation detector is implemented. It has two parameters: maximal dispersion and minimal duration.

The problem is that the actual dispersion calculation is not correct. For performance reasons, the old code calculates the dispersion as the distance between the current gaze point and the mean of all gaze points currently belonging to the fixation. Not only does the mean shift during time, it is also not the actual dispersion.

To calculate the dispersion correctly the distance of all gaze points to each other needs to be calculated. The biggest value is the dispersion.

cpicanco:
Ok, I got it. I will check if the algorithm you just described (the correct one) fits into one of the Salvucci & Goldberg (2000)'s categories. Will do this in the next week.

PR #460, opened 2016-09-09
Pupil Angle Fixation Detector

papr:
Correct duration-dispersion fixation detector which uses the pupil normals' pair-wise angles as distance function for the dispersion calculation.

### DIFF
--- a/pupil_src/player/vis_circle.py
+++ b/pupil_src/player/vis_circle.py
@@ -42,10 +42,21 @@ class Vis_Circle(Plugin):
         else:
             thickness = self.thickness
 
-        pts = [denormalize(pt['norm_pos'],frame.img.shape[:-1][::-1],flip_y=True) for pt in events.get('gaze_positions',[])]
-        for pt in pts:
-            fixations = self.g_pool.fixations_by_frame[frame.index]
-            c = (0,0,1,.2) if fixations else (self.b, self.g, self.r, self.a)
+        pts = [(
+            denormalize(pt['norm_pos'],frame.img.shape[:-1][::-1],flip_y=True),
+            pt['timestamp'])
+            for pt in events.get('gaze_positions',[])
+        ]
+        fixations = self.g_pool.fixations_by_frame[frame.index]
+        fixation_timeranges = [
+            (f['timestamp'], f['timestamp']+f['duration'])
+            for f in fixations
+        ]
+        for pt, ts in pts:
+            c = (self.b, self.g, self.r, self.a)
+            for timerange in fixation_timeranges:
+                if timerange[0] <= ts <= timerange[1]:
+                    c = (0,0,1,.2)
             transparent_circle(frame.img, pt, radius=self.radius, color=c, thickness=thickness)
 
     def init_gui(self):
--- a/pupil_src/player/vis_circle.py
+++ b/pupil_src/player/vis_circle.py
@@ -42,22 +42,9 @@ class Vis_Circle(Plugin):
         else:
             thickness = self.thickness
 
-        pts = [(
-            denormalize(pt['norm_pos'],frame.img.shape[:-1][::-1],flip_y=True),
-            pt['timestamp'])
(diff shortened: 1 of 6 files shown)


papr:
Offline and online fixation detection. #463

papr:
This needs documentation in the wiki.

pupil_src/player/vis_circle.py
```diff
@@ -42,9 +42,22 @@ def update(self,frame,events):
         else:
             thickness = self.thickness
 
-        pts = [denormalize(pt['norm_pos'],frame.img.shape[:-1][::-1],flip_y=True) for pt in events.get('gaze_positions',[])]
-        for pt in pts:
-            transparent_circle(frame.img, pt, radius=self.radius, color=(self.b, self.g, self.r, self.a), thickness=thickness)
+        pts = [(
```
mkassner: is it a good idea to put this into the vis_circle plugin. Maybe we do this as a separate visualizer?

pupil_src/player/vis_circle.py
```diff
@@ -42,9 +42,22 @@ def update(self,frame,events):
...
+            denormalize(pt['norm_pos'],frame.img.shape[:-1][::-1],flip_y=True),
+            pt['timestamp'])
+            for pt in events.get('gaze_positions',[])
+        ]
```
mkassner: could you add a comment explaining what these lines do on a high level?

pupil_src/player/vis_circle.py
papr: It was actually a debugging feature which appeared useful afterwards. The idea is that pupil positions which are part of a fixation are drawn red. If there are no fixations (or the fixation detector is disabled) the points are drawn as before, i.e. it does not assume a working fixation detector.

Therefore I would recommend to add this feature to the other visualizers as well instead of creating a separate one.

pupil_src/player/vis_circle.py
papr: Yes. Fixations are grouped by frame. The initial implementation would check if there is a fixation for the current frame and if yes render all pupil positions red. This led to the problem where saccades would be rendered red as well just because an fixation ended at the beginning of the frame.

Therefore it is needed to filter the pupil positions that are part of an fixations from those that are not — per frame.

pupil_src/player/vis_circle.py
mkassner: I think this should simply be a separate plugin.  Similar to viz_circle but that it works in fixations. Otherwise this one becomes kind of magic and non-intuitive.

pupil_src/player/vis_circle.py
mkassner: ok thanks!

pupil_src/shared_modules/fixation_detector.py
```diff
@@ -275,3 +320,464 @@ def get_init_dict(self):
...
+            detect_win = self.detection_windows[eye_id]
+            # add recent gaze positions one by one to make sure we do not overstep
+            for datum in recent[eye_id]:
+                detect_win.remove_surplus_data(self.min_duration)
+                detect_win.append(datum)
+                if (detect_win.duration >= self.min_duration and
+                    detect_win.max_distance_deg <= self.max_dispersion):
+                    new_fixation = detect_win.pop_fixation()
+                    events['fixations'].append(new_fixation)
+                    logger.debug('Fixation eye %i: %fdeg'%(new_fixation['eye_id'],new_fixation['dispersion']))
```
mkassner: how often does this happen? If more that 1hz please comment this line out.

papr: This is only displayed once a fixation is found. This means this happens at most each `min_duration ` seconds. This will probably be more than 1Hz.

pupil_src/player/vis_circle.py
papr: I will remove all the `vis_circle ` changes in favour of @snleu's visualization.

PR #466, opened 2016-09-16
ElSe Detector integration

papr:


mkassner:
For evaluation only. NO COMMERCIAL USE ALLOWED. NOT LGPLv3 compatible.

PR #467, opened 2016-09-16
Updated all instances of Thumb in Capture and Player

snleu:
player thumbs changed to icons, capture and plugin thumbs updated syntax and kept as letter


pupil_src/capture/world.py
```diff
@@ -73,7 +73,7 @@ def world(timebase,eyes_are_alive,ipc_pub_url,ipc_sub_url,ipc_push_url,user_dir,
 
     #check versions for our own depedencies as they are fast-changing
     from pyglui import __version__ as pyglui_version
-    assert pyglui_version >= '0.9'
+    assert pyglui_version >= '0.10'
```
mkassner: We decided to change versioning to api_change.revison so please change this to '1.0'

mkassner:
all perfect. Except a small change. I made the relevant comment in the commit. @snleu  please change and I will merge.

pupil_src/capture/world.py
papr: Same change in `pupil_src/player/main.py`. :)

PR #471, opened 2016-09-21
Vis_Fixation plugin and added it to Player

snleu:


pupil_src/player/vis_fixation.py
```diff
@@ -0,0 +1,92 @@
...
+from plugin import Plugin
+import numpy as np
+import cv2
+
+# TODO: Import pyglui
+from pyglui import ui
+
+from methods import denormalize
+
+from fixation_detector import Fixation_Detector
```
papr: This import is not used in the code below.

pupil_src/player/vis_fixation.py
```diff
@@ -0,0 +1,92 @@
...
+ License details are in the file license.txt, distributed as part of this software.
+----------------------------------------------------------------------------------~(*)
+'''
+
+from player_methods import transparent_circle
+from plugin import Plugin
+import numpy as np
+import cv2
+
+# TODO: Import pyglui
```
papr: TODO is resolved. See L17

snleu: this was a copied from vis_circle, i'll delete it

pupil_src/player/vis_fixation.py
snleu: correct forgot I added this, i'll delete this as well

mkassner:
Commit looks good! Please rebase onto current master so I can merge.

PR #480, opened 2016-09-27
improving pupil remote connection behaviour

mkassner:


pupil_src/shared_modules/pupil_remote.py
```diff
@@ -67,45 +68,106 @@ class Pupil_Remote(Plugin):
...
+            self.notify_all({'subject':'service_process.should_stop'})
+            return
+
+        #for capture we try to bind to a arbitrary port on the first external interface
         else:
-            logger.error(response)
-            if self.g_pool.app == 'service':
-                audio.say("Error: Port already in use.")
-                self.notify_all({'subject':'service_process.should_stop'})
+            self.thread_pipe.send_multipart(('Bind',"tcp://*:*"))
```
papr: Why do we bind to * instead of self.host?

pupil_src/shared_modules/pupil_remote.py
```diff
@@ -67,45 +68,106 @@ class Pupil_Remote(Plugin):
...
             sleep(.1)
 
     def init_gui(self):
+        self.menu = ui.Growing_Menu('Pupil Remote')
+        self.g_pool.sidebar.append(self.menu)
+        self.update_menu()
+
+    def update_menu(self):
+
+        self.menu.elements[:] = []
```
papr: Why not `del array[:]`?

pupil_src/shared_modules/pupil_remote.py
```diff
@@ -67,45 +68,106 @@ class Pupil_Remote(Plugin):
...
+                self.update_menu()
+
+
         help_str = 'Pupil Remote using ZeroMQ REQ REP scheme.'
-        self.menu = ui.Growing_Menu('Pupil Remote')
         self.menu.append(ui.Button('Close',close))
         self.menu.append(ui.Info_Text(help_str))
-        self.menu.append(ui.Text_Input('address',self,setter=self.start_server,label='Address'))
-        self.g_pool.sidebar.append(self.menu)
+        self.menu.append(ui.Switch('use_primary_interface',self,setter=set_iface,label="use primary network interface"))
```
papr: I think "use" should be capitalized here. Generally, this is something we should have a look at.

pupil_src/shared_modules/pupil_remote.py
mkassner: your suggestion is syntactically cleaner. I ll do that.

pupil_src/shared_modules/pupil_remote.py
mkassner: Agreed.

pupil_src/shared_modules/pupil_remote.py
mkassner: that would fail unrecoverably during startup of capture if the host ip changes.

papr: Right. I did not think about that.

papr: I am not sure if it is worth the complexity, but we could look up the name of the network interface used if `not use_primary_interface` and try to fallback to its ip address as first fallback and only if that does not work fallback to `*`.

papr: Generally it would be nice to be able to enumerate network interfaces instead of having the user type ip addresses. Initialisation would look somehow like this:

```python
success = False
if interface_available(last_interface):
  try:
    success = bind(last_interface, last_port)
  except:
    success = bind(last_interface, random_port)
if not success:
  try:
    success = bind(primary_interface, last_port)
  except:
    success = bind(primary_interface, random_port)
if not success:
  logger.error()
  return
```

mkassner: I think thats adding to much complexity.

papr: I thought so too. Just wanted to put the idea out there. :)

mkassner:
I made fixes according to suggestions. OK to merge?

papr:
Looks good but one thing: on `not self.use_primary_interface`, the address field could use information on the address format. It expects `host:port` but below it displays `tcp://host:port`. This could lead to user using the second format instead of the first one.

mkassner:
I think its clear enough from the filled in info.

PR #487, opened 2016-09-29
Export annotations to csv file in annotations_player plugin

snleu:


pupil_src/shared_modules/annotations.py
```diff
@@ -188,6 +190,38 @@ def fire_annotation(self,annotation_label):
...
+
+
+    def export_annotations(self,export_range,export_dir):
+
+        if not self.annotations:
+            logger.warning('No annotations in this recording nothing to export')
+            return
+
+        annotations_in_section = chain(*self.annotations_by_frame[export_range])
+        annotations_in_section = dict([(a['index'],a) for a in annotations_in_section]).values() #remove dublicates
```
papr: One could use a dict-comprehension in here. That could be copied to the old code as well.

mkassner: @papr  can you post a line what what you have in mind. We are only removing dublicates and need the output as a list anyways.

papr: `annotations_in_section = {a['index']:a for a in annotations_in_section}.values()`

mkassner: nice!

mkassner:
@snleu  once the comment by @papr  is implemented we can merge this.

PR #497, opened 2016-10-11
Marker Tracking

papr:


pupil_src/shared_modules/square_marker_detect.py
```diff
@@ -301,24 +319,197 @@ def detect_markers_robust(gray_img,grid_size,prev_markers,min_marker_perimeter=4
...
+    """docstring for MarkerTracker"""
+
+    vert_slc = slice(0,8)
+    loc_conf_idx = 8
+    id_slc = slice(9,15)
+
+    def __init__(self, detect_func=detect_markers):
+        super(MarkerTracker, self).__init__()
+        self.detect_in_frame = detect_func
+        self.history = []
```
mkassner: can we call this state ?

papr: I do not think that `state` is the best name for it. `History` is not either. Especially if we have observations in a similar structure.

papr:
Last commit includes forward projection using last velocity

willpatera:
## General feedback
There are some really nice changes introduced in this PR. The ability to "pause" tracking and edit surfaces is really key for interactively editing surfaces within Pupil Capture. Additionally the visual feedback of "learning" percentage for each surface is useful. 

## Performance comparison 
I would like to see some performance comparisons between this method and the previous method. Qualitatively it looks like surfaces markers and surfaces are more stable, but would be useful to see a formal comparison.

## Future changes required for `Opencv3` compatibility
For future reference and compatibility with `opencv3` the following changes will be required:
  - `calcOpticalFlowPyrLK` requires another argument `next_pts` which should have the same shape as `prev_pts` see docs [here](http://docs.opencv.org/3.0-last-rst/modules/video/doc/motion_analysis_and_object_tracking.html?highlight=calcopticalflowpyrlk)
  - `findContours` returns 3 values instead of 2. Returns `image`,`contours`, and `hierarchy`

mkassner:
merged PR #505 instead.

Issue #246, opened 2015-11-16
pupil player plugin list is getting very big


cpicanco:
Bottom plugins are being hidden outside of the user clickable area. Also, it is not possible to roll down the dropdown list as usual.

![screenshot from 2015-11-16 16 50 12](https://cloud.githubusercontent.com/assets/6511849/11192891/474dbf2e-8c82-11e5-86c9-3de2a77a31e5.png)

mkassner:
I agree. would splitting the lists make sense?

M

cpicanco:
Putting plugins in a growing menu as buttons makes sense?

willpatera:
Restructuring the plugin list is still needed. 

Splitting plugin lists into categories/topics might be a good first step. Plugins should also be sorted in some way (alphabetical?). Here is a proposal for splitting the list based on existing plugins:

  1. Visualization
    1. Vis Circle
    1. Vis Polyline
    1. Vis Lightpoints
    1. Vis Crosshair
    1. Vis Watermark
    1. Eye Video Overlay
    1. Scan Path
  2. Analysis  
    2. Fixation Detector
    2. Manual Gaze Correction
    2. Video Export Launcher
    2. Offline Marker Detector
    2. Raw Data Exporter
    2. Batch Exporter
    2. Annotation Player
3. Other
    3. Show Calibration
    3. Pupil Server

Once plugins are in separate categories/topics, then we can remove the prefix `Vis` for visualization plugins such that `Vis Circle` would just become `Circle`.

cpicanco:
Show Calibration should be included in the visualization category, shouldn't it?

mkassner:
implemented.

PR #500, opened 2016-10-17
Organized player plugin list into visualization, analysis, and other

snleu:


papr:
Nice idea! Could we have something similar for Capture? No idea about suitable categories though..

willpatera:
@snleu - I think one quick fix might be to simply sort the plugins by alphabetical order and name them appropriately (if not already). Many of the visualization plugins are already prefixed with `vis`.

mkassner:
I like this PR. I m good with merging unless somebody objects. The current list is to long for sure.

willpatera:
Ok, but please also sort alphabetically even if in separate lists before merging.

cpicanco:
related to #246

cpicanco:
PS.: A tab UI element would be great...

papr:
@cpicanco Do you have a specific use case for the tab ui in the current context? I would suggest to create an issue for that over at pupil/pyglui :)

cpicanco:
Please see: https://github.com/pupil-labs/pyglui/issues/42

PR #505, opened 2016-10-31
improve surface tracking

mkassner:
This PR:

- Improves marker tracking performace
- Improves surface learning
- Removed lens distortion model since it cause more harm then good
- Adds simple smoothing as a preview to Kalman filtering.


cpicanco:
Hi @mkassner , could give more details about the "surface learning" functionallity?

PR #518, opened 2016-11-12
fix indentation in line 248

cpicanco:
Please, see diffs.


mkassner:
what is the motivation for this? Just style?

cpicanco:
No, I thought it was a problem. Please, close if there is no problem at all.

PR #539, opened 2016-12-05
Surface Kalman Filtering

papr:
Uses [`pykalman`](https://pykalman.github.io/) to filter surfaces in Pupil Capture, thereby smoothing the vertices and predicting missing observation up to a specific confidence. Uses the `smooth` function in Pupil Player, thereby exploiting the knowledge of the whole data.

Todo:
- [ ] Adjust parameters


mkassner:
Current implementation is good enough for now.

PR #544, opened 2016-12-12
Non blocking captures

papr:
- [x] NDSI source
- [x] File source


papr:
Ready for review/merge

PR #570, opened 2017-01-02
Python3 and OpenCV3

mkassner:
This PR deprecates Python2.7 and OpenCV2 in favour of the newer versions Python3.6+ and OpenCV3. 

### Transition to Python3
Python3 is now at a stable state and offers new features. Its time to move forward!

Support for Python2.7 is possible but not sustainable from a testing and maintenance standpoint. We have thus decided **not** to support both python2 and python3 in parallel. The latest working version of Pupil using Python2.7 will exist at this [branch](https://github.com/pupil-labs/pupil/tree/python2.7). Backports are welcome!

### OpenCV3
OpenCV2 does not support Python3 so this is a no-brainer. At the same time OpenCV3 offers some exciting features we do not want to miss! More on that in future releases.

### Major changes

The one major change that goes beyond syntax with this PR is the fact that Python2 stings were **byte arrays** and Python3 strings are **unicode objects**. 

This change is visible to the user via the IPC backbone. We have decided to goes with the following convention:

- `topic` is a `unicode` string. It will be sent as `utf-8` encoded byte array.
- `payload` is a python dict. It will be sent as a msgpack serialized dict.


pupil_src/shared_modules/calibration_routines/manual_marker_calibration.py
```diff
@@ -271,7 +271,7 @@ def gl_display(self):
                 # cv2 requires integer arguments
                 pts = cv2.ellipse2Poly( (int(e[0][0]),int(e[0][1])),
                                     (int(e[1][0]/2),int(e[1][1]/2)),
-                                    int(e[-1]),0,360,int(360/self.counter_max))
+                                    int(e[-1]),0,360,360//self.counter_max)
```
willpatera: I didn't know about the `//` feature - very nice 👍

Tobias-Fischer:
Hi,
I am wondering how you're going to support Ubuntu - as even 16.04 does not support OpenCV3 out of the box. Are you going to recommend installing it through another repository? If so, which one? Or should it be compiled manually?

willpatera:
@Tobias-Fischer - Good point. I am testing out [LinuxBrew](http://linuxbrew.sh) right now to see if this could be an options for dependency installation on Linux.

mkassner:
@Tobias-Fischer do you have a recommendation?

Tobias-Fischer:
It seems like there are three alternatives:
1) LinuxBrew - never heard of it, so I don't have an opinion of it
2) Build from sources - pro: Works reliably, relatively straightforward; a tutorial on the installation should be written relatively quickly. The user should be made aware not to install it system wide, but instead point to the build directory. On the negative side, I don't know how you could bundle it, and compilation takes forever ..
3) Pre-build from a PPA. I had a brief look, and the most stable / reliable version seems to be the one from ROS. On the plus side, the packages are really stable, and installing `ros-xxx-opencv3` does not seem to install too many ROS specific dependencies (only `ros-xxx-catkin` which itself does not have ROS specific dependencies). On the negative side, it typically takes 2-3 months to release a new ROS distribution after a new Ubuntu has been released, and ROS only supports Ubuntu but no other distributions.

So if I were you, I'd go for option 2) and carefully write a tutorial for the installation, maybe setting all options which are not needed to run pupil to OFF so the compilation is a little faster. Alternatively and relatively easily, you could allow using the ROS packages as well (see my PR); without support and kind of unofficially ;).

What do you think @willpatera @mkassner?

Tobias-Fischer:
Update on the ROS / Ubuntu release cycle: http://wiki.ros.org/Distributions/ReleasePolicy
Basically there is a ROS release once a year in May, which means that the xx.10 version of Ubuntu is not being supported. There are probably ways around it, but I am not sure whether you want to have this hassle when supporting end-users. So this only provides support for Ubuntu xx.04 versions, and no other Linux distributions. I am not sure whether this would be suitable for pupil ..

willpatera:
Hi @Tobias-Fischer - yes that was a mistake. Fixing now.

mkassner:
pupil helpers PR is ready to be merged when this is ready.

PR #574, opened 2017-01-05
Fix batch script for Windows

PolynomialDivision:
The batch script wasn't working.
I have to do something like this.

### DIFF
--- a/pupil_src/capture/run_capture.bat
+++ b/pupil_src/capture/run_capture.bat
@@ -1,2 +1,2 @@
-PATH %PATH%..\..\pupil_external
+SET PATH=%PATH%;%~dp0..\..\pupil_external
 python main.py


pupil_src/capture/run_capture.bat
```diff
@@ -1,2 +1,2 @@
-PATH %PATH%..\..\pupil_external
+SET PATH=%PATH%;%~dp0..\..\pupil_external
```
willpatera: @PolynomialDivision - AFAIK using `SET` will add this path to the environment variables. 

This batch script without `SET` is intended to be used to add the `pupil_external` `.dll`s to the path only for the current session and not permanently. 

Regarding the addition of `%~dp0` - this will expand the path to the drive letter and the folder from which run_capture.bat resides. Example: `C:\work\pupil\pupil_src\capture` - so this seems like it makes sense to me. However, adding the extra `;` after `%PATH%` seems redundant (at least in my test it adds a double `;` in the path where only one is needed.

Can you please try:

`PATH %PATH%%~dp0..\..\pupil_external`

willpatera:
@PolynomialDivision - please see my review comments. I'd like to move forward with this PR or close it based on the comments.

pupil_src/capture/run_capture.bat
PolynomialDivision: PATH %PATH%%~dp0..\..\pupil_external

is not working for me.

I think that `SET` just temporary adds the path to the `PATH` variable. I thought this is what you want to achieve with it?

When you call 

    echo %PATH%
you see that there is no `;`

I'm using Windows 10 64 Bit.

PolynomialDivision: You can see this by running the batch script in a cmd and after closing do

    echo %PATH%

After this close the cmd and open a new one and do 

    echo %PATH%

You will see that ..\..\pupil_external will be removed.

willpatera: Hi @PolynomialDivision - thanks for the updates. You are correct in your usage of `set`. I was just re-reading documentation and `setx` is the command want to avoid. `set` is safe to use as you note. 

`SET` 
>Display, set, or remove CMD environment variables. Changes made with SET will remain **only for the duration of the current CMD session**.

`SETX` 
> Set environment variables permanently, SETX can be used to set Environment Variables for the machine or currently logged on user.

I will test this again and can merge after testing or follow up with more comments.

PR #576, opened 2017-01-09
Burn-in test changes

papr:


papr:
Merging for easier 3rd-party test installation.

@mkassner 6d480ee is the essential commit to review.

PR #581, opened 2017-01-13
Added blink detector

marc-tonsen:
The plugin was created on the Python 2.7 branch and will thus not automatically merge with the current master branch.


papr:
Replaced by #582

PR #582, opened 2017-01-13
Blink Detector Plugin

papr:
Mergable version from #581 + PEP 8 changes + code cleanup


pupil_src/capture/blink_detection.py
```diff
@@ -0,0 +1,92 @@
...
+class Blink_Detection(Plugin):
+    """
+    DisplayGaze shows the three most
+    recent gaze position on the screen
+    """
+
+    def __init__(self, g_pool):
+        super(Blink_Detection, self).__init__(g_pool)
+        self.order = .8
+        self.history_length = 21 # Should be odd
```
papr: What is the exact reason for this number to be odd?

pupil_src/capture/blink_detection.py
```diff
@@ -0,0 +1,92 @@
...
+            if self.is_blink and act < -2:
+                self.is_blink = False
+
+            print str(self.is_blink) + " " + str(act)
+
+            blink_entry = {
+                'topic': 'blink',
+                'activation': act,
+                'timestamp': self.timestamp_histories[pt['id']][10],
+                'is_blink': int(self.is_blink)
```
papr: What is the reason for `is_blink` to be an `int`? What does speak against it staying a `bool`?

pupil_src/capture/blink_detection.py
```diff
@@ -0,0 +1,92 @@
...
+                self.is_blink = True
+            if self.is_blink and act < -2:
+                self.is_blink = False
+
+            print str(self.is_blink) + " " + str(act)
+
+            blink_entry = {
+                'topic': 'blink',
+                'activation': act,
+                'timestamp': self.timestamp_histories[pt['id']][10],
```
papr: `10` can be replaced by `self. history_length // 2`, correct?

pupil_src/capture/blink_detection.py
```diff
@@ -0,0 +1,92 @@
...
+            if not self.is_blink and act > 7:
+                logger.error("Bink")
+                self.is_blink = True
+            if self.is_blink and act < -2:
+                self.is_blink = False
+
+            print str(self.is_blink) + " " + str(act)
+
+            blink_entry = {
+                'topic': 'blink',
```
papr: Is the `topic` entry necessary if the entry is added to `events['blinks']`? Are there other entries planned?

pupil_src/capture/blink_detection.py
marc-tonsen: I got pickle errors when I set that to a bool

pupil_src/capture/blink_detection.py
marc-tonsen: Correct. Ideally you would also set `self.history_length` to something that adapts to the framerate

papr:
Ready for merge.

marc-tonsen:
The plugin will only work well with the 3D detector, if the `confidence` value in events is changed to 2D pupil detection confidence

mkassner:
@papr  can you make this mergable?

papr:
@mkassner did you make the changes to confidence as @marc-tonsen requested?

mkassner:
@papr  no not yet.

PR #587, opened 2017-01-16
Fix brew glfw changes

### LINKED ISSUE #585: Problem: glfw3 brew install location changed
Hi there, 

I followed a thread about a similar issue on an old Mac and did exactly what it was specified.

I am running into 2 problems that I can't figure out.

First, when I:
cd pupil
git clone https://github.com/pupil-labs/pupil.git
git fetch
git checkout opencv3

I get:

error: pathspec 'opencv3' did not match any file(s) known to git

Second, when I try to run the main.py, I get:
File "main.py", line 49, in <module>
from multiprocessing import Process, Value, active_children, freeze_support, set_start_method
ImportError: cannot import name set_start_method

The reason why I'm actually trying to run the source code is that on my late 2013 MacBook Pro, the Pupil Capture crashes when I'm recording an experiment session (although I only have one other browser tab open).

Any ideas on how to deal with this?

papr:
Fix #585

### DIFF
--- a/pupil_src/shared_modules/glfw.py
+++ b/pupil_src/shared_modules/glfw.py
@@ -65,6 +65,8 @@ else:
         dll_path = find_library('glfw')
     elif os_name == "Darwin":
         dll_path = find_library('glfw3')
+        if not dll_path:
+            dll_path = find_library('glfw')
     elif os_name == "Windows":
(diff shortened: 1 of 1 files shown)


mkassner:
I thinks its better to update the install instruction to use brew install glfw (which is now glfw3) and simply do find library on glfw

mkassner:
This commit is also part of PR #586 . Please remove the commit from the other PR or close this PR  with a comment in the other PR

PR #602, opened 2017-01-26
Optimized pupil detector parameters

marc-tonsen:
Ran a series of grid searches to find the optimal parameter set on the LPW dataset.

### DIFF
--- a/pupil_src/capture/pupil_detectors/detector_2d.pyx
+++ b/pupil_src/capture/pupil_detectors/detector_2d.pyx
@@ -54,24 +54,24 @@ cdef class Detector_2D:
         self.coarseDetectionPreviousPosition =  (0,0)
         if not self.detectProperties:
             self.detectProperties["coarse_detection"] = True
-            self.detectProperties["coarse_filter_min"] = 150
-            self.detectProperties["coarse_filter_max"] = 300
-            self.detectProperties["intensity_range"] = 17
-            self.detectProperties["blur_size"] = 3
-            self.detectProperties["canny_treshold"] = 200
-            self.detectProperties["canny_ration"] = 3
+            self.detectProperties["coarse_filter_min"] = 128
+            self.detectProperties["coarse_filter_max"] = 280
+            self.detectProperties["intensity_range"] = 23
+            self.detectProperties["blur_size"] = 13
+            self.detectProperties["canny_treshold"] = 160
+            self.detectProperties["canny_ration"] = 2
             self.detectProperties["canny_aperture"] = 5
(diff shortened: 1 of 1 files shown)


mkassner:
I know this performs like a charm on the datasets. Does it work equally well in realtime? Does it change the behaviour of the 3d eye model?

marc-tonsen:
In a qualitative indoor performance test the new values seem to perform much better. Especially the robustness to occlusions is much better. The algorithm runs a lot slower though:

Old values: ~400 FPS
New values: ~160 FPS

We might want to to trade a little of the performance improvement in for higher FPS

marc-tonsen:
The performance decrease of the new parameter set is entirely due to the increased 'blur_size'. If that parameter is reset to it's old value the FPS actually increases compared to the original parameter set. There seems to be a clear sweet spot, when you set it to 5 (compare: 3 was the original value, 13 is the optimized value). For this value the computational performance is similar to that of the old parameter set and we still get ~95% of the accuracy gains. When increasing it further to 7 the FPS drop by 50% for small additional gains.

pupil_src/capture/pupil_detectors/detector_3d.pyx
```diff
@@ -63,26 +63,27 @@ cdef class Detector_3D:
...
-            self.detectProperties2D["strong_perimeter_ratio_range_max"] = 1.1
-            self.detectProperties2D["strong_area_ratio_range_min"] = 0.6
-            self.detectProperties2D["strong_area_ratio_range_max"] = 1.1
-            self.detectProperties2D["contour_size_min"] = 5
-            self.detectProperties2D["ellipse_roundness_ratio"] = 0.1
-            self.detectProperties2D["initial_ellipse_fit_treshhold"] = 1.8
-            self.detectProperties2D["final_perimeter_ratio_range_min"] = 0.6
-            self.detectProperties2D["final_perimeter_ratio_range_max"] = 1.2
-            self.detectProperties2D["ellipse_true_support_min_dist"] = 2.5
+            self.detectProperties["coarse_detection"] = True
```
willpatera: @marc-tonsen - all of these lines L66:85 should be `self.detectProperties2D` right?

marc-tonsen: Yes, that's right! I'll fix that.

mkassner:
Any reason not to merge?

willpatera:
@mkassner I just tested again on my _relatively_ low powered machine (MBAir 11" early 2014 i5 CPU 4gb ram) and it appears that performance is comparable to existing parameter settings. So, yes, I vote for merge. Shall I do it?

PR #618, opened 2017-02-09
Preserve Capture Software Version in info.csv

papr:
- Also fixes csv header handling

### DIFF
--- a/pupil_src/shared_modules/version_utils.py
+++ b/pupil_src/shared_modules/version_utils.py
@@ -73,7 +73,7 @@ def get_version(version_file=None):
 
 
 def read_rec_version(meta_info):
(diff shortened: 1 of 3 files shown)


pupil_src/shared_modules/version_utils.py
```diff
@@ -73,7 +73,7 @@ def get_version(version_file=None):
 
 
 def read_rec_version(meta_info):
-    version = meta_info["Capture Software Version"]
+    version = meta_info.get("Recording Version",meta_info["Capture Software Version"])
```
mkassner: this was not right I fixed it with 85ba470acfdacb823986a6d2adb38cb7a25dd6e5

PR #624, opened 2017-02-15
Pupil Mobile h.264 streaming

papr:
Requires pupil-labs/pyndsi#18


papr:
aab267e contains requested changes

PR #659, opened 2017-03-03
Fix Unicode Decode Errors 

papr:
...and use selection getter for Audio recording inputs


mkassner:
This should be two PR. Please refactor.

papr:
Removed audio source selection feature and split into #660

Issue #598, opened 2017-01-24
Offline Surface Tracker and Pupil Player crash issue


darksipu:
Hi there guys, 

So, we have 2 issues:

First, we ran Pupil Capture and recorded files around 1 minute with the source code version on a mid 2010 iMac. When we try to open any recordings made by using Pupil Capture on any other Macbook Pro running the GUI version of the Player, it instantly crashes. We tried with different file lengths and it doesn't work. Would it be a problem of using source code vs GUIs?

Second, the Offline Surface Tracker crashes when we want to export data with it in the Player (running the source code). In some instances, it also just crashes when you try to edit the surface that has to be tracked. 

If you can give us a hand, that'd be brilliant as we have to analyse some data for a research project that has to end relatively soon. Our aim is actually to get a surface defined on a screen and get the raw eye-gazes in this region, eventually splitting it up into pixels/different regions.

Thanks, 
Florin

papr:
It would be very helpful if you could upload the output of your terminal (assuming that you are running from source) to http://gist.github.com and paste the link below.

darksipu:
https://gist.github.com/darksipu/1f144f11b210c842992c7caaedd30772
This is opening a video, opening offline surface tracker, then adding a surface and it stopped working.

https://gist.github.com/darksipu/5e4121d1a0adc92c64efc10989edf421
This one is similar, as I tried to edit a surface.

For another video, I did manage to export from the offline surface tracker what was needed. Still, I have to go through 80 videos and get a surface tracked for each.

Thanks for the help.

darksipu:
And I tested a video recorded with the GUI Capture on the MBP on the source-code-run Player on the iMac and it worked. The issue only goes the other way around.

papr:
Hey @darksipu, thank you for the crash logs.

Could you clarify under which condition it does not work, i.e. Capture version + OS + Hardware during recording and Player version + OS + Hardware of the system on which you try to open the recording.

Am I correct to understand that only old recordings make Player crash?

darksipu:
So, recordings were made with a source-code-run 0.8.7.w.67 Capture on a El Capitan 10.11.6 iMac (2.8GHz, mid 2007, 4GB RAM, DDR2). On the same machine, the Player (same version) gave the crash reports that I've sent you.

[...]

PR #662, opened 2017-03-07
FIX issue #598

Aeefire:
min/maxing to in an integer value caused an integer to be returned in the edge case, instead of the wanted float


pupil_src/shared_modules/calibration_routines/gaze_mappers.py
```diff
@@ -28,7 +28,7 @@ def _clamp_norm_point(pos):
         Grossly bigger or smaller numbers are results bad exrapolation
```
willpatera: We could also fix this typo `exrapolation` to extrapolation

willpatera: @mkassner fix after merge.

willpatera: L27 should also be realistic

PR #678, opened 2017-03-22
add runtime calibration plugins to calibration plugins list

cpicanco:


mkassner:
Maybe better like this? https://github.com/pupil-labs/pupil/pull/679

cpicanco:
Yes, requiring the calibration base prevents loading other classes
errouneously.

PR #679, opened 2017-03-23
load calibration plugins appropriately

mkassner:


papr:
Could you add the same for Base_source?

mkassner:
I guess yes! base_manager would be the right one I guess.

papr:
Of course!

PR #680, opened 2017-03-24
Blinks

euryalus:
Hey! Just changed few lines in the code. Basically, the activations that were calculated were NaN in case the confidence was 1 over the whole history window. In that case activations should be zero, however. Also I updated the docstring and changed a few typos in the comments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,6 @@ SphereCircleTest.cpp
 *.mp4
 *.mkv
 *.egg-info
+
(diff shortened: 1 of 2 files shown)


euryalus:
Just realized that currently this is not working when both eye cameras are active. I am looking for the problem.

euryalus:
Ok, this should work now. When activations are integers and both cameras were active, there was a problem.

PR #686, opened 2017-03-30
Immutable dicts and msgpack as file format

papr:
Use `msgpack` instead of `pickle` to store configuration and data files. Unpack mappings as `Immutable_Dict` instead of `dict`. Bump data format 0.9.4


pupil_src/player/player_methods.py
```diff
@@ -165,10 +176,9 @@ def update_recording_v083_to_v086(rec_dir):
 
 def update_recording_v086_to_v087(rec_dir):
     logger.info("Updating recording from v0.8.6 format to v0.8.7 format")
-    pupil_data = load_object(os.path.join(rec_dir, "pupil_data"))
+    pupil_data = load_legacy_object(os.path.join(rec_dir, "pupil_data"))
```
mkassner: this should be called load_object_legacy

pupil_src/player/player_methods.py
```diff
@@ -200,25 +211,61 @@ def update_recording_v087_to_v091(rec_dir):
 
     update_meta_info(rec_dir,meta_info)
 
+
 def update_recording_v091_to_v093(rec_dir):
     logger.info("Updating recording from v0.9.1 format to v0.9.3 format")
     meta_info_path = os.path.join(rec_dir,"info.csv")
-    pupil_data = load_object(os.path.join(rec_dir, "pupil_data"))
-
+    pupil_data = load_legacy_object(os.path.join(rec_dir, "pupil_data"))
```
mkassner: load_object_legacy

pupil_src/player/player_methods.py
```diff
@@ -200,25 +211,61 @@ def update_recording_v087_to_v091(rec_dir):
...
+        if file.startswith('.') or os.path.splitext(file)[1] in ('.mp4', '.avi'):
+            continue
+        rec_file = os.path.join(rec_dir, file)
+
+        def convert(data):
+            if isinstance(data, np.ndarray):
+                return data.tolist()
+            elif isinstance(data, str):
+                return data
+            elif isinstance(data, collections.Mapping):
```
mkassner: why convert cases other than array?

pupil_src/player/player_methods.py
```diff
@@ -200,25 +211,61 @@ def update_recording_v087_to_v091(rec_dir):
...
+                return dict(map(convert, data.items()))
+            elif isinstance(data, collections.Iterable):
+                return type(data)(map(convert, data))
+            else:
+                return data
+
+        try:
+            rec_object = load_legacy_object(rec_file)
+            converted_object = convert(rec_object)
+            if converted_object != rec_object:
```
mkassner: this check will not work with mutable nested objects, I think.

pupil_src/shared_modules/calibration_routines/calibrate.py
```diff
@@ -46,13 +46,13 @@ def calibrate_2d_polynomial(cal_pt_cloud,screen_size=(1,1),threshold = 35, binoc
         logger.info('used {} data points out of the full dataset {}: subset is {:.2f} percent'.format(
             used_num, complete_num, 100*float(used_num)/complete_num))
 
-        return map_fn, err_dist <= threshold, (cx, cy, model_n)
+        return map_fn, err_dist <= threshold, ([p.tolist() for p in cx], [p.tolist() for p in cy], model_n)
```
mkassner: are these lists of ndarrays? Or why can we not just do cx.tolist()

pupil_src/shared_modules/calibration_routines/finish_calibration.py
```diff
@@ -72,13 +72,13 @@ def finish_calibration(g_pool,pupil_list,ref_list):
...
             scales = list(np.linspace(0.7,1.4,20))
-            K = camera_intrinsics["camera_matrix"]
+            K = np.asarray(camera_intrinsics["camera_matrix"])
 
             for s in scales:
                 scale = np.ones(K.shape)
                 scale[0,0] *= s
                 scale[1,1] *= s
-                camera_intrinsics["camera_matrix"] = K*scale
+                camera_intrinsics["camera_matrix"] = (K*scale).tolist()
```
mkassner: can this conversion not be done futher down outside of the loop?

pupil_src/shared_modules/calibration_routines/finish_calibration.py
```diff
@@ -184,12 +183,12 @@ def toWorld1(p):
...
             smallest_residual = 1000
             scales = list(np.linspace(0.7,1.4,20))
-            K = camera_intrinsics["camera_matrix"]
+            K = np.asarray(camera_intrinsics["camera_matrix"])
             for s in scales:
                 scale = np.ones(K.shape)
                 scale[0,0] *= s
                 scale[1,1] *= s
-                camera_intrinsics["camera_matrix"] = K*scale
+                camera_intrinsics["camera_matrix"] = (K*scale).tolist()
```
mkassner: same as above

mkassner:
I think changes to hmd calibration plugin are missing.

pupil_src/shared_modules/calibration_routines/calibrate.py
papr: Exactly. Lists of ndarrays. Alternatively, one could use `np.asrray(obj).tolist()`.

papr:
> However I would recommend using the normal msgpack.loads without custom dict for all cases except when loading pupil data.

This should include unpacking iterables as lists instead of tuples. pupil data would be unpacked into immutable objects (tuple and Immutable_Dict), all other into mutable ones (dicts and lists).

pupil_src/player/player_methods.py
papr: So that cases like `{key: <np.ndarray>}` or `[<np.ndarray>]` are also converted. `str` is a special case because `str([*'abc'])` results in `"['a', 'b', 'c']"`.

pupil_src/player/player_methods.py
papr: Not true :)
```python
{'a': [{'b': 1}]} == {'a': [{'b': 1}]}
> True
{'a': [{'b': 1}]} == {'a': [{'b': 0}]}
> False
```

pupil_src/shared_modules/calibration_routines/calibrate.py
mkassner: I guess your first way was more explicit. So we keep that.

pupil_src/player/player_methods.py
mkassner: would you mind adding inline comments to make this clear?

mkassner:
merging into work branch

PR #715, opened 2017-05-03
Offline Pupil Detection and Gaze Mapping

papr:


cpicanco:
Hi, I know this is WIP, but I would like to leave some comments. While testing the offline calibration plugin, which worked beautifully AFAIK, I thought that a plugin to show on screen calibration markers and to control some marker detection variables would be great. Also, a plugin for post-processing of the world image, like sharpness, would be very useful. Please, make know a little bit more about how to pushing this branch forward.

cpicanco:
Also, why not use trim markers to avoid searching for marker through the whole recording out?

papr:
@cpicanco Please be patient with us. We have a lot of ideas that need testing first. You will see commits in this PR as soon as they are implemented.

cpicanco:
@papr , I was asking for some words to do some work by myself. You're awesome, sorry for being inpatient :smile:

papr:
Oh, I understand. I really would like to lay some work off to you ;) but we are still in the designing phase. e.g. yes, we had the trim marker idea as well, but what happens if people want to calibrate multiple times per recording? To which time range is the current calibration applied? etc.

We got ideas that solve these issues. But on the other side, Player did not get the same backbone code refactory as Capture did with the IPC. We are missing a good infrastructure that
* handles long recordings
* runs such that the UI is not getting blocked
* handles plugin dependencies
* provides a processing chain where plugins can announce and listen to changes in data streams
* is easy to implement
* is easy to understand
* is easily extendable
* requires as  few conceptional changes to existing plugins as possible

This list is just from the top of my head, there is a lot missing

A lot of plugins will need a refactory as soon as this infrastructure stands. At this point in time your help will be very appreciated!

If you really want to do something now have a look at `background_helper.py` and `offline_calibration.py`. Understand how the background helper is used and apply the same pattern to the offline fixation detector. Fixation detection is happening currently in the main thread and blocks the ui for a while on plugin init. This is not great and needs to be fixed. Please be aware, that this plugin might require an other refactory as soon the "backbone" is in place, although the background helper is the first step to this new backbone.

cpicanco:
@papr , thank you for making me aware of your early blue prints. Despite really interested, I feel I am not ready to and would not have time to contribute to the backbone. However, It gives me a sense about how many time I will need to wait until some features be available.

Multiple calibrations during a recording are possible, but are them desirable for the end-user? Building a whole system over this assumption does not sounds wise considering how fast your are, considering slip. compensation and other planned improvements in the 3d eye model. Am I over expecting stuff from you? rsrsr

But let consider that multiple calibrations during a recording proves to be unavoidable (assuming that a depth sensible calibration is impossible) or, who knows, it proves to be the best idea in the short time (assuming that people will need to wait too long for better features). An alternative (brute forced) idea for Pupil recordings is to record calibration IDs and let people choose to which data apply each of them. A default may be applying a calibration to the following data until a new successful calibration occurs. Another alternative is to, somehow, incrementally improve calibration.

pupil_src/player/gaze_producers.py
```diff
@@ -32,6 +32,23 @@
...
 
+class Gaze_Producer_Base(Plugin):
+    uniqueness = 'by_base_class'
+    order = .02
+
+
+class Gaze_From_Recording(Gaze_Producer_Base):
+    def __init__(self, g_pool):
+        super().__init__(g_pool)
+        gaze_data_path = os.path.join(g_pool.rec_dir, "pupil_data")
```
mkassner: this is slow and takes a loot of RAM. Rather let the main.py script load pupil_data into g_pool.pupil_data and then access that from here.

pupil_src/player/main.py
```diff
@@ -254,9 +263,9 @@ def set_scale(new_scale):
 
     # load pupil_positions, gaze_positions
     pupil_data = load_object(pupil_data_path)
-    pupil_list = pupil_data['pupil_positions']
-    gaze_list = pupil_data['gaze_positions']
-    g_pool.pupil_data = pupil_data
+    g_pool.pupil_data = {'pupil_positions': [],  # set by pupil producers
```
mkassner: load real pupil_data here.

pupil_src/player/main.py
```diff
@@ -268,8 +277,8 @@ def set_scale(new_scale):
     g_pool.rec_dir = rec_dir
     g_pool.meta_info = meta_info
     g_pool.min_data_confidence = session_settings.get('min_data_confidence', 0.6)
-    g_pool.pupil_positions_by_frame = correlate_data(pupil_list, g_pool.timestamps)
-    g_pool.gaze_positions_by_frame = correlate_data(gaze_list, g_pool.timestamps)
+    g_pool.pupil_positions_by_frame = [[] for x in g_pool.timestamps]
```
mkassner: this is good.

pupil_src/player/pupil_producers.py
```diff
@@ -39,13 +41,27 @@
...
-    """docstring for Offline_Pupil_Detection"""
+class Pupil_Producer_Base(Plugin):
+    uniqueness = 'by_base_class'
+    order = 0.01
+
+
+class Pupil_From_Recording(Pupil_Producer_Base):
     def __init__(self, g_pool):
         super().__init__(g_pool)
+        pupil_data_path = os.path.join(g_pool.rec_dir, "pupil_data")
```
mkassner: same here. Dont load pupil_data. Just access g_pool.pupil_data['pupil_positions'] here.

pupil_src/player/main.py
```diff
@@ -149,13 +151,20 @@ def session(rec_dir):
                               Pupil_Angle_3D_Fixation_Detector,
                               Manual_Gaze_Correction, Video_Export_Launcher,
                               Offline_Surface_Tracker, Raw_Data_Exporter,
-                              Batch_Exporter, Annotation_Player,
-                              Offline_Pupil_Detection, Offline_Calibration], key=lambda x: x.__name__)
+                              Batch_Exporter, Annotation_Player], key=lambda x: x.__name__)
+
+    global pupil_producers
```
mkassner: whats the reason for global here?

pupil_src/player/main.py
```diff
@@ -149,13 +151,20 @@ def session(rec_dir):
...
+    global gaze_producers
+
+    pupil_producers += [p for p in runtime_plugins if issubclass(p, Pupil_Producer_Base)]
```
mkassner: I d rather not have these lists here. lets refactor the plugin logic so that spliting into categories happens via a base class in the menu selector and not explicitly in code here.

pupil_src/player/gaze_producers.py
mkassner: dont load pupil_object here. its slow and doublicates ram every time. better: access self.g_pool.pupil_data

pupil_src/player/main.py
```diff
@@ -254,9 +263,9 @@ def set_scale(new_scale):
 
     # load pupil_positions, gaze_positions
     pupil_data = load_object(pupil_data_path)
-    pupil_list = pupil_data['pupil_positions']
-    gaze_list = pupil_data['gaze_positions']
-    g_pool.pupil_data = pupil_data
```
mkassner: leave this in

pupil_src/player/main.py
mkassner: remove this.

pupil_src/player/main.py
mkassner: this is good.

pupil_src/player/pupil_producers.py
mkassner: same here as with gaze_producer.

pupil_src/shared_modules/background_helper.py
```diff
@@ -0,0 +1,60 @@
+'''
```
mkassner: I really like this abstraction. However I feel it could be more pythonic to make the tast_proxy a class with methods instead of fns with proxy_obj. The current implementation feels like C. 

Any reason not to make this a class?

mkassner: resolved

pupil_src/shared_modules/annotations.py
```diff
@@ -117,7 +114,7 @@ def cleanup(self):
         self.deinit_gui()
 
 
-class Annotation_Player(Annotation_Capture):
+class Annotation_Player(Annotation_Capture, Analysis_Plugin_Base):
```
mkassner: should Annotation_Capture not already be an subcalss of Analysis_Plugin_Base ?

pupil_src/shared_modules/calibration_routines/gaze_mappers.py
```diff
@@ -434,7 +437,7 @@ def _map_binocular(self, p0, p1):
             image_point = _clamp_norm_point(image_point)
 
 
-        if self.visualizer.window:
+        if hasattr(self, 'visualizer') and self.visualizer.window:
```
mkassner: better add self.visualizer = None in this classes __init__

mkassner: then replace all hasattr with if self.visualizer

mkassner:
I have continued work based on this PR. The work of this PR will be merged into master here: #756

PR #717, opened 2017-05-04
Add version requirement for pyre >= 0.3.1

papr:
Pyre does not have .version() implemented before 0.3.2,
therefore the relatively complex version test.

+ small typo fixes

### DIFF
--- a/pupil_src/shared_modules/time_sync.py
+++ b/pupil_src/shared_modules/time_sync.py
@@ -23,10 +23,19 @@ import logging
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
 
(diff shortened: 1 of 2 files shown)


pupil_src/shared_modules/time_sync.py
```diff
@@ -23,18 +23,27 @@
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
 
+pyre_requirement = 'Pyre >= 0.3.1 is required for Time Sync to work'
```
mkassner: what about this:
```python
try:
    from pyre import __version__
    assert __version__ >= '0.3.1'
except (ImportError,AssertionError):
    raise Exception("Pyre version is to old. Please upgrade")
    
```

PR #720, opened 2017-05-12
Require pyndsi version 0.3.0 and higher

papr:
To be merged when Pupil Mobile has implemented Time Sync.


papr:
This should not have been merged yet -- as the PR test said. pyndsi 0.3.0 is not released yet nor are the new changes implemented in Pupil Mobile.

I will push a revert of this merge.

papr:
I will re-do this PR when it is the appropriate time.

PR #721, opened 2017-05-12
TimeSync: Introduce protocol versioning and change group name convention

papr:


mkassner:
I will merge this shortly before release.

PR #730, opened 2017-05-18
Fix inline module builds on windows

papr:


mkassner:
@willpatera  please test on windows and merge.

PR #732, opened 2017-05-26
Added one target calibration

marc-tonsen:
Added a new calibration routine: A single gaze target is displayed statically on the screen, the user gazes onto the target while moving his head to quickly sample many different gaze angles. 
With this routine one can quickly sample a great range densely.

### DIFF
--- a/pupil_src/shared_modules/calibration_routines/one_target_calibration.py
+++ b/pupil_src/shared_modules/calibration_routines/one_target_calibration.py
@@ -1,0 +1,324 @@
+'''
+(*)~---------------------------------------------------------------------------
+Pupil - eye tracking platform
+Copyright (C) 2012-2017  Pupil Labs
+
+Distributed under the terms of the GNU
+Lesser General Public License (LGPL v3.0).
+See COPYING and COPYING.LESSER for license details.
+---------------------------------------------------------------------------~(*)
+'''
+
+import os
+import cv2
+import numpy as np
+from methods import normalize,denormalize
+from gl_utils import adjust_gl_view,clear_gl_screen,basic_gl_setup
+import OpenGL.GL as gl
+from glfw import *
+from circle_detector import find_concetric_circles
+from file_methods import load_object,save_object
+from platform import system
+
+import audio
+
+from pyglui import ui
+from pyglui.cygl.utils import draw_points, draw_points_norm, draw_polyline, draw_polyline_norm, RGBA,draw_concentric_circles
+from pyglui.pyfontstash import fontstash
+from pyglui.ui import get_opensans_font_path
+from . calibration_plugin_base import Calibration_Plugin
+from . finish_calibration import finish_calibration
+
+#logging
+import logging
+logger = logging.getLogger(__name__)
+
+# window calbacks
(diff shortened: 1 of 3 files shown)


pupil_src/shared_modules/calibration_routines/one_target_calibration.py
```diff
@@ -0,0 +1,323 @@
...
+def on_resize(window,w,h):
+    active_window = glfwGetCurrentContext()
+    glfwMakeContextCurrent(window)
+    adjust_gl_view(w,h)
+    glfwMakeContextCurrent(active_window)
+
+
+
+class One_Target_Calibration(Calibration_Plugin):
+    """Calibrate using a marker on your screen
```
mkassner: please update this.

mkassner:
Final touch: think about the name? maybe single_marker calibration? or stationary_marker_calibration?

cpicanco:
+1 stationary_marker_calibration
+2 static_marker_calibration

willpatera:
The word `stationary` could be more easily misinterpreted IMO. I would opt for the word `static` if we only have a few choices. However, both of these terms could be misleading to the first-time user. 

This calibration procedure might be better described by what moves instead of what is not moving. The marker is fixed in place. The participant fixes their gaze on the marker center throughout the duration of the calibration. The participant _rotates_ their head while fixating on the marker center to collect samples within their FOV. 

Here is another alternative that is participant-centric: `head_rotation_calibration` or `neck_rotation_calibration` - "Look at the marker center, rotate your head while looking at the marker center to cover your FOV."

cpicanco:
Will, your naming suggestion is not consistent with existing calibration
names. Calibration routines are being described by stuff in the environment
and presentation styles, not by eye movements. Regardless of what name will
be chosen, I would suggest to keep consistency with existing names. For
users, pupil docs will describe the procedure in detail. For programmers,
the doc section will also describe further details and so avoid any
misleading name only feature.

On Jun 2, 2017 00:14, "Will" <notifications@github.com> wrote:

The word stationary could be more easily misinterpreted IMO. I would opt
for the word static if we only have a few choices. However, both of these
terms could be misleading to the first-time user.

This calibration procedure might be better described by what moves instead
of what is not moving. The marker is fixed in place. The participant fixes
their gaze on the marker center throughout the duration of the calibration.
The participant *rotates* their head while fixating on the marker center to
collect samples within their FOV.

Here is another alternative that is participant-centric:
head_rotation_calibration or neck_rotation_calibration - "Look at the
marker center, rotate your head while looking at the marker center to cover
your FOV."

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/pupil-labs/pupil/pull/732#issuecomment-305678151>, or mute
the thread

mkassner:
OK. lets conclude on this:

single_marker_calibration
one_marker_calibration
...

willpatera:
@mkassner I vote for `single_marker_calibration`

papr:
+1 on `single*`

mkassner:
@marc-tonsen  please change the plugin name and I will merge this PR.

PR #739, opened 2017-06-06
Fix eye overlay plugin

### LINKED ISSUE #738: Minor issue - Frame scrubber problem in Pupil Player
When I drag the frame scrubber in Pupil Player, when I release white ball, it jumps ahead to a different part of the video. As near as I can tell, this is proportional to where I am trying to go in the video, e.g. if I am scrubbing to a point early in the video, it jumps ahead a small amount. If I am scrolling near the end of the video, it jumps ahead a large amount. 

It looks as though when you try to scroll to a particular part of a video it adds the frame number to the current frame upon release of the mouse button (i.e, dragging the white ball to frame#300 jumps to frame#600 upon release). 

Not a major issue, just annoying :)

papr:
Fixes #738

### DIFF
--- a/pupil_src/player/vis_eye_video_overlay.py
+++ b/pupil_src/player/vis_eye_video_overlay.py
@@ -268,22 +268,27 @@ class Vis_Eye_Video_Overlay(Plugin):
                     if pd['id'] == eye_index and pd['timestamp'] == self.eye_frames[eye_index].timestamp:
                         break
 
-                el = pd['ellipse']
-                conf = int(pd.get('model_confidence', pd.get('confidence', 0.1)) * 255)
-                center = list(map(lambda val: int(self.eye_scale_factor*val), el['center']))
-                el['axes'] = tuple(map(lambda val: int(self.eye_scale_factor*val/2), el['axes']))
-                el['angle'] = int(el['angle'])
-                el_points = cv2.ellipse2Poly(tuple(center), el['axes'], el['angle'], 0, 360, 1)
+                try:
+                    el = pd['ellipse']
+                except KeyError:
+                    logger.warning('3d pupil data is required to visualize detected ellipses')
+                    self.show_ellipses = False
+                else:
+                    conf = int(pd.get('model_confidence', pd.get('confidence', 0.1)) * 255)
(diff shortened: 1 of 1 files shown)


mkassner:
@papr  2d pupil data also has an ellipse. this fix should not be required. Have you tested running a normal 2d recording without this fix?

 @cpicanco pupil_data file had data from a  very old 2d pupil detector written in python.

papr:
@mkassner No, I did not test it with 2d data. I assumed @cpicanco data to be up-to-date. But this means we should change the warning to something containing the words "deprecated data format".

mkassner:
@papr  I dont want to add a warning like this. We need to assume that the data is conform with what we expect at the version. Glue and fixing like this belongs in player methods.  We cant start proving code against non-confirm pupil_data.

papr:
Ok, are you suggesting to remove the `try-except` fix completely? Player-methods will only be called on upgrading the recording. It is very likely that most of the recordings are already upgraded to the point where player methods are not called anymore. And even if they were called, there would be no way to generate the correct ellipses for the deprecated data.

mkassner:
Yes this is what I am suggesting. Just remove your fix. 

If @cpicanco  confirms that his recording was made on a unmodified version of Pupil Im happy to invest time to add a fix in player methods.

mkassner:
the renaming part should remain in this PR

mkassner:
I just confirmed the current master branch eye visulizer for player works just fine with 2d and 3d data.

@papr please do not fix problems without reproducing the error first. Please make a new PR that contains only the renaming commit.

cpicanco:
I can confirm that I was using an unmodified version of an old pupil 2d
capture. My modifications are being done in pupil player plugins. I do
remember that ellipese data was added after 4.x capture version by requests
from other pupil community members. I would suggest to leave any
redetection task for the offline detector. Is breaking support for 4.x data
and older in your mind?

mkassner:
In this case I would say we add a converter in player_methods to add dummy data.

papr:
Or we generate circles from `norm_pos` and `diameter`. That would not be total dummy data at least.

mkassner:
@cpicanco alright this info helps a lot! Let add this to the conversion step in player_methods.

PR #771, opened 2017-07-25
Add on_unconsumed_key/char events

papr:
Requires pyglui >=1.6

This is the possibility for plugins to access unconsumed* GLFW events without having to overwrite the GLFW callbacks, as requested internally.

\* "unconsumed" means that there was no gui element that used the event for itself, e.g. no `Text_Input` used the char as input.


papr:
Changed as requested.

PR #782, opened 2017-08-03
Realsense Backend

papr:


mkassner:
I made changes to your branch, cygl and pyglui. Please check out the v1.7 branch of pylgui and install to test with my changes. Please pull my changes to your pr into your fork.

General feedback. I think the backend on Mac is not quite as smooth as it is on Linux. Some settings do not work and the framerate control seems to be reversed.

Image quality of the rgb stream is worse than when running the librealsense examples. I m sure this is due to settings on the camera being funny (very grainy and laggy video.)

It would actually be nice to expose the ir l and r video streams. But that can wait.

Issue #605, opened 2017-01-30
replace current camera distortion model with fisheye model


mkassner:
its all in the `cv2.fisheye` namespace.
Note that using a recent version of opencv is highly recommended for some bugfixes like:

https://github.com/opencv/opencv/pull/6454

papr:
Am I correct that [`camera_intrinsics_estimation.py`](https://github.com/pupil-labs/pupil/blob/master/pupil_src/shared_modules/calibration_routines/camera_intrinsics_estimation.py) is the place to apply these changes?

mkassner:
Yes. But also all occurrences of distort and undistort in the codebase

Tobias-Fischer:
Hi @papr @mkassner @marc-tonsen,
Do you have any news on this issue? I'd be happy to contribute.

Best, Tobias

marc-tonsen:
Hi @Tobias-Fischer !

I have created a PR #789 that includes a draft of how the fisheye camera model can be incorporated into Capture. Implementing this functionality is a bit messy because the cv2.fisheye module is buggy and requires extra code to deal with the bugs. Also I had to change the way the camera intrinsics are accessed in Capture, respectively the way the functionality dealing with lens distortion is accessed. As you can see in the PR I have introduced _camera model objects_ that encapsulate this functionality.

The parts of the code dealing with surface tracking have not been correctly updated yet, so they do not work in this PR. I will try to this soon. Camera calibration and gaze mapping is open for testing!

Best,
Marc

mkassner:
closed with #789

PR #789, opened 2017-08-08
Fisheye Camera Model

marc-tonsen:
In this PR camera intrinsics are modeled explicitly in _camera model objects_. The `g_pool.capture` object has an attribute `intrinsics`, which is such a camera model. The camera model provides all functionality necessary to deal with lense distortion, which includes functions like `projectPoints` or `undistortPoints`. The modules for _gaze mapping_ and _camera intrinsics calibration_ have been updated to use the functionality  provided by the camera model instead of calling opencv functions directly.

Besides the standard opencv camera model, which uses radial distortion and was used throughout Capture previously, the PR also includes a camera model with fisheye distortion. The camera intrinsics calibration plugin has been updated to support both types of distortion. One can select the appropriate type via a dropdown menu.

The pre-recorded calibrations for `Pupil Labs Cam 1 ID2` have been updated to fisheye models. Custom calibrations recorded with the camera intrinsics calibration plugin are now accumulated in one file per camera (identified by name). Calibrations for different resolutions are thus stored in the same file.

The modules concerned with surface tracking have not yet been converted properly and might not work correctly.

This PR is connected to #605


marc-tonsen:
Update: The surface tracker is now also using the camera model functionality. The locate 3D function has only been implemented with a best effort, since tracking surfaces in 3D the way Capture does it has more fundamental problems and this functionality will be deprecated soon in favor of a proper 3D pose tracker based on environment markers.

Tobias-Fischer:
Hi,
unfortunately I cannot try this PR as pupil capture fails to start:
```
Process world:
  File "/usr/lib/python3.5/multiprocessing/process.py", line 249, in _bootstrap
    self.run()
  File "/usr/lib/python3.5/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/home/tobias/robot/pupil/pupil/pupil_src/launchables/world.py", line 99, in world
    from calibration_routines import calibration_plugins, gaze_mapping_plugins, Calibration_Plugin
  File "/home/tobias/robot/pupil/pupil/pupil_src/shared_modules/calibration_routines/__init__.py", line 21, in <module>
    from . hmd_calibration import HMD_Calibration, HMD_Calibration_3D
  File "/home/tobias/robot/pupil/pupil/pupil_src/shared_modules/calibration_routines/hmd_calibration.py", line 24, in <module>
    from . camera_intrinsics_estimation import idealized_camera_calibration
ImportError: cannot import name 'idealized_camera_calibration'
```

marc-tonsen:
@Tobias-Fischer Those errors should be fixed now! Thanks for the report!

Tobias-Fischer:
Seems to be working nicely now :).
Just out of interest - which advantages does the fish eye model have? Is the accuracy going to be higher? And is the standard calibration taken transformed from the radial model (i.e. basically the same as before) or did you calibrate again using the fisheye model?

marc-tonsen:
I am glad to hear that @Tobias-Fischer ! Since our lenses offer a large FOV the distortion introduced by them should be more accurately modeled with a fisheye distortion model. This should in turn lead to greater accuracy in the 3D gaze mapper, the surface tracker and (once it gets implemented) the 3D marker tracker. We have not yet properly evaluate this though.

Tobias-Fischer:
Great, that sounds good. Thanks! Do you think I can use this PR in "production code" (i.e. my daily research), or shall I wait further until you evaluated this further?

papr:
@Tobias-Fischer We still need to adapt this new feature to Pupil Player. Therefore we **do not recommend** to use this in production yet. Additionally, tests on other platforms than Linux are required.

Tobias-Fischer:
Okay .. seeing that I only use capture (along with real-time streaming) on Linux though (i.e. I never touch player / Windows / OS X), what would you recommend in that case?

papr:
No, we are still testing and making changes. We recommend using it as soon as it is merged into `master`.

PR #819, opened 2017-08-29
Final fixes before v0.9.14 release

papr:
See commit list

### DIFF
--- a/pupil_src/shared_modules/calibration_routines/single_marker_calibration.py
+++ b/pupil_src/shared_modules/calibration_routines/single_marker_calibration.py
@@ -9,11 +9,6 @@ See COPYING and COPYING.LESSER for license details.
 ---------------------------------------------------------------------------~(*)
 '''
 
-'''
-Reference:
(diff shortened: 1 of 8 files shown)


pupil_src/shared_modules/calibration_routines/single_marker_calibration.py
```diff
@@ -9,6 +9,11 @@
 ---------------------------------------------------------------------------~(*)
 '''
 
+'''
+Reference:
+Santini, Thiago, Wolfgang Fuhl, and Enkelejda Kasneci. "CalibMe: Fast and Unsupervised Eye Tracker Calibration for Gaze-Based Pervasive Human-Computer Interaction." Proceedings of the 2017 CHI Conference on Human Factors in Computing Systems. ACM, 2017.
```
willpatera: @papr should this be be broken into a couple of lines, so that it does not create one long line of text here?

papr: This will be moved to the docs.

willpatera: Yes, that is a good idea.

PR #820, opened 2017-08-29
Dilation History plugin for Pupil Capture

papr:
Uses `pylgui.Bar_Graph`s to visualize the `diameter_3d` value for pupil positions with confidence `>0.6`

### DIFF
--- a/pupil_src/shared_modules/dialation_history.py
+++ b/pupil_src/shared_modules/dialation_history.py
@@ -1,0 +1,84 @@
+'''
+(*)~---------------------------------------------------------------------------
+Pupil - eye tracking platform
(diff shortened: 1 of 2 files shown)


pupil_src/shared_modules/dialation_history.py
```diff
@@ -0,0 +1,84 @@
...
+
+import gl_utils
+
+# logging
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+
+
+class Dialation_History(Plugin):
+    """docstring for Dialation_History"""
```
willpatera: High level documentation would be desirable here @papr

PR #839, opened 2017-09-26
Add 3d Visualization

pointcontrols:


papr:
I will close this in favor of #863 which includes the visualization within an extra window.

Issue #856, opened 2017-10-10
Pupil Group Issue/Enhancement


fneitzel:
If the world camera is not initialized, then the recording will not start on the uninitialized computer. However it will start on the other computers in the group. 
If you press recording again on the uninitialized computer the recording will not stop on the other computers.

mkassner:
@papr I see some options: make clear that the user should switch to fakeCapture or allow recordings with no world video. Any other ideas?

papr:
I think it will be difficult to not have a world video. Where would we get the world timestamps from? We could automatically start a fake capture though...

papr:
We decided to not go for the automatic fake capture activation. To solve this issue one would have to intercept notifications after they have been sent out. Implementing this would increase the code complexity disproportionately. In the end this is a configuration issue not a program issue.

PR #867, opened 2017-10-16
Start Fake Capture on recording start if current capture is offline

papr:
A possible fix for #856. @fneitzel @mkassner what do you think?

### DIFF
--- a/pupil_src/shared_modules/recorder.py
+++ b/pupil_src/shared_modules/recorder.py
@@ -189,7 +189,14 @@ class Recorder(System_Plugin_Base):
             if self.running:
                 logger.info('Recording already running!')
             elif not self.g_pool.capture.online:
-                logger.error("Current world capture is offline. Please reconnect or switch to fake capture")
+                logger.warning("Current world capture is offline. Starting fake capture.")
+                settings = {}
+                settings['frame_rate'] = self.g_pool.capture.frame_rate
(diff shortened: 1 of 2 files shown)


pupil_src/shared_modules/recorder.py
```diff
@@ -189,7 +189,14 @@ def on_notify(self, notification):
...
                 logger.info('Recording already running!')
             elif not self.g_pool.capture.online:
-                logger.error("Current world capture is offline. Please reconnect or switch to fake capture")
+                logger.warning("Current world capture is offline. Starting fake capture.")
+                settings = {}
+                settings['frame_rate'] = self.g_pool.capture.frame_rate
+                settings['frame_size'] = self.g_pool.capture.frame_size
+                settings['name'] = self.g_pool.capture.name
+                self.notify_all({'subject': 'start_plugin', "name": "Fake_Source", 'args': settings})
+                self.notify_all({'subject': 'recording.should_start',
```
mkassner: you a reacting to a recoding.should start with another recording.should start notification. has this be tested? What happens when used with pupil groups?

papr: I tested it only locally. The idea was to ensure that the recording would start after the fake capture had started. Resending the `recording.should_start` would trigger a `Recording already running!` info message on other pupil groups computers if these had started their recording previously. Any idea on how to fix this conflict? We cannot prevent other group members from receiving the first message.

PR #885, opened 2017-10-24
Added frame_publishing.set_format to Frame_Publisher

majoge:


papr:
@majoge Thank you very much for your contribution! It is very appreciated.

PR #886, opened 2017-10-24
Use as many threads to transcode video as there are cpus

papr:
The transcoding thread count was previously set to one. This was due to a bug on Linux where the program would crash if more threads than available cpus were used.


pupil_src/shared_modules/av_writer.py
```diff
@@ -116,7 +117,7 @@ def __init__(self, file_loc,fps=30, video_stream={'codec':'mpeg4','bit_rate': 15
         self.video_stream = self.container.add_stream(video_stream['codec'], 1/self.time_base)
         self.video_stream.bit_rate = video_stream['bit_rate']
         self.video_stream.bit_rate_tolerance = video_stream['bit_rate']/20
-        self.video_stream.thread_count = 1
+        self.video_stream.thread_count = mp.cpu_count()
```
mkassner: maybe max(1,p.cpu_count()-1)?

PR #931, opened 2017-11-14
Refactor Batch Exporter

### LINKED ISSUE #913: Batch Exporter: AssertionError when starting export twice
This is due to how the background process is implemented. The plugin does not test for running processes nor cancels them if necessary. The correct way to fix this issue is by implementing #914 

```
  File "/home/papr/work/pupil/pupil_src/launchables/player.py", line 441, in player
    p.recent_events(events)
  File "/home/papr/work/pupil/pupil_src/shared_modules/batch_exporter.py", line 190, in recent_events
    self.workers[i].start()
  File "/usr/lib/python3.5/multiprocessing/process.py", line 99, in start
    assert self._popen is None, 'cannot start a process twice'
AssertionError: cannot start a process twice
```

papr:
This PR uses #927 to fix #913.

### Behaviour

The batch export is split into 4 stages:
1. Search for available recordings
2. Queued exports
3. Active exports
4. Previous exports

- Canceling a queued export (2) puts it back to the list of available recordings (1).
- Canceling an active export (3) adds the export to the list of previous exports (4).

### Discussion
I feel like (1) and (2-4) should be split into two different menus. What do you think about that?

### DIFF
--- a/pupil_src/launchables/player.py
+++ b/pupil_src/launchables/player.py
@@ -94,7 +94,7 @@ def player(rec_dir, ipc_pub_url, ipc_sub_url,
         from offline_surface_tracker import Offline_Surface_Tracker
         # from marker_auto_trim_marks import Marker_Auto_Trim_Marks
         from fixation_detector import Offline_Fixation_Detector
-        # from batch_exporter import Batch_Exporter
+        from batch_exporter import Batch_Exporter
         from log_display import Log_Display
         from annotations import Annotation_Player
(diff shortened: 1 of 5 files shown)


papr:
Moved exporting ui into sub plugins. The order of icon display is
1. Active exports
2. Queued Exports
3. Finished Exports

All exports can be canceled at once from within the Batch Exporter or one by one from within the pugins' menus. The search for available exports remains part of the batch exporter.

PR #970, opened 2017-12-05
Improve hmd 3d calibration stability

papr:


pupil_src/launchables/eye.py
```diff
@@ -31,10 +31,13 @@ def __init__(self, is_alive, ipc_socket, eye_id, logger):
 
     def __enter__(self):
         if self.is_alive.value:
-            raise Exception("eye process already running!")
+            # indicates eye process that this is a duplicated startup
```
papr: Exceptions thrown in __enter__ do not trigger __exit__. Therefore we need an other way to communicate a duplicated startup.

papr:
Reviewed and approved for merging by @mkassner

PR #980, opened 2017-12-07
Singleeyefitter: Fix type where solve() returned a wrong variable

### LINKED ISSUE #975: Bug in singleeyefitter::solve(T,T,T,T)
This line (shared_modules/pupil_detectors/singleeyefitter/solve.h:98):

     return std::tuple<T, T, T>(-p, -p, -p);

looks like it should be:

    return std::tuple<T,T,T>(y,y,y);

(Currently `y` is calculated, then ignored.)

papr:
Fixes #975. I would like to know if this fix has an impact on the model fitter and if this might ave been the cause for instable models...


euryalus:
I did some tests on that. The case of three identical real roots in practice does not occur. Note, for a cubic ax^3+bx^2+cx+d to have three identical real roots y, one needs to have ax*3+bx^2+cx+d=(x-y)^3. As a,b,c,d for practical purposes can be assumed to be random, this essentially never happens. So this change in the code base, will have no effect on model stability.

PR #984, opened 2017-12-14
3d marker pose

marc-tonsen:
Added a function to compute the 3D pose of a calibration marker given it's 2D description


pupil_src/shared_modules/circle_detector.py
```diff
@@ -464,6 +465,46 @@ def get_nested_clusters(contours, hierarchy, min_nested_count):
...
+    return pts_rot
+
+def marker_3d_pose(marker, cam_model, marker_diameter=7.6):
+
+    target_circle = [[0, 0], [marker_diameter, marker_diameter], 0]
+    target_pts = getEllipsePts(target_circle)
+    target_pts3D = np.zeros((target_pts.shape[0], target_pts.shape[1] + 1))
+    target_pts3D[:, :-1] = target_pts
+    target_pts3D.shape = -1, 1, 3
+    target_pts3D = target_pts3D.astype('float32')
```
papr: The type could be defined on creation (np.zeros). This way you save one copy.

pupil_src/shared_modules/circle_detector.py
```diff
@@ -464,6 +465,46 @@ def get_nested_clusters(contours, hierarchy, min_nested_count):
             clusters[cluster[1]] = cluster
     return clusters.values()
 
+def getEllipsePts(e, num_pts=10):
```
mkassner: can we not use this here? https://docs.opencv.org/3.0-beta/modules/imgproc/doc/drawing_functions.html#ellipse2poly

pupil_src/shared_modules/circle_detector.py
marc-tonsen: ok, I changed that!

pupil_src/shared_modules/circle_detector.py
mkassner: We now know this fn sucks.

PR #1009, opened 2018-01-09
Opt-out online pupil detection

papr:
Feature idea by @anligert. Online pupil detection is not required if one plans to use offline pupil detection anyway. Select `disable` as `detection & mapping mode` in the general settings to turn off pupil detection and gaze mapping. Do not forget to enable the `Record eye` option in the recorder settings if you do not use online pupil detection.

### DIFF
--- a/pupil_src/launchables/eye.py
+++ b/pupil_src/launchables/eye.py
@@ -127,9 +127,10 @@ def eye(timebase, is_alive_flag, ipc_pub_url, ipc_sub_url, ipc_push_url,
         from video_capture import manager_classes
 
         # Pupil detectors
-        from pupil_detectors import Detector_2D, Detector_3D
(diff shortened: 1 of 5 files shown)


performlabrit:
I like this idea, but also liek the idea of adding an option to pupil player to remove the evidence of online calibration.   This way, I can use the resource-light 2D pupil detection/calibration to have a vague idea of the behavior while I'm recording data, but can later use offline methods to fine-tune   3D pupil detection and natural feature detection.  The ability to reset the data stored from the online routines will reduce loading times, and prevent accidental use of a poor 2D calibration.

papr:
I do not like the idea of having a button that deletes recorded data. Deleting data should only possible for reproducible data. If people did not record the eye they will not be able to use offline detection.

We are currently looking into possible long-term solutions to the loading issue. One idea would be on-demand data serialization/loading.

Issue #982, opened 2017-12-11
How to translate or map Pupil player timestamps to realtime 


jooni41:
I want to map or translate pupil player timestamps to real time for synchronisation purposes. what would be the most appropriate way

cpicanco:
I am afraid it is not possible, unless you have annotated the (real) start time of your target Pupil recording. If you have the start time, then you can convert any pupil player timestamps to real time in two steps. First convert pupil timestamps from Epoch start to zero start (by subtracting all timestamps from the first one). Then add your annotated real time to all timestamps. Hope this helps.

PR #1017, opened 2018-01-15
Recorder: Add unix and synced recording start times to info.csv

### LINKED ISSUE #989: Request - Save "Start Time" in "Info.csv" with floating point accuracy
At the moment the "Start Time" value saved in the "info.csv file is accurate down to seconds (e.g, Start Time, 11:12:55). 

It would be very helpful if the accuracy of this value was increased, so that start time is labeled with fractions of seconds (e.g, Start Time, 11:12:55.12345). This would help synchronize data recorded from pupil with data recorded from other programs on the same computer.

papr:
The unix timestamp corresponds to the Python system timebase, the synced timestamp corresponds to the synced timebased used for all data points in the recording.

Fixes #989, #982

### DIFF
--- a/pupil_src/shared_modules/recorder.py
+++ b/pupil_src/shared_modules/recorder.py
@@ -219,6 +219,7 @@ class Recorder(System_Plugin_Base):
         self.running = True
         self.menu.read_only = True
         self.start_time = time()
+        start_time_synced = self.g_pool.get_timestamp()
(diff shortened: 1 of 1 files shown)


jonmatthis:
Thank you for making this change! Do you know when it will be included in the Linux/Mac bundles?

PR #1019, opened 2018-01-15
Capture/Player: Collapsed main menu by default

papr:
The idea is to collapse the main menu by default. We were not able to do so before the introduction of the icon bar since the collapse icon was not obvious to the user. It is more obvious that the icons allow interaction. Having the menu collapsed initially shows the user that the main menu can be collapsed. This is important for those users that used to drag the menu to collapse it. This is not possible anymore with the introduction of the minimum menu width.


willpatera:
@papr I am fine with this change. @mkassner @papr do you think the eye window menu should also start out as collapsed so that we have consistent behavior?

papr:
Actually it is not that easy to implement the same behavior for the eye windows since the pupil detectors are launched after the initialization and expand the main menu automatically.

PR #1072, opened 2018-02-12
v1.4 feature collection

papr:
- Fixation detector: Jump to previous fixation button and wrap-around seeking
- Avoid GLFW window creation failures where passed window size dimension is zero
- Trimmed section looped playback
- Consistent export range handling in all built-in plugins
- Export folder names are named by trim section time instead of index range

### DIFF
--- a/pupil_src/launchables/world.py
+++ b/pupil_src/launchables/world.py
@@ -307,12 +307,7 @@ def world(timebase, eyes_are_alive, ipc_pub_url, ipc_sub_url,
 
         width, height = 1280, 720
         width += icon_bar_width
-        default = width, height
-
-        width, height = session_settings.get('window_size', default)
-
-        if 0 in (width, height):  # Avoid glfw window creation error
-            width, height = default
+        width, height = session_settings.get('window_size', (width, height))
 
         # window and gl setup
         glfw.glfwInit()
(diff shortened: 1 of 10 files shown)


pupil_src/launchables/world.py
```diff
@@ -305,9 +305,17 @@ def handle_notifications(n):
                                         'actor': p.class_name,
                                         'doc': p.on_notify.__doc__})
 
+        width, height = 1280, 720
+        width += icon_bar_width
+        default = width, height
+
+        width, height = session_settings.get('window_size', default)
```
mkassner: I dont like this kind of fix. Can we not rather catch this when the settings are written to file?

papr: Agreed! Will change tomorrow.

pupil_src/launchables/world.py
```diff
@@ -305,9 +305,12 @@ def handle_notifications(n):
                                         'actor': p.class_name,
                                         'doc': p.on_notify.__doc__})
 
+        width, height = 1280, 720
+        width += icon_bar_width
+        width, height = session_settings.get('window_size', (width, height))
```
mkassner: I get this is cleaner but was the old version not also easy to read? and its more compact.

Issue #1083, opened 2018-02-19
Frame Publisher crashes when Fake Capture for World Video is used


mattem86:
2018-02-19 11:33:17,080 - world - [ERROR] launchables.world: Process Capture crashed with trace:
Traceback (most recent call last):
  File "launchables/world.py", line 467, in world
  File "shared_modules/zmq_tools.py", line 140, in send
  File "zmq/backend/cython/socket.pyx", line 636, in zmq.backend.cython.socket.Socket.send (zmq/backend/cython/socket.c:7314)
  File "zmq/backend/cython/socket.pyx", line 683, in zmq.backend.cython.socket.Socket.send (zmq/backend/cython/socket.c:7057)
  File "zmq/backend/cython/socket.pyx", line 188, in zmq.backend.cython.socket._send_copy (zmq/backend/cython/socket.c:2824)
  File "zmq/utils/buffers.pxd", line 200, in zmq.utils.buffers.asbuffer_r (zmq/backend/cython/socket.c:8885)
  File "zmq/utils/buffers.pxd", line 151, in zmq.utils.buffers.asbuffer (zmq/backend/cython/socket.c:8211)
TypeError: None does not provide a buffer interface.

This issue is on Windows and Mac, Pupil Version 1.3 and 1.4 affected (lower not tested)

papr:
Thank you for the report. The issue is that the default frame publisher format is set to jpeg and Fake_Source does not support yuv or jpeg buffers. For now, there two possible work-arounds:
1. Start the frame publisher plugin and set its format to `BGR` before activating the Fake Capture.
2. Start the frame publisher plugin via a the `open_plugin` and pass the `BGR` format as an argument.

The frame publisher has logic in place to recognize unsupported formats. The problem is that did not trigger in the case of Fake Capture.

PR #1085, opened 2018-02-19
Video source fixes

papr:
- Fix `Fake_Source` compatibility with Frame Publisher (#1083)
~- Remove necessity for video sources to define `jpeg_buffer`/`yuv_buffer` attributes~
- Fix an issue where `Fake_Source` would return frames at 2 fps in Capture if it was restored from the session settings.

### DIFF
--- a/pupil_src/shared_modules/video_capture/fake_backend.py
+++ b/pupil_src/shared_modules/video_capture/fake_backend.py
@@ -34,6 +34,8 @@ class Frame(object):
         self._gray = None
         self.index = index
         # indicate that the frame does not have a native yuv or jpeg buffer
+        self.yuv_buffer = None
+        self.jpeg_buffer = None
 
     @property
     def img(self):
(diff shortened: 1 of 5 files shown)


pupil_src/shared_modules/video_capture/fake_backend.py
```diff
@@ -34,8 +34,6 @@ def __init__(self, timestamp, img, index):
         self._gray = None
         self.index = index
         # indicate that the frame does not have a native yuv or jpeg buffer
-        self.yuv_buffer = None
```
mkassner: why not make these required form the frame api? Then you dont have to check for attributes...

papr: Frame publisher's incompatible format detection is based on AttributeErrors. We need to check if this attribute is `None` on multiple occasions anyway. My thought was that we could change these to an existence test of the attribute and keep the simple fail logic in frame publisher. Alternatively, we always define `yuv_buffer`, need to check multiple times for `None` and increase frame publishers fail logic complexity.

Issue #1052, opened 2018-02-02
do not filter based on confidence in gaze mappers


mkassner:
just map compute gaze confidence and filter this in consuming plugins.

performlabrit:
I'll bump this.  Most eye trackers behave this way - attempt to detect the pupil regardless of the quality, and spit the data out along with its confidence metric and perhaps a blink flag.  In its current setup, the pupil tracker is effectively throwing away data.

cpicanco:
Oh my gosh. That is way my confidence comparisons were indifferent. I will need to redo the job.

mkassner:
Just to clarify. We are reporting pupil detection results for each eye frame. We just have not been mapping low confidence pupil data to gaze. This is what this issue addresses.

mkassner:
But yes. I think this is a good improvement. We will implement and test.

mkassner:
this is related: #1047

cpicanco:
@mkassner , When you say `gaze mapping` does it includes exports or just visualizations?

cpicanco:
@papr Can you clarify this issue a little bit? Whe the boss said that "We just have not been mapping low confidence pupil data to gaze." Does it includes exports or not?

papr:
@cpicanco He was referring to the fact that all binocular mappers dismiss pupil datums with less than 0.6 confidence to avoid bad mappings due to low confidence data. The solution would be to map this low confidence data in a monocular fashion instead of dismissing it.

cpicanco:
So far I am understanding that binocular mappers will dismiss low confidence data (hence, dismissed data cannot be exported).

I have one simple question:

- Lets say your data have 100 items, 90 above, 10 below the confidence threshold. Is it possible to map (calibrate) using only these 90 points but apply this mapping to all points?

papr:
Yes, I am implementing this right now.

Issue #1047, opened 2018-01-31
Player:  When writing out raw data, ignore confidence threshold (or allow a custom)


performlabrit:
Raw data should include *all* data, and allow the user to filter post-hoc.

papr:
@performlabrit You can set a custom confidence threshold in the Pupil Player general settings.

cpicanco:
@papr , I am afraid that a proper description of what is going on in the exporting process is lacking in the docs. If you export data from a surface, all data will be included and confidence will be reported:

https://github.com/pupil-labs/pupil/blob/13535b753e3ac0fc85c44281d9ce6645436ab362/pupil_src/shared_modules/offline_surface_tracker.py#L476-L488

However, heatmap data will **not** include all the exported data:

https://github.com/pupil-labs/pupil/blob/ceaa3d26d547d1f156e4fd9ca93b37611ef2e2d8/pupil_src/shared_modules/offline_reference_surface.py#L153-L154

Nor fixations:
https://github.com/pupil-labs/pupil/blob/ceaa3d26d547d1f156e4fd9ca93b37611ef2e2d8/pupil_src/shared_modules/fixation_detector.py#L309-L310

Nor any other visualization. So for **visualizations** and **heatmaps** adjusting the confidence has an effect. 

World gaze exports are **NOT** being filtered by the confidence threshold. All data is exported:

https://github.com/pupil-labs/pupil/blob/ceaa3d26d547d1f156e4fd9ca93b37611ef2e2d8/pupil_src/shared_modules/player_methods.py#L28-L68


Hope this helps.

papr:
@cpicanco Could you elaborate on what you meant with this note?
> note the logical inconsistency in the filter

cpicanco:
@papr , I think both should be `>=`

> pupil/pupil_src/shared_modules/offline_reference_surface.py
`if gp['confidence'] >= self.g_pool.min_data_confidence`

> pupil/pupil_src/shared_modules/fixation_detector.py
`if gp['confidence'] > self.g_pool.min_data_confidence`

PR #1100, opened 2018-02-26
Confidence threshold consistency

papr:
1. Previously, all binocular mappers dismissed pupil data with confidence lower than `0.6` to prevent bad binocular mapping. Now, low confidence pupil data is mapped monocularly.
2. Introduction of a user-settable global **minimum calibration confidence** threshold: A threshold that lower bounds the pupil data confidence that is used for calibration. This value was previously fixed to `0.8` in Pupil Capture and to `0.0` in Pupil Player -- effectively using low confidence pupil data for calibration.

These changes were introduced based on #1052 and #1047.

**Minimum data confidence**: Lower bounds _gaze_ data by confidence that is used to generate visualizations and high level data, e.g. fixations. This does not effect the raw data export. Following data points are effected by it:
- Fixations
- Offline surface heatmaps
- Circle, cross, light points and polyline visualizations

### DIFF
--- a/pupil_src/launchables/player.py
+++ b/pupil_src/launchables/player.py
@@ -252,6 +252,7 @@ def player(rec_dir, ipc_pub_url, ipc_sub_url,
         g_pool.rec_dir = rec_dir
         g_pool.meta_info = meta_info
         g_pool.min_data_confidence = session_settings.get('min_data_confidence', 0.6)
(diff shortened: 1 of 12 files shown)


cpicanco:
Now everything makes sense. Luckly enough, I was manually removing low confidence calibration samples from Pupil Player Offline calibration using the natural features functionality. From now on (v1.5), people should achieve similar (or better) calibration results when fine tuning calibration with pupil player.

PR #1104, opened 2018-03-01
Workaround for minimize crash

pointcontrols:


pupil_src/launchables/world.py
```diff
@@ -220,6 +220,8 @@ def on_resize(window, w, h):
             nonlocal window_size
             nonlocal camera_render_size
             nonlocal hdpi_factor
+            if h == 0:
```
mkassner: maybe also test for w?

PR #1108, opened 2018-03-02
Proof of concept - Fingertip calibration

ChingT:
The steps for fingertip calibration:
1. Covering the rectangles on the screen with your index finger.
2. Look at your fingertip and then start calibration!


pupil_src/shared_modules/calibration_routines/fingertip_calibration.py
```diff
@@ -0,0 +1,139 @@
...
+                ref['norm_pos'] = self.pos
+                ref['screen_pos'] = marker_pos
+                ref['timestamp'] = frame.timestamp
+                self.ref_list.append(ref)
+            else:
+                self.pos = None  # indicate that no reference is detected
+
+            # Always save pupil positions
+            for p_pt in recent_pupil_positions:
+                if p_pt['confidence'] > self.pupil_confidence_threshold:
```
mkassner: this is no longer required in calibration routines. Please remove the confidence threshold check.

pupil_src/shared_modules/calibration_routines/fingertip_calibration.py
```diff
@@ -0,0 +1,139 @@
...
+        self.fingertip_tracker = FingertipTracker()
+        self.fingertip = None
+        self.press_key = 0
+
+        self.menu = None
+
+    def init_ui(self):
+        super().init_ui()
+        self.menu.label = 'Fingertip Calibration'
+        self.menu.append(Info_Text('Calibrate gaze parameters using finger'))
```
mkassner: also add:

self.menu.append(Info_Text('This is a experimental calibration routine. Move you finger in your field of view while looking at the finger tip. This plugin is for prototyping and experimentation only. The detection robustness is not production grade. We will put a lot more effort into this going forward but wanted to release the idea and hope for feedback!'))

willpatera:
@ChingT just tested this out. It is working well for me. In future revisions it would be nice to visualize the fingertip detector with a cyan rectangle and a cyan dot (for example) so that the fingertip detector visualization looks explicitly different from the gaze points shown in the world window. This is only relevant for a debugging standpoint or for an operator and not for the participant who is self calibrating. 

Furthermore, I would suggest adding a bit more information in info text about the two step process (1) calibrating for skin tone of the participant (2) collecting calibration samples.

ChingT:
@willpatera Thank you for the comments. I've modified the fingertip detector visualization. There is a rectangle showing the detected finger and a dot on the detected fingertip.
> "The rectangles that are used to calibrate skin tone are difficult to see when the are on the right hand side under the open plugin menu." 

--> How about once starting calibration, the plugin menu collapses automatically?

> "Users may also need to know how to hold their hand as fingertip could be considered as "fingernail" or "fingerprint" side of the finger." 

--> I think we may need to take an instruction video for that.

ChingT:
@willpatera Thank you for the comments. I've modified the fingertip detector visualization. There is a rectangle showing the detected finger and a dot on the detected fingertip.
> "The rectangles that are used to calibrate skin tone are difficult to see when the are on the right hand side under the open plugin menu." 

--> How about once starting calibration, the plugin menu collapses automatically?

> "Users may also need to know how to hold their hand as fingertip could be considered as "fingernail" or "fingerprint" side of the finger." 

--> I think we may need to take an instruction video for that.

willpatera:
@ChingT The changes look good. We will certainly need documentation for this issue. I have made an issue in pupil-docs for this here: https://github.com/pupil-labs/pupil-docs/issues/188

PR #1110, opened 2018-03-05
Audio capture support on Windows

pointcontrols:
1. Use pyaudio to list capture devices
2. Fix stream finalization issue
3. Set output sampling rate equal to input rate
4. Simpler output timestamps calculation

### DIFF
--- a/pupil_src/shared_modules/audio_capture.py
+++ b/pupil_src/shared_modules/audio_capture.py
@@ -152,6 +152,8 @@ class Audio_Capture(Plugin):
                 in_container = av.open('none:{}'.format(audio_src), format="avfoundation")
             elif platform.system() == "Linux":
                 in_container = av.open('hw:{}'.format(audio_src), format="alsa")
+            elif platform.system() == "Windows":
+                in_container = av.open('audio={}'.format(audio_src), format="dshow", options={'audio_buffer_size':'23'})
             else:
                 raise av.AVError('Platform does not support audio capture.')
         except av.AVError as err:
@@ -172,7 +174,6 @@ class Audio_Capture(Plugin):
         out_container = None
         out_stream = None
         timestamps = None
-        out_frame_num = 0
         in_frame_size = 0
 
         stream_epoch = in_stream.start_time * in_stream.time_base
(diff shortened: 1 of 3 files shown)


pupil_src/shared_modules/audio_capture.py
```diff
@@ -152,6 +152,8 @@ def capture_thread(self, audio_src, running, recording):
                 in_container = av.open('none:{}'.format(audio_src), format="avfoundation")
             elif platform.system() == "Linux":
                 in_container = av.open('hw:{}'.format(audio_src), format="alsa")
+            elif platform.system() == "Windows":
+                in_container = av.open('audio={}'.format(audio_src), format="dshow", options={'audio_buffer_size':'23'})
```
papr: What is the reasoning to set this value to `23`? Is it possible that this value is device-dependent?

pupil_src/shared_modules/audio_capture.py
```diff
@@ -196,22 +197,28 @@ def close_recording():
...
-                old_ts_idx = np.arange(0, len(timestamps) * in_frame_size, in_frame_size) / in_frame_rate
                 new_ts_idx = np.arange(0, out_frame_num * out_frame_size, out_frame_size) / out_frame_rate
-                interpolate = interp1d(old_ts_idx, timestamps, bounds_error=False, fill_value='extrapolate')
-                new_ts = interpolate(new_ts_idx)
+                if in_frame_rate != out_frame_rate:
+                    old_ts_idx = np.arange(0, len(timestamps) * in_frame_size, in_frame_size) / in_frame_rate
+                    interpolate = interp1d(old_ts_idx, timestamps, bounds_error=False, fill_value='extrapolate')
+                    new_ts = interpolate(new_ts_idx)
+                else:
+                    new_ts = timestamps[0] + new_ts_idx
```
papr: Do I understand it correctly that you are adding up a timestamp and an index here?

pupil_src/shared_modules/audio_capture.py
pointcontrols: No it's not device depdendent. It's the size of the capture buffer in miliseconds.

pupil_src/shared_modules/audio_capture.py
pointcontrols: Adding a fixed offset (stream epoch) to the generated sequence of output timestamps (numpy array)

papr: I am still a bit confused about this line. `timestamps` is a Python list that contains floats, if I am not mistaken. And `new_ts_idx` is a numpy array. Is it possible that you meant this line to be `new_ts = np.array(timestamps) + new_ts_idx[0]`?

papr: Never mind, I just understood it. 🤦‍♂️

pointcontrols: No. new_ts_idx contains the exact timestamps calculated with respect to the output frame size, zero based. We just need to add the correct stream epoch which is in fact the value in timestamps[0].

mkassner:
@pointcontrols  is this ready to merge?

pointcontrols:
@mkassner Yes.

PR #1111, opened 2018-03-05
Seek control changes for real-time playback

pointcontrols:


pupil_src/shared_modules/video_capture/base_backend.py
```diff
@@ -199,10 +199,26 @@ def replace_backend_manager(manager_class):
...
     allowed_speeds = [.25, .5, 1., 1.5, 2., 4.]
 
-    def __init__(self, g_pool, timed_playback=False, playback_speed=1., *args, **kwargs):
+    def __init__(self, g_pool, stand_alone=True, timed_playback=False, playback_speed=1., *args, **kwargs):
+        '''
+        Possible configurations:
+                                    stand_alone
+                             |   True    | False
+                       ------+-----------+--------
+                        True |  Capture  |   /
```
mkassner: why not make one var:

playback_speed

possible values: 
a) a float for actual playback speed
b) "external" for use in player
c) "free_running"

pupil_src/shared_modules/video_capture/base_backend.py
```diff
@@ -199,10 +199,26 @@ def replace_backend_manager(manager_class):
...
+
+
+        Capture: Fake/File Source is used as UVC Source replecement.
+        Player: Seek Control provides
+        '''
         super().__init__(g_pool)
-        self.playback_speed = playback_speed
+        self.stand_alone = stand_alone
         self.timed_playback = timed_playback
+        assert stand_alone or not timed_playback, 'Invalid configuration'
```
mkassner: this stuff is why you may want one value :-)

pupil_src/shared_modules/video_capture/base_backend.py
papr: You could argue that there is no need to change playback speed for `Playback_Source`'s in Capture. Therefore we could move the `playback_speed` field to `Seek_Control` and merge `stand_alone` + `timed_playback` into one `timing` (?) field that allows one of the following values:
- `'external`, currently `stand_alone=False, timed_playback=False`, use case: Player
- `'own`, currently  `stand_alone=True, timed_playback=True`, use case: Capture source
- `None`

mkassner: I like this. lets do it like this.

pupil_src/launchables/player.py
```diff
@@ -387,14 +387,14 @@ def set_window_size():
                            ('System_Timelines', {}),
                            ('Video_Export_Launcher', {}),
                            ('Pupil_From_Recording', {}),
-                           ('Gaze_From_Recording', {})]
+                           ('Gaze_From_Recording', {}),
+                           ('Audio_Playback', {})]
 
-        g_pool.plugins = Plugin_List(g_pool, session_settings.get('loaded_plugins', default_plugins))
+        plugins = session_settings.get('loaded_plugins', default_plugins)
+        g_pool.plugins = Plugin_List(g_pool, plugins)
```
papr: Is this change to improve readability or just leftovers from debugging the contents of `plugins`?

pupil_src/shared_modules/video_capture/base_backend.py
mkassner: @papr  can we still add this sugenstion? I like you solution

pupil_src/launchables/player.py
```diff
@@ -515,12 +508,16 @@ def handle_notifications(n):
...
                 glfw.glfwSwapBuffers(main_window)
 
-            # present frames at appropriate speed
-            g_pool.capture.wait(frame)
             glfw.glfwPollEvents()
 
+        # Remove manually to avoid being included in `loaded_plugins`
+        # and `deinit_ui()` call on `g_pool.plugins.clean()`
+        g_pool.plugins._plugins.remove(g_pool.capture)
+
```
mkassner: is the proper way not to raise a notimplemented error in init_gui?

papr: The important part is the second line of the comment. `deinit_ui()` causes an exception

papr: Ah I misread your comment. Calling deinit_ui() would be legit if init_ui() would have been called before. But we do not do that in Player to avoid having the plugin icon in the side menu. Either we add a switch-case into init_ui and handle it as a normal plugin, or we handle it as a special plugin. But we need to init and add it manually either way. Therefore I would argue to keep this special case.

mkassner: @papr  since the video_playback source is a plugin whats the harm in having a menu. We could add some info like number of frames, current frame, resolution camera name etc...

mkassner: the more we unify the behaviour the better /

PR #1164, opened 2018-04-27
bundle.sh: Format version based on most recent maj.min tag

papr:
Instead of simply using the most recent tag that could also be maj.min.patch.

### DIFF
--- a/deployment/bundle.sh
+++ b/deployment/bundle.sh
@@ -1,8 +1,15 @@
 #!/bin/bash
 
-# create release dir from latest tag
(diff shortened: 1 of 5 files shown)


willpatera:
@papr This works, has been tested on on the 16.04 image, and can be merged. However, now there is a mismatch between the name of the bundle directory and the name of the actual bundles. We need to also update `finalize_bundle.py` scripts the `deploy_*` directories. 

Currently the packages look like this:

```
pupil_capture_linux_os_x64_v1.7.9.13.deb
```

And the directory looks like this

```
pupil_v1.7.22_linux_x64
```

PR #1173, opened 2018-05-09
Add switch for Check Stripes in the menu

ChingT:
The switch would only appear in Pupil Cam2 menu

### DIFF
--- a/pupil_src/shared_modules/video_capture/uvc_backend.py
+++ b/pupil_src/shared_modules/video_capture/uvc_backend.py
@@ -327,8 +327,6 @@ class UVC_Source(Base_Source):
 
         if self.check_stripes and ("Pupil Cam2" in self.uvc_capture.name):
             self.checkframestripes = Check_Frame_Stripes()
-        else:
-            self.checkframestripes = None
 
     @property
     def frame_rate(self):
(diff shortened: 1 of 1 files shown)


pupil_src/shared_modules/video_capture/uvc_backend.py
```diff
@@ -450,6 +462,18 @@ def frame_rate_getter():
...
+            self.check_stripes = check_stripes
+            if self.check_stripes:
+                self.checkframestripes = Check_Frame_Stripes()
+                logger.info("Check Stripes for camera {} is now on".format(self.uvc_capture.name))
+            else:
+                self.checkframestripes = None
+                logger.info("Check Stripes for camera {} is now off".format(self.uvc_capture.name))
+
+        if ("Pupil Cam2" in self.uvc_capture.name):
+            ui_elements.append(ui.Switch('check_stripes', self, setter=set_check_stripes, on_val=1, off_val=0, label="Check Stripes"))
```
mkassner: on val and off val should be able to use just the defaults I think.

pupil_src/shared_modules/video_capture/uvc_backend.py
```diff
@@ -450,6 +462,18 @@ def frame_rate_getter():
             ui_elements.append(image_processing)
         ui_elements.append(ui.Button("refresh",gui_update_from_device))
         ui_elements.append(ui.Button("load defaults",gui_load_defaults))
+
```
mkassner: all of this could be inside the if below.

papr:
I am confused why Github is showing all these other changes that are already part of pupil-labs/master...

PR #1187, opened 2018-05-18
Coarse detection implemented in C++

mattem86:
I disabled the coarse pupil detection in Python and implemented a similar approach in C++. Please review the performance and accuracy. The main parameters are minimum coverage of the pupil in a ROI frame, the min and max estimated pupil radii. A user specified ROI is taken into account.

Please test on Windows and Linux as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,18 @@ hmd_cal_data
 
 mprofile_*.dat
 .ipynb_checkpoints/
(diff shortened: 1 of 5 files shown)


mkassner:
Thank you for the contribution! However, we actually don't use coarse detection anymore in most cases. I ll closes this therefore.

PR #1240, opened 2018-07-27
[uvc_backend] Remove "load defaults" button of camera setting

ChingT:
1. Remove "load defaults" button
2. Add exception for 'Absolute Exposure Time' KeyError


mkassner:
There is a lot more in here then removing a button. I dont see the reason for the other changes. removing saturation I dont agree with the others I dont understand.

ChingT:
@mkassner There are two same line "try:controls_dict['Saturation'].value = 0", so I remove one.
Adding exception for 'Absolute Exposure Time' KeyError is because if the camera has no key 'Absolute Exposure Time', then there would be an exception, which I had with one of the cameras before.

PR #1243, opened 2018-08-01
Enabled recording dir to be set via notifcation

aprowe:
Added line of comments
Added notifcation hook for setting `rec_root_dir` of `Recorder`

The code seemed set up for this, yet did not support it


papr:
Hey @aprowe 
Thank you for your contribution! 

As the documentation above your change states, one can use the session name to overwrite the `rec_root_dir` if the session name starts with `/`. Therefore it looks like your PR would introduce an implementation for an already existing feature. Or am I missing a different use case that was not covered before?

aprowe:
Ah ok I didn't know about that. This should be closed then. 
Thanks for the feedback!

PR #1262, opened 2018-08-10
Improved docs and naming of a couple files and exported values.

marc-tonsen:
Several minor changes:

Offline Surface Tracker Data Format: Changed naming of file `srf_positons_<surface_name>_<surface_id>.csv` to `surface_positons_<surface_name>_<surface_id>.csv` to match the naming of the other files. The acronym `srf` is not used anywhere else.

Also changed the naming of the contained values `m_to_screen` and `m_from_screen` to `m_to_surface` and `m_from_surface`. The word screen is only accurate in a subset of all use cases.

Reduced the surface id from a long float number to a 4 digit integer, which should easily be sufficient while being a lot more manageable.

Minor correction/improvement to formulation to the `pupil_gaze_positions_info.txt` file.

### DIFF
--- a/pupil_src/shared_modules/offline_surface_tracker.py
+++ b/pupil_src/shared_modules/offline_surface_tracker.py
@@ -471,9 +471,9 @@ class Offline_Surface_Tracker(Surface_Tracker, Analysis_Plugin_Base):
             surface_name = '_'+s.name.replace('/','')+'_'+s.uid
 
             #save surface_positions as csv
-            with open(os.path.join(metrics_dir,'srf_positons'+surface_name+'.csv'),'w',encoding='utf-8',newline='') as csvfile:
+            with open(os.path.join(metrics_dir,'surface_positons'+surface_name+'.csv'),'w',encoding='utf-8',newline='') as csvfile:
                 csv_writer =csv.writer(csvfile, delimiter=',')
-                csv_writer.writerow(('frame_idx','timestamp','m_to_screen','m_from_screen','detected_markers'))
+                csv_writer.writerow(('frame_idx','timestamp','m_to_surface','m_from_surface','detected_markers'))
                 for idx,ts,ref_srf_data in zip(range(len(self.g_pool.timestamps)),self.g_pool.timestamps,s.cache):
                     if in_mark <= idx < out_mark:
                         if ref_srf_data is not None and ref_srf_data is not False:
--- a/pupil_src/shared_modules/raw_data_exporter.py
+++ b/pupil_src/shared_modules/raw_data_exporter.py
@@ -30,7 +30,7 @@ class Raw_Data_Exporter(Analysis_Plugin_Base):
     keys:
         timestamp - timestamp of the source image frame
         index - associated_frame: closest world video frame
-        id - 0 or 1 for left/right eye
+        id - 0 or 1 for right and left eye (from wearer's POV)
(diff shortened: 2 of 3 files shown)


pupil_src/shared_modules/offline_surface_tracker.py
```diff
@@ -471,9 +471,9 @@ def save_surface_statsics_to_file(self, export_range, export_dir):
             surface_name = '_'+s.name.replace('/','')+'_'+s.uid
 
             #save surface_positions as csv
-            with open(os.path.join(metrics_dir,'srf_positons'+surface_name+'.csv'),'w',encoding='utf-8',newline='') as csvfile:
+            with open(os.path.join(metrics_dir,'surface_positons'+surface_name+'.csv'),'w',encoding='utf-8',newline='') as csvfile:
                 csv_writer =csv.writer(csvfile, delimiter=',')
-                csv_writer.writerow(('frame_idx','timestamp','m_to_screen','m_from_screen','detected_markers'))
```
papr: Are you sure that `m_to_screen` does not mean `surfce to original frame`? In this case the rename would be the other way around.

pupil_src/shared_modules/raw_data_exporter.py
```diff
@@ -30,7 +30,7 @@ class Raw_Data_Exporter(Analysis_Plugin_Base):
     keys:
         timestamp - timestamp of the source image frame
         index - associated_frame: closest world video frame
-        id - 0 or 1 for left/right eye
+        id - 0 or 1 for right and left eye (from wearer's POV)
```
papr: I would write out `point of view`

pupil_src/shared_modules/reference_surface.py
```diff
@@ -84,7 +85,7 @@ def __init__(self, g_pool, name="unnamed", saved_definition=None):
         self.camera_pose_3d = None
         self.use_distortion = True
 
-        self.uid = str(time())
+        self.uid = "{:04}".format(random.randint(0,9999))
```
papr: Nice use of zero-padding!

pupil_src/shared_modules/offline_surface_tracker.py
marc-tonsen: My current understanding is that the matrix `m_to_screen` transforms points in scene image coordinates into points in normalized surface coordinates. Cant't say that I am sure though, I am still investigating the code!

pupil_src/shared_modules/reference_surface.py
mkassner: please replace this with a proper uuid or time.time with stripped '.'

marc-tonsen: The idea behind this change was to make sure that generated file names, which contain the `uid`, are more manageable. My thinking was that a for digit random number should be more than enough in this case while being very manageable. Would you still prefer the longish but more unique timestamp?

papr: I think it would be worth it to check out the uuid python module and see if it contains anything helpful for this case.

pupil_src/shared_modules/offline_surface_tracker.py
marc-tonsen: I can now confidently state that `m_to_screen` is mapping from surface coordinates to normalized image coordinates. I'd now propose to change the naming in accordance to what I use in #1268 : `surf_to_img_trans` or something similarly explicit.

papr:
Is this PR still relevant in regard to #1268? What about https://github.com/pupil-labs/pupil/pull/1262/files#r209469129 ?

marc-tonsen:
Only the changes to `raw_data_exporter.py` are still relevant. There is another change I wanted to propose in this file though to make it more consistent with the exports in the surface tracker. I will close this PR and create a new one with these changes.

Issue #1261, opened 2018-08-10
Load surface names and sizes in Player


marc-tonsen:
When opening a recording with pre-existing surface definitions in Player the surface names and sizes are not loaded but are initialized with default values. These values should be saved and loaded.

papr:
Is this being fixed through #1268 ?

marc-tonsen:
Yes, I think it should be. In Capture this is already solved through #1268 , I'll check if changes to Player are necessary for this to carry over!

papr:
Fixed in #1268

Issue #1260, opened 2018-08-09
Does surface tracker take account of camera distortion?  


OscartGiles:
We're trying to figure out where someone is looking on a screen in pixel coordinates. To do this we are using the surface tracker which returns a 'norm_pos'. 

Does this currently take account of the camera's fish eye distortion? I can't see any correction for it in the source code? 

All the best,
Oscar

marc-tonsen:
Hi Oscar! The surface tracker does currently take distortion into account only partially. Distortion is considered e.g. in the localisation of the surface corners ([here](https://github.com/pupil-labs/pupil/blob/master/pupil_src/shared_modules/reference_surface.py#L328) in the code), but the mapping of the gaze point into the located surface does not compensate for the distortion. Therefore there can be an error in the gaze point in surface coordinates, if the surface is in a highly distorted region of the camera's FOV.
We ran into a few problems when trying to compensate distortion also in the gaze mapping, but this is certainly still on our todo-list!

marc-tonsen:
I started fixing this issue here #1268 . The current version of the PR should now correctly compensate for camera distortion when mapping gaze onto a surface, but it requires more testing. The current version of the surface tracking code is quite messy and I will take this opportunity to clean it up!

PanBartosz:
Hello! If I understand correctly this fix will enable to precisely track gaze on the flat surface (i.e.) screen. I tried to implement simple gaze tracking on the screen in following manner:
1. Register surface using markers in pupil capture
2. Fetch the data (norm_position) of gaze and fixation from pupil eye tracker
3. Map these values to the screen and display marker 

This workflow works currently pretty bad because there is an offset to my marker (depending on the part of the screen it can be pretty bad). My question is:

1. Is your fix going to help?
2. Are there other techniques for mapping gaze to the screen which can be used by pupil?
3. If the answer to 1) is positive, can I help to test it somehow?

marc-tonsen:
@PanBartosz Your workflow sounds correct. When you register the surface using the markers, do you edit the surface to make sure that the surface corners exactly match the display (the display panel rather than the full frame of the monitor!)? Besides that of course an accurate calibration is necessary. 

The fix is going to improve the accuracy just very slightly if the surface is in a highly distorted part of the image. In most cases the improvement will not be noticeable. 

[...]

Issue #1247, opened 2018-08-06
Surface marker are not detected until restart of surface tracker plugin


ewolf29:
System: MacOS
Pupil Capture: V1.8.12
Pupil Player: V1.8.12

After I drag a recording with surfaces into Pupil Player, the Surface Cache is not filled automatically. I have to restart the Surface Tracker Plugin to start the detection of the marker.

Reconding: https://oc.informatik.uni-wuerzburg.de/index.php/s/KyNeyRX66pKgkX5
Password: pupillabs

papr:
Hey, just to clarify, are these the steps to reproduce the issue:

1. Open recording A in Player
2. Load the offline surface tracker
-> Starts filling surface cache
3. Close Player
4. Open recording B in Player
-> offline surface tracker is loaded
-> Surface cache is not being filled

ewolf29:
@papr Yes.

papr:
Fixed in #1268

Issue #1233, opened 2018-07-19
Surface Tracker: Improve heatmap generation


papr:
The generation of surface heatmaps is slow if the surfaces have a high resolution.

### Example
Surface data taken from [demo recording](https://drive.google.com/uc?export=download&id=0Byap58sXjMVfZUhWbVRPWldEZm8).

CPU time profile excerpt: 

![player_cpu_time](https://user-images.githubusercontent.com/168390/42943061-127d9eae-8b62-11e8-9b43-91372d2715da.png)

As one can see, `cv2.GaussianBlur()` is responsible for ~35% of the loading time in this case. 

Code to improve: [`reference_surface.py#L249-L254`](https://github.com/pupil-labs/pupil/blob/master/pupil_src/shared_modules/reference_surface.py#L249-L254)

Surface details and resulting filter sizes:

Name | Surface height | Surface width | Filter height | Filter width
---| ---|---|---|----
A | 768 | 1366 | 153 | 273
B | 1400 | 800 | 281 | 161
C | 1400 | 1600 | 281 | 321
D | 1500 | 790 | 301 | 159
E | 500 | 600 | 101 | 121
F | 1200 | 800 | 241 | 161
G | 300 | 800 | 61 | 161
H | 1400 | 1600 | 281 | 321

papr:
Improved in #1268

PR #1268, opened 2018-08-14
Surfaces Revamp

marc-tonsen:
The surface tracking feature needs a lot of refactoring and a few bug fixes and enhancements.

Progress:
- [x] #1260 Compensate camera distortion correctly when mapping gaze onto the surface
- [x] Improve naming in notification API for surfaces
- [x] Refactor code to be more readable / comply with style guides
- [x] Do not block when recomputing surface location cache in offline surface tracker
- [x] Use more than a single image to define a surface in offline surface tracker
- [x] #1261 Load surface names and sizes in Player
- [x] #1247 Surface marker are not detected until restart of surface tracker plugin
- [x] #1088 Offline Surface Tracker: Recalculation only after finishing add/edit of surface; "Fixed" by making calculations in the background and cancelling/restarting this calculation upon further incoming edits
- [x] #1280 recalculate gaze on surfaces in separate thread
- [x] Improve file naming for exports
- [x] Run export calculations in a background process
- [x] #1233 Surface Tracker: Improve heatmap generation
- [x] Get rid of surface id and instead enforce unique names
- [x] UI improvements
- [x] Improved timeline visualizations
- [x] Option to freeze the image in Capture to setup surfaces easier
- [x] Enforce unique surface names


marc-tonsen:
I pushed a first more or less complete refactor of the Online Surface Tracker.

One issue with surface tracking is camera undistortion. The undistortion function that is defined by the camera intrinsics is only correct inside the bounds of the image. However parts of the surface may lie outside of the image. Given the location of the surface corner outside of the image in undistorted image space, it is impossible for us to correctly compute it's location in distorted image space for visualization. This problem also occurs inside of the image in the very corners.

In the previous version of the surface tracker this problem was not solved, i.e. the visualization was broken if one or more corners of the surface lie outside of the image.
In this version this problem is solved by saving the surface definition in both distorted and undistorted image space. For visualization the version in distorted image space is used, so we never have to undistort. For mapping gaze onto the surface the undistorted version is used to ensure a correct mapping to surface coordinates. The underlying assumption is that the gaze point will always be inside of the image, so it's undistortion is possible.

The refactored version is contained in a new plugin called `Surface Tracker Future` and the according code is contained in separate module folder in `shared_modules`. I'd like to receive a first round of feedback / tests before moving on extending this to offline surface tracking.

pupil_src/launchables/world.py
```diff
@@ -124,6 +124,7 @@ def set_detection_mapping_mode(new_mode):
         from pupil_remote import Pupil_Remote
         from pupil_groups import Pupil_Groups
         from surface_tracker import Surface_Tracker
+        from surface_tracker_future import Surface_Tracker_Future
```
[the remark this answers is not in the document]
marc-tonsen: Correct. For now I wanted to leave them in the software side by side to make it easy to switch between the two for testing. Let me know if you think I should already do the replacement.

marc-tonsen:
Currently the homographys output by the surface tracker transform normalized surface coordinates into normalized image coordinates. This makes it difficult to directly use the homographys in e.g. OpenCV to transform an image, which requires the image coordinates to be in pixels. Therefore I propose to switch to transformations between normalized surface coordinates and pixel image coordinates.

marc-tonsen:
Some of the surface related naming is currently inconsistent/not very descriptive, therefore I propose the following changes. What do you think @papr  ?

Goals of renaming:
- make "surf" the consistent acronym for surface
- in filenames always write out surface, no acronym
- remove "screen" from names, as surfaces are not necessarily screens, replace with "surf"
- Make column names consistent across exported files


Changes in Surface Event API:

```
gaze_on_srf -> gaze_on_surf
fixations_on_srf -> fixations_on_surf
m_to_screen -> surf_to_img_trans
m_from_screen -> img_to_surf_trans
```


In 'gaze_positions_on_surface_<surface_name>_<surface_id>.csv' file:
```
world_frame_idx -> world_idx
on_srf -> on_surf
```

In 'fixations_on_surface_<surface_name>_<surface_id>.csv' file:
`on_srf -> on_surf`

In 'srf_positions_<surface_name>_<surface_id>.csv' file:
```
'srf_positions_<surface_name>_<surface_id>.csv' -> 'surface_detections_<surface_name>_<surface_id>.csv'
frame_idx -> world_idx
timestamp -> world_timestamp
m_to_screen -> surf_to_img_trans
m_from_screen -> img_to_surf_trans
```

In 'surface_events.csv' file:
```
frame_number -> world_idx
timestamp -> world_timestamp
```


!!! This file is not part of the surface tracker exports !!!
```
In 'gaze_positions.csv' file:
index -> world_idx
```

!!! This file is not part of the surface tracker exports !!!
```
In 'pupil_positions.csv' file:
index -> world_idx
```

papr:
Approved.

pupil_src/shared_modules/background_helper.py
```diff
@@ -56,6 +58,14 @@ def _wrapper(self, pipe, _should_terminate_flag, generator, *args, **kwargs):
             pipe.close()
             logger.debug('Exiting _wrapper')
 
+    def _fix_background_logger(self):
```
ckbaumann: Why is this part of this PR? I thought we already reviewed this in another PR?

marc-tonsen:
Status update: 
- The PR is feature complete.
- It has gone through a first session of testing.
- All code is ready for review except `background_tasks.py`, which still needs some cleaning up.

Please go ahead with further tests and reviews!

pupil_src/shared_modules/surface_tracker/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class Heatmap_Mode(Enum):
+    WITHIN_SURFACE = "Gaze within each surface"
+    ACROSS_SURFACES = "Gaze across different surfaces"
+
+
+class _Surface_Marker_Aggregate(object):
+    """
+    A Surface Marker Aggregate is located in normalized surface space, unlike regular Markers which are
```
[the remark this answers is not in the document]
marc-tonsen: Thinking about this, this class does actually not require the marker detections to live in normalized surface space, it could be any space. I will push a version that follows your suggestion, but does not specify the coordinates space.

pupil_src/shared_modules/surface_tracker/gui.py
```diff
@@ -0,0 +1,568 @@
...
+        if self._window:
+            self.close_window()
+        else:
+            self.open_window()
+
+    def open_window(self):
+        if not self._window:
+
+            monitor = None
+            # open with same aspect ratio as surface
```
ckbaumann: With the line breaks this does not look very nice, maybe
```
surface_aspect_ratio = self.surface.real_world_size["x"] / self.surface.real_world_size["y"]
win_h = 640
win_w = int(win_h / surface_aspect_ratio)
```

pupil_src/shared_modules/surface_tracker/gui.py
```diff
@@ -0,0 +1,568 @@
...
+            gl_utils.clear_gl_screen()
+
+            # cv uses 3x3 gl uses 4x4 transformation matricies
+            width, height = self.tracker.camera_model.resolution
+            img_corners = np.array(
+                [(0, height), (width, height), (width, 0), (0, 0)], dtype=np.float32
+            )
+            denorm_trans = _get_norm_to_points_trans(img_corners)
+
+            m = self.surface.dist_img_to_surf_trans @ denorm_trans
```
ckbaumann: same `m` discussion as above

pupil_src/shared_modules/surface_tracker/gui.py
```diff
@@ -0,0 +1,568 @@
...
+            m = gl_utils.cvmat_to_glmat(m)
+
+            gl.glMatrixMode(gl.GL_PROJECTION)
+            gl.glPushMatrix()
+            gl.glLoadIdentity()
+            gl.glOrtho(0, 1, 0, 1, -1, 1)
+            gl.glMatrixMode(gl.GL_MODELVIEW)
+            gl.glPushMatrix()
+            # apply m  to our quad - this will stretch the quad such that the ref surface will span the window extends
```
ckbaumann: change this if you rename m. Please review the comment `will span the window extends` sounds not right

ckbaumann: and the line is probably too long

pupil_src/shared_modules/surface_tracker/gui.py
```diff
@@ -0,0 +1,568 @@
...
+    def gl_display_in_window(self, world_tex):
+        """
+        here we map a selected surface onto a separate window.
+        """
+        if self._window and self.surface.detected:
+            active_window = glfw.glfwGetCurrentContext()
+            glfw.glfwMakeContextCurrent(self._window)
+            gl_utils.clear_gl_screen()
+
+            # cv uses 3x3 gl uses 4x4 transformation matricies
```
ckbaumann: matrices

pupil_src/shared_modules/surface_tracker/gui.py
```diff
@@ -0,0 +1,568 @@
...
+            gl.glLoadMatrixf(m)
+
+            world_tex.draw()
+
+            gl.glMatrixMode(gl.GL_PROJECTION)
+            gl.glPopMatrix()
+            gl.glMatrixMode(gl.GL_MODELVIEW)
+            gl.glPopMatrix()
+
+            # # Draw recent pupil positions onto the surface:
```
ckbaumann: Imo, this method is quite a bit too long. I think keeping all the ugly gl stuff in this single method is ok, but it would be good to extract `_draw_recent_pupil_positions_onto_surface` or so

pupil_src/shared_modules/surface_tracker/gui.py
```diff
@@ -0,0 +1,568 @@
...
+            if surface in self._edit_surf_markers:
+                self._draw_marker_toggles(surface)
+
+            if surface in self._edit_surf_corners:
+                self._draw_surface_corner_handles(surface)
+
+            # The offline surface tracker does not have the freeze feature and thus
+            # never uses the frozen_scene_tex
+            try:
+                if self.tracker.freeze_scene:
```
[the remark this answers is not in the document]
marc-tonsen: Option 1) would introduce an import cycle. Option 2) seems like a less implicit version of 3), so I choose 3) to resolve this!

pupil_src/shared_modules/surface_tracker/gui.py
```diff
@@ -0,0 +1,568 @@
...
+                    self.surface_windows[surface].update(self.tracker.frozen_scene_tex)
+                else:
+                    self.surface_windows[surface].update(self.tracker.g_pool.image_tex)
+            except AttributeError:
+                self.surface_windows[surface].update(self.tracker.g_pool.image_tex)
+
+    def _draw_markers(self):
+        color = pyglui_utils.RGBA(*self.color_secondary, .5)
+        for m in self.tracker.markers:
```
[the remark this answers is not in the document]
marc-tonsen: Absolutely agreed!

pupil_src/shared_modules/surface_tracker/gui.py
```diff
@@ -0,0 +1,568 @@
...
+        self.glfont.draw_text(loc[0], loc[1], text)
+
+    def _draw_marker_toggles(self, surface):
+        active_markers = []
+        inactive_markers = []
+        for marker in self.tracker.markers:
+            if marker.perimeter < self.tracker.marker_min_perimeter:
+                continue
+
+            centroid = np.mean(marker.verts_px, axis=0)
```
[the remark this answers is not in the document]
marc-tonsen: Agreed, but I would like to exclude this into a separate issue that changes the square marker detection to output objects like this already. There all functionality a marker class should have can be added. I will add a TODO on this in the code though.

pupil_src/shared_modules/surface_tracker/gui.py
```diff
@@ -0,0 +1,568 @@
...
+        m = norm_trans @ surface.surf_to_dist_img_trans
+        m = gl_utils.cvmat_to_glmat(m)
+
+        gl.glMatrixMode(gl.GL_PROJECTION)
+        gl.glPushMatrix()
+        gl.glLoadIdentity()
+        gl.glOrtho(0, 1, 0, 1, -1, 1)
+        gl.glMatrixMode(gl.GL_MODELVIEW)
+        gl.glPushMatrix()
+        # apply m  to our quad - this will stretch the quad such that the ref suface will span the window extends
```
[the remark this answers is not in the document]
marc-tonsen: I love PyCharm for suggesting to do this automatically :)

pupil_src/shared_modules/surface_tracker/gui.py
```diff
@@ -0,0 +1,568 @@
...
+        self.heatmap_textures[surface].draw()
+        gl.glMatrixMode(gl.GL_PROJECTION)
+        gl.glPopMatrix()
+        gl.glMatrixMode(gl.GL_MODELVIEW)
+        gl.glPopMatrix()
+
+    def on_click(self, pos, button, action):
+        pos = np.array(pos, dtype=np.float32)
+
+        # Menu Buttons
```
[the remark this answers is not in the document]
marc-tonsen: The names you suggest kinda loose the connection to `on_click` IMO. What about `_on_click_menu_buttons`, `_on_click_corner_handles` and `_on_click_marker_toggles`?

pupil_src/shared_modules/surface_tracker/gui.py
```diff
@@ -0,0 +1,568 @@
...
+    def _check_surface_button_pressed(self, surface, pos):
+        frame, hat, text_anchor, surface_edit_anchor, marker_edit_anchor = self._get_surface_anchor_points(
+            surface
+        )
+
+        surface_edit_anchor = np.array(surface_edit_anchor)
+        marker_edit_anchor = np.array(marker_edit_anchor)
+
+        dist_surface_edit = np.linalg.norm(pos - surface_edit_anchor)
+        surface_edit_pressed = dist_surface_edit < 15
```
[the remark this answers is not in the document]
marc-tonsen: Yeah, I copied this from the original version did not go through the hassle of replacing this with a variable yet :s

pupil_src/shared_modules/surface_tracker/surface.py
```diff
@@ -0,0 +1,513 @@
...
+        """Update surface definition by moving one of the corners to a new position.
+
+        Args:
+            corner_idx: Identifier of the corner to be moved. The order of corners is
+            `[(0, 0), (1, 0), (1, 1), (0, 1)]`
+            new_pos: The updated position of the corner in image pixel coordinates.
+            camera_model: Camera Model object.
+
+        """
+        # Markers undistorted
```
ckbaumann: The parts for undistorted and distorted markers look quite identical, can you maybe extract an method with some appropriate parameters?

pupil_src/shared_modules/surface_tracker/surface.py
```diff
@@ -0,0 +1,513 @@
...
+        new_corners[corner_idx] = new_corner_pos
+
+        transform = cv2.getPerspectiveTransform(new_corners, old_corners)
+        for m in self.reg_markers_dist.values():
+            m.verts_uv = cv2.perspectiveTransform(
+                m.verts_uv.reshape((-1, 1, 2)), transform
+            ).reshape((-1, 2))
+
+    def add_marker(self, marker_id, verts_px, camera_model):
+        """Add a marker to the surface definition."""
```
ckbaumann: Pedantic: This comment doesn't add any info and can be removed

pupil_src/shared_modules/surface_tracker/surface.py
```diff
@@ -0,0 +1,513 @@
...
+        surface_marker_dist = _Surface_Marker_Aggregate(marker_id)
+        marker_verts_dist = np.array(verts_px).reshape((4, 2))
+        uv_coords_dist = self.map_to_surf(
+            marker_verts_dist, camera_model, compensate_distortion=False
```
ckbaumann: Are you sure about `compensate_distortion=False`? Looks like code duplication similar to above

pupil_src/shared_modules/surface_tracker/surface.py
```diff
@@ -0,0 +1,513 @@
...
+            for g in gaze_on_surf
+            if g["on_surf"] and g["confidence"] >= self._heatmap_min_data_confidence
+        ]
+        aspect_ratio = self.real_world_size["y"] / self.real_world_size["x"]
+        grid = (
+            int(self._heatmap_resolution),
+            int(self._heatmap_resolution * aspect_ratio),
+        )
+        if heatmap_data:
+            xvals, yvals = zip(*((x, 1. - y) for x, y in heatmap_data))
```
ckbaumann: Please apply the new black to turn `1.` into `1.0` etc

pupil_src/shared_modules/surface_tracker/surface.py
```diff
@@ -0,0 +1,513 @@
...
+
+            hist = cv2.GaussianBlur(hist, (filter_h, filter_w), 0)
+            hist_max = hist.max()
+            hist *= (255. / hist_max) if hist_max else 0.
+            hist = hist.astype(np.uint8)
+        else:
+            self.within_surface_heatmap = self._get_dummy_heatmap(empty=True)
+            return
+
+        c_map = cv2.applyColorMap(hist, cv2.COLORMAP_JET)
```
ckbaumann: I would prefer `color_map`, this is clearer and still short

pupil_src/shared_modules/surface_tracker/surface.py
```diff
@@ -0,0 +1,513 @@
...
+        # reuse allocated memory if possible
+        if self.within_surface_heatmap.shape != (*grid, 4):
+            self.within_surface_heatmap = np.ones((*grid, 4), dtype=np.uint8)
+            self.within_surface_heatmap[:, :, 3] = 125
+        self.within_surface_heatmap[:, :, :3] = c_map
+
+    def _get_dummy_heatmap(self, empty=False):
```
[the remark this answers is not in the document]
ckbaumann: I don't understand what `empty` does here. From my understanding, the returned map should always be empty in the sense that it contains no information and is just uniform.... Could you think of some more descriptive name maybe?

pupil_src/shared_modules/surface_tracker/surface_online.py
```diff
@@ -0,0 +1,51 @@
...
+from surface_tracker.surface import Surface
+
+
+class Surface_Online(Surface):
+    """Surface_Online recalculates it's location on demand."""
+
+    def __init__(self, name="unknown", init_dict=None):
+        super().__init__(name=name, init_dict=init_dict)
+
+        self.gaze_history_length = 1
```
ckbaumann: If you don't intend to change this, it would be better to make it a class variable `GAZE_HISTORY_LENGTH` (aka constant).

pupil_src/shared_modules/surface_tracker/surface_online.py
```diff
@@ -0,0 +1,51 @@
...
+            self._update_definition(frame_idx, vis_markers_dict, camera_model)
+
+        # Get dict of current transformations
+        transformations = self.locate(
+            vis_markers_dict,
+            camera_model,
+            self.reg_markers_undist,
+            self.reg_markers_dist,
+        )
+        self.__dict__.update(transformations)
```
ckbaumann: I think this is kind of dangerous because you won't get an error if the structure of the returned dictionary changes and will only get weird errors at other places.

pupil_src/shared_modules/surface_tracker/surface.py
```diff
@@ -0,0 +1,513 @@
...
+        # used for visualization.
+        self.reg_markers_undist = {}
+        self.reg_markers_dist = {}
+        self.detected = False
+        self.img_to_surf_trans = None
+        self.surf_to_img_trans = None
+        self.dist_img_to_surf_trans = None
+        self.surf_to_dist_img_trans = None
+
+        self._required_obs_per_marker = 5
```
ckbaumann: I suspect that a few of those variable are not supposed to be changed (i.e. are constants). You could think about making them class variable with all-uppercase names, but actually I like it that similar variables are grouped together here, so I think it's ok to leave them as instance variables.

pupil_src/shared_modules/surface_tracker/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+    "Square_Marker_Detection", ["id", "id_confidence", "verts_px", "perimeter"]
+)
+
+
+class Heatmap_Mode(Enum):
+    WITHIN_SURFACE = "Gaze within each surface"
+    ACROSS_SURFACES = "Gaze across different surfaces"
+
+
+class _Surface_Marker_Aggregate(object):
```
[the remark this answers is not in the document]
marc-tonsen: @papr do you have an opinion on this?

pupil_src/shared_modules/surface_tracker/surface_tracker.py
```diff
@@ -0,0 +1,458 @@
...
+        ]
+        surface_definitions.save()
+
+    def deinit_ui(self):
+        self.g_pool.quickbar.remove(self.add_button)
+        self.add_button = None
+        self.remove_menu()
+
+    def cleanup(self):
+        self.save_surface_definitions_to_file()
```
ckbaumann: After reading the whole class, I think one could split it up into several parts, e.g.:

- The actual plugin (mostly holds and calls references to the other classes)
- General menu UI
- UI for individual surfaces
- Logic (updating surfaces etc)
- Persistence (loading / storing surfaces)

And perhaps more.

This would be quite some additional work, but currently you have to thoughtfully read the file from top to bottom to understand what all its tasks are (even after this review, I don't have a very clear picture of this in my head)

pupil_src/shared_modules/surface_tracker/surface_tracker_online.py
```diff
@@ -0,0 +1,154 @@
...
+        self.freeze_scene = False
+        self.frozen_scene_frame = None
+        self.frozen_scene_tex = None
+        super().__init__(
+            g_pool,
+            marker_min_perimeter=marker_min_perimeter,
+            inverted_markers=inverted_markers,
+        )
+
+        self.ui_info_text = "This plugin detects and tracks fiducial markers visible in the scene. You can define surfaces using 1 or more marker visible within the world view by clicking *add surface*. You can edit defined surfaces by selecting *Surface edit mode*."
```
ckbaumann: Line too long. You can split like
```
self.ui_info_text = "This plugin detects and tracks fiducial markers "
    "visible in the scene. You can define surfaces using 1 or "
    "more marker visible within the world view by clicking *add surface*."
    ...
```

pupil_src/shared_modules/surface_tracker/surface_tracker_online.py
```diff
@@ -0,0 +1,154 @@
...
+            else:
+                self.frozen_scene_tex = None
+
+        self.menu.append(
+            pyglui.ui.Switch(
+                "freeze_scene", self, label="Freeze Scene", setter=set_freeze_scene
+            )
+        )
+
+    def _per_surface_ui_custom(self, surface, s_menu):
```
ckbaumann: I would prefer `surf_menu`

pupil_src/shared_modules/surface_tracker/surface_tracker_online.py
```diff
@@ -0,0 +1,154 @@
...
+                surface,
+                label="Gaze History Length [seconds]",
+                setter=set_gaze_hist_len,
+            )
+        )
+
+    def recent_events(self, events):
+        if self.freeze_scene:
+            current_frame = events.get("frame")
+            events["frame"] = self.current_frame
```
ckbaumann: You need to set it at the beginning AND at the end of `recent_events`? If yes, please explain in a short comment.

pupil_src/shared_modules/surface_tracker/surface_tracker_online.py
```diff
@@ -0,0 +1,154 @@
...
+        if self.gui.show_heatmap:
+            self._update_surface_heatmaps()
+
+        if self.freeze_scene:
+            events["frame"] = current_frame
+
+    def _update_markers(self, frame):
+        self._detect_markers(frame)
+
+    def _update_surface_locations(self, idx):
```
ckbaumann: `surf_idx` would be better

pupil_src/shared_modules/surface_tracker/surface_tracker_online.py
```diff
@@ -0,0 +1,154 @@
...
+
+        gl_utils.make_coord_system_norm_based()
+
+        self.frozen_scene_tex.draw()
+
+        gl_utils.make_coord_system_pixel_based(
+            (self.g_pool.capture.frame_size[1], self.g_pool.capture.frame_size[0], 3)
+        )
+
+    def deinit_ui(self):
```
ckbaumann: If you just call `super()` you don't need to override, or do you want to emphasize that you don't change the deinit?

pupil_src/shared_modules/surface_tracker/surface.py
```diff
@@ -0,0 +1,513 @@
...
+    @abstractmethod
+    def update_location(self, frame_idx, visible_markers, camera_model):
+        """Update surface location based on marker detections in the current frame."""
+        pass
+
+    @staticmethod
+    def locate(visible_markers, camera_model, reg_markers_undist, reg_markers_dist):
+        """Computes homographys mapping the surface from and to image pixel space."""
+
+        result = {
```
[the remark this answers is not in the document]
marc-tonsen: I like this proposal a lot!

pupil_src/shared_modules/surface_tracker/surface.py
```diff
@@ -0,0 +1,513 @@
...
+            "detected": False,
+            "dist_img_to_surf_trans": None,
+            "surf_to_dist_img_trans": None,
+            "img_to_surf_trans": None,
+            "surf_to_img_trans": None,
+            "num_detected_markers": 0,
+        }
+
+        vis_reg_marker_ids = set(visible_markers.keys()) & set(
```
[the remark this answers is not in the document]
marc-tonsen: I do not like the idea of a list of acronyms, instead I wrote out the acronyms.

pupil_src/shared_modules/surface_tracker/surface.py
```diff
@@ -0,0 +1,513 @@
...
+            reg_markers_undist.keys()
+        )
+
+        if not vis_reg_marker_ids or len(vis_reg_marker_ids) < min(
```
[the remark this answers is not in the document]
marc-tonsen: The logic here is not super intuitive:
_If the surface is defined by 2+ markers, we require 2+ markers to be detected.
If the surface is defined by 1 marker, we require 1 marker to be detected._

If only one marker is detected the localisation is very noisy. Therefore we usually skip in this case. The exception is the rare and not recommended case in which users define a surface with a single marker. There are apparently use cases where this is necessary and therefore we support it.

I tried to clarify this with a comment.

pupil_src/shared_modules/surface_tracker/surface.py
```diff
@@ -0,0 +1,513 @@
...
+        self._avg_obs_per_marker = num_observations / len(self.reg_markers_undist)
+        self.build_up_status = self._avg_obs_per_marker / self._required_obs_per_marker
+
+        if self.build_up_status >= 1:
+            self._finalize_def()
+
+    def _bounding_quadrangle(self, verts):
+        hull = cv2.convexHull(verts, clockwise=False)
+
+        if hull.shape[0] > 4:
```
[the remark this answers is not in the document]
marc-tonsen: Not really. I added a comment and changed the notation a bit to clearly convey the logic here.

pupil_src/shared_modules/surface_tracker/surface.py
```diff
@@ -0,0 +1,513 @@
...
+        # verts space has its origin at the image center.
+        # adding 1 to the coordinates puts the origin at the top left.
+        distance_to_top_left = np.sqrt(
+            (hull[:, :, 0] + 1) ** 2 + (hull[:, :, 1] + 1) ** 2
+        )
+        bot_left_idx = np.argmin(distance_to_top_left) + 1
+        hull = np.roll(hull, -bot_left_idx, axis=0)
+        return hull
+
+    def _get_trans_to_norm(self, verts):
```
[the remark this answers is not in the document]
marc-tonsen: `cv2.getPerspectiveTransform` returns a transformation, so yes this function does return a transformation. The returned transformation transforms a set of corners (in pixel space) onto the normalized surface corners. Would `_get_trans_to_norm_corners` be more clear?

pupil_src/shared_modules/surface_tracker/surface.py
```diff
@@ -0,0 +1,513 @@
...
+        norm_corners = np.array([(0, 0), (1, 0), (1, 1), (0, 1)], dtype=np.float32)
+        return cv2.getPerspectiveTransform(verts, norm_corners)
+
+    def _denormalize(self, points, camera_model):
```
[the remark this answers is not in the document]
marc-tonsen: This method is an artifact that is no longer used anywhere. Thanks for finding this!

pupil_src/shared_modules/surface_tracker/surface.py
marc-tonsen: That is a bug! Good find!

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -0,0 +1,785 @@
...
+    def add_surface(self, _=None, init_dict=None):
+        super().add_surface(init_dict=init_dict)
+        # Plugin initialization loads surface definitions before UI is initialized. Changing timeline height will fail in this case.
+        if self.markers or init_dict is not None:
+            try:
+                self.timeline.content_height += self.timeline_line_height
+                self._fill_gaze_on_surf_buffer()
+            except AttributeError:
+                pass
+            self.surfaces[-1].on_surface_changed = self.on_surface_change
```
ckbaumann: Maybe rename to make consistent (`changed` vs `change`)? In C++ people usually used the passive version for event handlers, but I think `change` is shorter and has the same information content. Also, the "best of best practices" for Python that forms the basis for our current style guide prefers "actions to descriptions"

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -0,0 +1,785 @@
...
+
+    def remove_surface(self, _):
+        super().remove_surface(_)
```
ckbaumann: Imo, the convention is that if one uses `_` as an argument, one is not interested its value. However, in this case you are interested because you need to pass it to the super call. I would suggest to rename it to something meaningful. Any opinion on this @papr ?

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -0,0 +1,785 @@
...
+            for surface in self.surfaces:
+                if surface.name == notification["name"]:
+                    surface.location_cache = None
+                    surface.within_surface_heatmap = surface.get_placeholder_heatmap()
+                    self._heatmap_update_requests.add(surface)
+                    break
+            self._fill_gaze_on_surf_buffer()
+
+        elif notification["subject"] == "should_export":
+            self.make_export = True
```
ckbaumann: Let's discuss how the export is handled in person, maybe we find a solution that does not rely on these trigger variables.

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -0,0 +1,785 @@
...
+            self.export_params = (notification["range"], notification["export_dir"])
+            self._fill_gaze_on_surf_buffer()
+
+        elif notification["subject"] == "gaze_positions_changed":
+            for surface in self.surfaces:
+                self._heatmap_update_requests.add(surface)
+                surface.within_surface_heatmap = surface.get_placeholder_heatmap()
+            self._fill_gaze_on_surf_buffer()
+
+    def on_surface_change(self, surface):
```
ckbaumann: Nice!

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -0,0 +1,785 @@
...
+        self.save_surface_definitions_to_file()
+        self._heatmap_update_requests.add(surface)
+        self._fill_gaze_on_surf_buffer()
+
+    def save_surface_statsics_to_file(self):
```
ckbaumann: Typo: `statistics`

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -0,0 +1,785 @@
...
+        """
+        between in and out mark
+
+            report: gaze distribution:
+                    - total gazepoints
+                    - gaze points on surface x
+                    - gaze points not on any surface
+
+            report: surface visisbility
```
ckbaumann: Typo: `visibility`

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
ckbaumann: This method could use some extractions

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -0,0 +1,785 @@
...
+
+            if surface.within_surface_heatmap is not None:
+                logger.info("Saved Heatmap as .png file.")
+                cv2.imwrite(
+                    os.path.join(metrics_dir, "heatmap" + surface_name + ".png"),
+                    surface.within_surface_heatmap,
+                )
+
+        logger.info("Done exporting reference surface data.")
+        # TODO enable export of surface image?
```
ckbaumann: Please include the code or delete it. Even if you delete it, git will remember and you can later restore. The problem with out-commented code in production code is that no one will ever delete it because it might be valuable, but slowly it will diverge from the surrounding code.

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -0,0 +1,785 @@
...
+
+    def deinit_ui(self):
+        super().deinit_ui()
+        self.g_pool.user_timelines.remove(self.timeline)
+        self.timeline = None
+        self.glfont = None
+
+    def cleanup(self):
+        super().cleanup()
+
```
ckbaumann: Probably unnecessary blank line

pupil_src/shared_modules/surface_tracker/surface_offline.py
```diff
@@ -0,0 +1,270 @@
...
+
+                # Stop searching if we looped once through the entire recording
+                if def_idx == frame_idx - 1:
+                    self.build_up_status = 1.0
+                    self._finalize_def()
+                    break
+
+                def_idx += 1
+
+            if self.defined:
```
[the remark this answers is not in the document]
marc-tonsen: Very good point!

pupil_src/shared_modules/surface_tracker/surface_tracker.py
```diff
@@ -0,0 +1,458 @@
...
+
+        def set_robust_detection(val):
+            self.robust_detection = val
+            self.notify_all(
+                {"subject": "surface_tracker.marker_detection_params_changed"}
+            )
+
+        try:
+            self.menu.elements[:] = []
+        except AttributeError:
```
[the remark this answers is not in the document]
marc-tonsen: Yes, this happens when surfaces are loaded from file. Each new surface triggers a call to `add_surface`, which calls `_update_ui`. Loading surfaces from file happens in the constructor of the class, which is executed before the UI is initialized. I added a comment in the code to clarify this.

pupil_src/shared_modules/surface_tracker/surface_tracker.py
```diff
@@ -0,0 +1,458 @@
...
+        )
+        self.menu.append(advanced_menu)
+        self.menu.append(pyglui.ui.Button("Add surface", self.add_surface))
+        for surface in self.surfaces:
+            self._per_surface_ui(surface)
+
+    def _update_ui_custom(self):
+        pass
+
+    def _per_surface_ui(self, surface):
```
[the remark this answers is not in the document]
marc-tonsen: How would you suggest to integrate the call to `_per_surface_ui_custom` into the `SurfaceUI` class?

pupil_src/shared_modules/surface_tracker/surface_tracker.py
```diff
@@ -0,0 +1,458 @@
...
+    def recent_events(self, events):
+        frame = events.get("frame")
+        self.current_frame = frame
+        if not frame:
+            return
+
+        self._update_markers(frame)
+        self._update_surface_locations(frame.index)
+        self._update_surface_corners()
+        self._add_surface_events(events, frame)
```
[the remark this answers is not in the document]
marc-tonsen: That is great input, thanks!

pupil_src/shared_modules/surface_tracker/surface_tracker.py
```diff
@@ -0,0 +1,458 @@
...
+                events["surfaces"].append(surface_event)
+
+    @abstractmethod
+    def _update_surface_heatmaps(self):
+        pass
+
+    def gl_display(self):
+        self.gui.update()
+
+    def add_surface(self, _=None, init_dict=None):
```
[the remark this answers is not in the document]
marc-tonsen: This function is used as callback function for the `add surface thumb`. The API for those callback functions requires them to accept an argument. This is my way of handling this. Do you think it would make more sense to use a named parameter which is never used?

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -0,0 +1,781 @@
...
+import os
+import logging
+import time
+import platform
+import multiprocessing
+import csv
+import itertools
+
+logger = logging.getLogger(__name__)
+if platform.system() == "Darwin":
```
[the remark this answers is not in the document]
marc-tonsen: @papr I am not sure why this has to be set. Can you comment on this?

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -0,0 +1,785 @@
...
+     - self.cache is a list containing marker positions for each frame.
+     - self.surfaces[i].cache is a list containing surface positions for each frame
+    Both caches are build up over time. The marker cache is also session persistent.
+    See marker_tracker.py for more info on this marker tracker.
+    """
+
+    def __init__(self, g_pool, marker_min_perimeter=60, inverted_markers=False):
+        self.timeline_line_height = 16
+        self.Surface_Class = Surface_Offline
+        super().__init__(g_pool, marker_min_perimeter, inverted_markers)
```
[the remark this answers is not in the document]
marc-tonsen: I think the best solution for this would be to make this a property, which is abstract in `Surface_Tracker`. I used capitals in the name because this is a reference to a class rather than an object. If new surfaces are added, they will be initialized using this class. I wanted to reflect this in the name by using the naming convention for classes.

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -0,0 +1,785 @@
...
+        self.ui_info_text = "The offline surface tracker will look for markers in the entire video. By default it uses surfaces defined in capture. You can change and add more surfaces here. \n \n Press the export button or type 'e' to start the export."
+        self.supported_heatmap_modes = [
+            Heatmap_Mode.WITHIN_SURFACE,
+            Heatmap_Mode.ACROSS_SURFACES,
+        ]
+
+        self.order = 0.2
```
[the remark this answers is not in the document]
marc-tonsen: `order` is a value defined by `Plugin`. `plugin.py` contains this explanation:

> between 0 and 1 this indicated where in the plugin excecution order you plugin lives

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -0,0 +1,785 @@
...
+                self.fixations_on_surf_buffer_filler = None
+                self._update_surface_heatmaps()
+                if self.make_export:
+                    self.save_surface_statsics_to_file()
+                self.gaze_on_surf_buffer = None
+                self.fixations_on_surf_buffer = None
+
+    def _update_markers(self, frame):
+        if self.cache_filler is not None:
+            self._update_marker_and_surface_caches()
```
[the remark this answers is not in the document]
marc-tonsen: At this location I do however not want to return if `cache_filler is None`. Here I want to update the cache with data from the background process if there is any. If not I proceed without the updates nonetheless.

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -0,0 +1,785 @@
...
+        # Move seek index to current frame if caches do not contain data for it
+        self.markers = self.marker_cache[frame.index]
+        self.markers_unfiltered = self.marker_cache_unfiltered[frame.index]
+        if self.markers is False:
```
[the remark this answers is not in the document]
marc-tonsen: The cache is setup to contain False if the according index has not been computed yet. That is a convention made in `Cache_List`.

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -0,0 +1,785 @@
...
+            self.markers = []
+            self.markers_unfiltered = []
+            self.cache_seek_idx.value = frame.index
+
+    def _update_marker_and_surface_caches(self):
+        for idx, markers in self.cache_filler.fetch():
```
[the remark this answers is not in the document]
marc-tonsen: No it is not, it is the index of the current world frame. At this point it is about marker detection, which is independent of the surfaces. An even better reason to improve the naming!

pupil_src/shared_modules/surface_tracker/surface_tracker_online.py
marc-tonsen: You mean `_per_surface_ui_custom --> surf_menu` ? This would be a bit ambiguous with `_per_surface_ui`.

ckbaumann: oh sorry, I meant instead of `s_menu`

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -0,0 +1,785 @@
...
+        self.glfont.set_size(self.timeline_line_height * 0.8 * scale)
+        self.glfont.draw_text(width, 0, "Marker Cache")
+        for idx, s in enumerate(self.surfaces):
+            gl.glTranslatef(0, self.timeline_line_height * scale, 0)
+            self.glfont.draw_text(width, 0, s.name)
+
+    def add_surface(self, _=None, init_dict=None):
+        super().add_surface(init_dict=init_dict)
+        # Plugin initialization loads surface definitions before UI is initialized. Changing timeline height will fail in this case.
+        if self.markers or init_dict is not None:
```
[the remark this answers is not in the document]
ckbaumann: It's hard to see for a reader why you do this. Maybe return `True` or `False` in `add_surface` depending on the success? Then you could do `successfully_added = super().add_surface(...)` which would make your code clearer. Otherwise, a comment might be appropriate.

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -0,0 +1,785 @@
...
+
+        self.gaze_on_surf_buffer_filler = background_tasks.background_gaze_on_surface(
+            self.surfaces,
+            section,
+            all_world_timestamps,
+            all_gaze_events,
+            self.camera_model,
+        )
+
+        if self.make_export:
```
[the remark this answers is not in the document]
ckbaumann: Something like
```
def _fill_gaze_on_surf_buffer(self):
    self._update_gaze_filler()
    if self.make_export()
        self._update_fixations_filler()
```

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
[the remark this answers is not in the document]
ckbaumann: Ahh, right. I think its a class attribute. I think its better to override it as a class attribute as well [like this](https://github.com/pupil-labs/pupil/blob/642ff39377c91b2cd285319f5339abea043fada7/pupil_src/shared_modules/annotations.py#L57)

pupil_src/shared_modules/surface_tracker/__init__.py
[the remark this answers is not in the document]
papr: I don't think that any module-internal code should be made accessable in `__init__.py`. Proposal: Internal code that multiple sub-modules depend on should be placed in a separate file.

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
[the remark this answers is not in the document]
papr: > https://docs.python.org/3.4/library/multiprocessing.html?highlight=process#contexts-and-start-methods

In `main.py` we set the default `start_method` to `spawn` for `Darwin`. IIRC this is to avoid issues with opengl/pylgui in the child processes. Generally, `fork` is more efficient than `spawn` and in the case of surface tracking it is not required to `spawn` child processes. I will ask @mkassner to verify this though.

pupil_src/shared_modules/surface_tracker/surface_tracker.py
papr: I suggest we postpone this refactor until we need to extend surface tracking.

papr:
```python
MainProcess - [INFO] os_utils: Disabled idle sleep.
player - [INFO] launchables.player: Session setting are from a  different version of this app. I will not use those.
player - [INFO] launchables.player: Starting new session with '/Volumes/Untitled/recordings/csbender/001'
player - [INFO] player_methods: Updating meta info
player - [INFO] player_methods: Checking for world-less recording
player - [INFO] camera_models: Previously recorded calibration found and loaded!
player - [INFO] player_methods: Updating meta info
player - [INFO] video_capture: Install pyrealsense to use the Intel RealSense backend
player - [INFO] calibration_routines.optimization_calibration.build: Building extension modules...
player - [INFO] pupil_detectors.build: Building extension modules...
player - [INFO] launchables.player: Application Version: 1.9.167
player - [INFO] launchables.player: System Info: User: papr, Platform: Darwin, Machine: Bilbo.fritz.box, Release: 18.2.0, Version: Darwin Kernel Version 18.2.0: Fri Oct  5 19:41:49 PDT 2018; root:xnu-4903.221.2~2/RELEASE_X86_64
Starting buffering thread!
player - [INFO] camera_models: Previously recorded calibration found and loaded!
player - [INFO] launchables.player: Session setting are a different version of this app. I will not use those.
player - [INFO] video_export.plugins.world_video_exporter: iMotions Exporter has been launched.
player - [ERROR] file_methods: `pupil` data file corrupted
player - [WARNING] surface_tracker.surface: You have loaded an old and deprecated surface definition! Please re-define this surface for increased mapping accuracy!
player - [ERROR] launchables.player: Process Player crashed with trace:
  File "/Users/papr/work/pupil/pupil_src/launchables/player.py", line 558, in player
    handle_notifications(n)
  File "/Users/papr/work/pupil/pupil_src/launchables/player.py", line 529, in handle_notifications
    g_pool.plugin_by_name[n["name"]], args=n.get("args", {})
  File "/Users/papr/work/pupil/pupil_src/shared_modules/plugin.py", line 340, in add
    plugin_instance = new_plugin(self.g_pool, **args)
  File "/Users/papr/work/pupil/pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py", line 64, in __init__
    self._init_marker_cache()
  File "/Users/papr/work/pupil/pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py", line 107, in _init_marker_cache
    self._recalculate_marker_cache()
  File "/Users/papr/work/pupil/pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py", line 149, in _recalculate_marker_cache
    self.cache_seek_idx,
  File "/Users/papr/work/pupil/pupil_src/shared_modules/surface_tracker/background_tasks.py", line 18, in background_video_processor
    (video_file_path, callable, seek_idx, visited_list),
  File "/Users/papr/work/pupil/pupil_src/shared_modules/background_helper.py", line 43, in __init__
    self.process.start()
  File "/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/process.py", line 112, in start
    self._popen = self._Popen(self)
  File "/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/context.py", line 223, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/context.py", line 284, in _Popen
    return Popen(process_obj)
  File "/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/popen_spawn_posix.py", line 32, in __init__
    super().__init__(process_obj)
  File "/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/popen_fork.py", line 20, in __init__
    self._launch(process_obj)
  File "/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/popen_spawn_posix.py", line 47, in _launch
    reduction.dump(process_obj, fp)
  File "/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
AttributeError: Can't pickle local object 'marker_detection_callable.<locals>.callable'
```

I just got this error while testing the PR on a user's recording. We encountered a similar problem with the Video Exporters on systems that `spawn` new proceses instead of `forking` them (Windows, MacOS).

The problem is that one cannot pickle inner functions. The solution is to implement `offline_utils.*_callable` as classes and implement `__call__`.

papr:
@romanroibu Could you please review the last two commits. Afterwards, I would merge and we can start creating incremental PRs.

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -34,6 +34,13 @@
 from surface_tracker.gui import Heatmap_Mode
 from surface_tracker.surface_offline import Surface_Offline
 
+# On macOS, "spawn" is set as default start method in main.py. This is not required
+# here and we set it back to "fork" to improve performance.
+if platform.system() == "Darwin":
+    mp = multiprocessing.get_context("fork")
```
romanroibu: nitpick: since `mp` is frequently used as a shorthand for `multiprocessing`, it could be better to explicitly name this `mp_context` or `mp_default_context`

papr: Agreed

papr: I actually found an other mp-inconsistency thanks to the renaming. 👍

Issue #1246, opened 2018-08-03
Player: cache offline natural feature calibration, gaze mapping, and fixations between sessions.


performlabrit:
Self explanatory, I think.  

The 3D pupil detection is cached.  Offline gaze-mapping should be similarly cached.  ...as well as the detected fixations.

papr:
Partially fixed in #1434 

Fixation cache support will be added in [v1.12](https://github.com/pupil-labs/pupil/milestone/19)

PR #1289, opened 2018-09-04
Offline data cache for gaze and fixation producers

### LINKED ISSUE #1254: Calculated fixations should be saved when using (Offline) Fixation Detector
Pupil Player: 1.8.26

When fixations are calculated via the Offline Fixation Detector plugin, the results should be saved in a file and it should be possible to reload it when opening a recording again. The calculation of the fixations might take far longer as the duration of the actual recording. At the moment, you always have to do this again when either making a change in the parameter of the plugin or when reopening a recording. There also should be a button to start the recalculation like in other plugins to make sure you are not accidentally overriding your calculated fixations.

papr:
Fixes #1254 and #1246 

Open bugs:
- [ ] Gaze sections are not restored properly if `Offline_Calibration` was closed during mapping

### DIFF
--- a/pupil_src/shared_modules/fixation_detector.py
+++ b/pupil_src/shared_modules/fixation_detector.py
@@ -1,4 +1,4 @@
-'''
+"""
 (*)~---------------------------------------------------------------------------
 Pupil - eye tracking platform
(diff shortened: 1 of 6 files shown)


papr:
Closed in favor of #1434

PR #1301, opened 2018-09-26
Eye Video Exporter

### LINKED ISSUE #1294: make eye video exporter
Add a plugin to  player that does eye video exports

ckbaumann:
Adds an Eye Video Export plugin that also supports rendering the detected pupil on the video.

Common functionality of this plugin and the iMotions plugin was extracted into a VideoExporter

fixes #1294

### DIFF
--- a/pupil_src/shared_modules/player_methods.py
+++ b/pupil_src/shared_modules/player_methods.py
@@ -60,18 +60,21 @@ class Bisector(object):
 
     def by_ts(self, ts):
         '''
-        :param ts: matching or closest timestamp to extract.
-        :return: datum that is matching or None if none is found
+        :param ts: timestamp to extract.
+        :return: datum that is matching
+        :raises: ValueError if no matching datum is found
         '''
-        found_i = np.searchsorted(self.data_ts, ts)
-        # searchsorted returns len(array) if the query is bigger than all elements
-        in_range = found_i < len(self.data_ts)
-        if not in_range:
-            return None
-        found = self.data_ts[found_i] == ts
+        found_index = np.searchsorted(self.data_ts, ts)
+        try:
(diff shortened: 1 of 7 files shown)


ckbaumann:
Please don't merge before PR https://github.com/pupil-labs/pupil/pull/1307 is not merged

pupil_src/shared_modules/player_methods.py
```diff
@@ -58,6 +58,21 @@ def __init__(self, data=(), data_ts=()):
...
             self.data = self.data[self.sorted_idc].tolist()
 
+    def by_ts(self, ts):
+        '''
+        :param ts: matching or closest timestamp to extract.
+        :return: datum that is matching or None if none is found
+        '''
+        found_i = np.searchsorted(self.data_ts, ts)
+        # searchsorted returns len(array) if the query is bigger than all elements
+        in_range = found_i < len(self.data_ts)
```
mkassner: I would not check like this. Rather catch the exception. better even leave it unhandles. This is something the caller should handle.

papr: Actually, I disagree that the caller should be handling this. Yes, this check should not be performed explicitly but rather through a exception -- but it should be handled within the method. Either the `ts` is present in `data_ts` and the datum is returned, or it is not and `None` is returned. There is no reason why this function should throw an exception if `ts` is bigger than the last element in `data_ts`

ckbaumann: Thank you, I was not aware of this. It now throws a ValueError

papr: For completion: A `ValueError` is raised instead of returning `None`. This function therefore returns a valid result or raises a `ValueError`.

PR #1303, opened 2018-09-26
Video Export Launcher: Fix filenames 

### LINKED ISSUE #1295: rename exported world video
drop the `_viz` suffix form file.

emredog:
This fixes #1295 

The file path and the file name are displayed once the export is done, so I removed most of the text from the right pane (but left them commented for the moment).

### DIFF
--- a/pupil_src/shared_modules/video_export_launcher.py
+++ b/pupil_src/shared_modules/video_export_launcher.py
@@ -66,9 +66,11 @@ class Video_Export_Launcher(Analysis_Plugin_Base):
 
     def _update_ui(self):
         del self.menu.elements[:]
-        # self.menu.append(ui.Info_Text('The export can be found in the exports folder, under the recording directory.'))
-        # self.menu.append(ui.Info_Text('The world video file will be named as {}'.format(self.rec_name)))
-        # self.menu.append(ui.Text_Input('rec_name',self,label='Export name', setter=lambda x: None))
+        self.menu.append(
+            ui.Info_Text(
+                "The export can be found in the exports folder, under the recording directory."
+            )
+        )
         self.menu.append(
             ui.Info_Text(
                 "Select your export frame range using the trim marks in the seek bar. This will affect all exporting plugins."
--- a/pupil_src/shared_modules/video_export_launcher.py
+++ b/pupil_src/shared_modules/video_export_launcher.py
@@ -26,30 +26,21 @@ logger = logging.getLogger(__name__)
 def verify_out_file_path(out_file_path, rec_dir):
     # Out file path verification
     if not out_file_path:
-        out_file_path = os.path.join(rec_dir, "world_viz.mp4")
+        out_file_path = os.path.join(rec_dir, "world.mp4")
     else:
         file_name = os.path.basename(out_file_path)
         dir_name = os.path.dirname(out_file_path)
         if not dir_name:
             dir_name = rec_dir
         if not file_name:
-            file_name = "world_viz.mp4"
+            file_name = "world.mp4"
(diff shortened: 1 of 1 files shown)


pupil_src/shared_modules/video_export_launcher.py
```diff
@@ -77,12 +66,9 @@ def init_ui(self):
 
     def _update_ui(self):
         del self.menu.elements[:]
-        self.menu.append(
-            ui.Info_Text(
-                "Supply export video recording name. The export will be in the recording dir. If you give a path the export will end up there instead."
-            )
-        )
-        self.menu.append(ui.Text_Input("rec_name", self, label="Export name"))
+        # self.menu.append(ui.Info_Text('The export can be found in the exports folder, under the recording directory.'))
```
ckbaumann: I would suggest to delete the out-commented code because we can restore these lines from git history if we need them. Anyway, I'm not sure if I as a user would appreciate the info in the panel, the log output tends to vanish very quickly, any thoughts on that @papr ?

emredog: Thanks for the comments @ckbaumann. My motivation behind removing most of the text, is that when you export, the full path to exported file is revealed. 
screenshot from 2018-09-26 16-13-46

So it's redundant information, but only after the export is executed.

And I agree with you on removing the commented code.

pupil_src/shared_modules/video_export_launcher.py
```diff
@@ -45,57 +44,59 @@ def verify_out_file_path(out_file_path, rec_dir):
...
-        self.menu.append(ui.Info_Text('Supply export video recording name. The export will be in the recording dir. If you give a path the export will end up there instead.'))
-        self.menu.append(ui.Text_Input('rec_name',self,label='Export name'))
-        self.menu.append(ui.Info_Text('Select your export frame range using the trim marks in the seek bar. This will affect all exporting plugins.'))
-        self.menu.append(ui.Info_Text("Press the export button or type 'e' to start the export."))
+        # self.menu.append(ui.Info_Text('The export can be found in the exports folder, under the recording directory.'))
+        # self.menu.append(ui.Info_Text('The world video file will be named as {}'.format(self.rec_name)))
+        # self.menu.append(ui.Text_Input('rec_name',self,label='Export name', setter=lambda x: None))
+        self.menu.append(
+            ui.Info_Text(
+                "Select your export frame range using the trim marks in the seek bar. This will affect all exporting plugins."
```
papr: This text is deprecated as well. Export range is selected in the general settings since a few versions.

pupil_src/shared_modules/video_export_launcher.py
```diff
@@ -26,17 +26,16 @@
 def verify_out_file_path(out_file_path, rec_dir):
```
papr: I think we can get rid of this function completely.

pupil_src/shared_modules/video_export_launcher.py
papr: @ckbaumann fyi you can open the Log History plugin for more persistent log messages.

@ckbaumann @emredog 
I think this sentence is general enough and that we can keep it:
> The export can be found in the exports folder, under the recording directory.

pupil_src/shared_modules/video_export_launcher.py
```diff
@@ -145,13 +156,15 @@ def cleanup(self):
 
     def precomputed_for_range(self, export_range):
```
papr: @ckbaumann I think that this function might be relevant for your work on #1301

papr:
Well done!

Issue #1239, opened 2018-07-27
Set parameter for Raw Data Exporter in Pupil Player


ewolf29:
I just recognized that you always export all raw data to the export folder. Having long recordings create huge exports when always exporting all data. Wouldn't it be a good idea to select whether you want to export pupil_positions, gaze_position, surfaces, etc. to optimize the export time.

papr:
You are totally right. Did you know that you can trim the recording and only export data that belongs to this section?

ewolf29:
Yes, that's already a nice feature. :) Just see this issue as a feature request to improve your product.

papr:
Ok, I understand. Something that should help already: The surface data is exported by the offline surface tracker and not by the Raw Data Exporter. If you are not interested in exporting surface data, simply disable the surface tracker plugin. To further iterate on that: The raw data exporter only exports pupil and gaze positions. This is comparably fast. Nonetheless, adding ui elements to turn on/off the exports is easy. Thank you for your feedback. We will consider adding this feature to the next release.

ewolf29:
Ah, it already helps a lot to know that all the other "raw"-data is exported by the corresponding plugin. Thanks.

papr:
Closed via #1310

PR #1310, opened 2018-10-04
Set parameter for Raw Data Exporter in Pupil Player #1239

fneitzel:
I didn't know how to claim the task ;)

Integrated Switches.

### DIFF
--- a/pupil_src/shared_modules/raw_data_exporter.py
+++ b/pupil_src/shared_modules/raw_data_exporter.py
@@ -106,22 +106,22 @@ class Raw_Data_Exporter(Analysis_Plugin_Base):
     icon_chr = chr(0xe873)
     icon_font = 'pupil_icons'
     
-    def __init__(self, g_pool, export_pupil_positions=True, export_pupil_gaze_positions_info=True, export_gaze_positions=True):
+    def __init__(self, g_pool, should_export_pupil_positions=True, should_export_field_info=True, should_export_gaze_positions=True):
         super().__init__(g_pool)
-        self.export_pupil_positions = export_pupil_positions
-        self.export_pupil_gaze_positions_info = export_pupil_gaze_positions_info
-        self.export_gaze_positions = export_gaze_positions
+        self.should_export_pupil_positions = should_export_pupil_positions
+        self.should_export_field_info = should_export_pupil_field_info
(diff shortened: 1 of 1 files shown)


pupil_src/shared_modules/raw_data_exporter.py
```diff
@@ -105,12 +105,24 @@ class Raw_Data_Exporter(Analysis_Plugin_Base):
         '''
     icon_chr = chr(0xe873)
     icon_font = 'pupil_icons'
+    
+    def __init__(self, g_pool, export_pupil_positions=True, export_pupil_gaze_positions_info=True, export_gaze_positions=True):
+        super().__init__(g_pool)
+        self.export_pupil_positions = export_pupil_positions
+        self.export_pupil_gaze_positions_info = export_pupil_gaze_positions_info
+        self.export_gaze_positions = export_gaze_positions
```
papr: Would it be possible to rename these to `should_export_*`? Also, may I suggest to rename `export_pupil_gaze_positions_info` to `should_export_field_info`?

papr:
I caught a small undefined variable name bug. I applied the black code format and removed the `== True` while I was at it.

Thank you again @fneitzel, for contributing and investing your time into this project!

pupil_src/shared_modules/raw_data_exporter.py
```diff
@@ -102,162 +102,254 @@ class Raw_Data_Exporter(Analysis_Plugin_Base):
...
+
     icon_chr = chr(0xe873)
-    icon_font = 'pupil_icons'
+    icon_font = "pupil_icons"
+
+    def __init__(
+        self,
+        g_pool,
+        should_export_pupil_positions=True,
+        should_export_field_info=True,
```
mkassner: I know this is terribly late but I wonder. The initial motivation for adding this complexity was that exports a large/take a long time. This reason does not apply to writing the filed info. Why is this a setting then. Whats the cost of always doing this?

mkassner: @papr

mkassner: I know this is nit-picky but I just want to drive home the point that every slider and setting comes at a cost and need justification.

PR #1325, opened 2018-10-11
Handle time base changes in sliding window data collections

### LINKED ISSUE #1324: Timebase changes break fixation and blink detection in Capture
Plugins collecting data with a sliding window are not able to handle time base changes. Related to #1323

papr:
Fixes #1324 

- [x] blink detection
- [ ] fixation detector

### DIFF
--- a/pupil_src/shared_modules/blink_detection.py
+++ b/pupil_src/shared_modules/blink_detection.py
@@ -112,11 +112,16 @@ class Blink_Detection(Analysis_Plugin_Base):
         self.history.extend(events.get("pupil", []))
 
         try:  # use newest gaze point to determine age threshold
+            assert self.history[-1]["timestamp"] > self.history[0]["timestamp"]
             age_threshold = self.history[-1]["timestamp"] - self.history_length
             while self.history[1]["timestamp"] < age_threshold:
                 self.history.popleft()  # remove outdated gaze points
         except IndexError:
             pass
+        except AssertionError:
+            # negative time jump detected, reset history
+            del self.history[:]
+            return
 
         filter_size = len(self.history)
         if (


pupil_src/shared_modules/blink_detection.py
```diff
@@ -112,11 +112,16 @@ def recent_events(self, events={}):
         self.history.extend(events.get("pupil", []))
 
         try:  # use newest gaze point to determine age threshold
+            assert self.history[-1]["timestamp"] > self.history[0]["timestamp"]
```
ckbaumann: Do you need this assert once or should it rather be part of the while loop? Otherwise, I would go with something like
```
inconsistent_history = self.history[-1]["timestamp"] > self.history[0]["timestamp"]
if inconsistent_history:
    self.reset_history()
    return
```
I think asserts make most sense for "debugging" (i.e. you expect it to be false all the time), not for something that you expect can actually happen. The above code makes it imo clearer and is also faster (assert also always evaluates the condition). As an additional bonus, you no longer need the comment.

ckbaumann: Apparently it's also possible to just disable assert globally.

pupil_src/shared_modules/blink_detection.py
```diff
@@ -112,11 +112,16 @@ def recent_events(self, events={}):
         self.history.extend(events.get("pupil", []))
 
         try:  # use newest gaze point to determine age threshold
```
ckbaumann: This comment should be right above `age_threshold = `

pupil_src/shared_modules/blink_detection.py
papr: You are right that we could keep the newest one.

papr: Unfortunately, I forgot that history is a `deque` does not allow sliced deletions. Which leaves us with two options:
1. `history.clear()`, single fast call, clears complete history
2. `history.popLeft()` until only one element remains

I favor option 1. heavily since
a) such big timebase changes rarely happen,
b) the history is comparably short and fills quite quickly,
c) option 2. would mean much more code

ckbaumann: agreed

pupil_src/shared_modules/fixation_detector.py
```diff
@@ -681,7 +681,7 @@ def __init__(
         self, g_pool, max_dispersion=3.0, min_duration=300, confidence_threshold=0.75
     ):
         super().__init__(g_pool)
-        self.queue = deque()
+        self.history = deque()
```
ckbaumann: :+1:

pupil_src/shared_modules/fixation_detector.py
```diff
@@ -698,21 +698,34 @@ def recent_events(self, events):
...
+                return
+
+            age_threshold = ts_newest - self.min_duration / 1000.
+            # pop elements until only one element below the age threshold remains:
+            while self.history[1]["timestamp"] < age_threshold:
+                self.history.popleft()  # remove outdated gaze points
+
         except IndexError:
             pass
+        except AssertionError:
```
ckbaumann: This should be removed

PR #1332, opened 2018-10-16
match frame resolution in both dimensions

helei0147:
match frame resolution in both dimensions instead of only the first dimension.


papr:
@helei0147 Thank you very much! Your contribution is very much appreciated!

PR #1334, opened 2018-10-17
Fix Annotations

### LINKED ISSUE #1275: Fix receiving high frequency notifications
Sometimes the mp4 files are not readable (even with VLC) – I understand, that is due to corrupted headers (although I do not understand why it is happening so often) .... what I missed until today, is that the eye*.timestamps file for corrupted videos is missing!

we use windows, 1.8.26.... issue is apparent with both codecs (big and small file) ... world video seems to be always fine. we're using pupil capture, not mobile.

In rare cases only one eye is unreadable, but the other is fine.

ffmpeg says "moov atom not found" for the video files.

So far no clue inside the logs – both eye videos in this recording are damaged and the timestamps are missing. 

[capture.log](https://github.com/pupil-labs/pupil/files/2310431/capture.log)

ckbaumann:
fixes #1275

### DIFF
--- a/pupil_src/shared_modules/annotations.py
+++ b/pupil_src/shared_modules/annotations.py
@@ -10,32 +10,24 @@ See COPYING and COPYING.LESSER for license details.
 """
 
 import csv
-
-# logging
 import logging
 import os
-from itertools import chain
 
-import numpy as np
-from OpenGL.GL import *
 from pyglui import ui
-from pyglui.cygl.utils import RGBA, draw_polyline
-from pyglui.pyfontstash import fontstash
-from pyglui.ui import get_opensans_font_path
 
-import player_methods as pm
 import file_methods as fm
-from glfw import glfwGetCurrentContext, glfwGetWindowSize
+import player_methods as pm
 from plugin import Analysis_Plugin_Base, Plugin
 
 logger = logging.getLogger(__name__)
 
 
 class Annotation_Capture(Plugin):
-    """Describe your plugin here
+    """
+    Pupil Player plugin to view, edit, and add annotations.
     """
 
-    icon_chr = chr(0xE866)
+    icon_chr = chr(0xe866)
     icon_font = "pupil_icons"
 
     def __init__(self, g_pool, annotation_definitions=(("My annotation", "E"),)):
@@ -47,7 +39,7 @@ class Annotation_Capture(Plugin):
         self.annotation_definitions = list(annotation_definitions)
 
         self.new_annotation_name = "new annotation name"
-        self.new_annotation_hotkey = "e"
+        self.new_annotation_hotkey = "E"
 
         self.current_frame = -1
 
@@ -113,7 +105,7 @@ class Annotation_Capture(Plugin):
     def fire_annotation(self, annotation_label):
         t = self.g_pool.get_timestamp()
         logger.info('"{}"@{}'.format(annotation_label, t))
-        # you may add more field to this dictionary if you want.
+        # you may add more fields to this dictionary if you want.
         notification = {
             "subject": "annotation",
             "label": annotation_label,
@@ -132,8 +124,8 @@ class Annotation_Capture(Plugin):
 
 
 class Annotation_Player(Annotation_Capture, Analysis_Plugin_Base):
-    """Describe your plugin here
-    View,edit and add Annotations.
+    """
+    Pupil Player plugin to view, edit, and add annotations.
     """
 
     def __init__(self, g_pool, *args, **kwargs):
@@ -144,6 +136,7 @@ class Annotation_Player(Annotation_Capture, Analysis_Plugin_Base):
             self.load_cached_annotations()
         else:
             self.extract_annotations_from_recorded_notifications()
+        self.last_frame_ts = None
 
     def load_cached_annotations(self):
         annotations = fm.load_pldata_file(self.cache_dir, "annotations")
@@ -186,9 +179,11 @@ class Annotation_Player(Annotation_Capture, Analysis_Plugin_Base):
         self.update_buttons()
 
     def fire_annotation(self, annotation_label):
+        if self.last_frame_ts is None:
+            return
         t = self.last_frame_ts
         logger.info('"{}"@{}'.format(annotation_label, t))
-        # you may add more field to this dictionary if you want.
+        # you may add more fields to this dictionary if you want.
         annotation_new = {
             "subject": "annotation",
             "topic": "notify.annotation",
@@ -201,8 +196,8 @@ class Annotation_Player(Annotation_Capture, Analysis_Plugin_Base):
             annotation_new["timestamp"], fm.Serialized_Dict(python_dict=annotation_new)
         )
 
-    @classmethod
-    def parse_csv_keys(self, annotations):
+    @staticmethod
+    def parse_csv_keys(annotations):
         csv_keys = ("index", "timestamp", "label", "duration")
         system_keys = set(csv_keys)
         user_keys = set()
@@ -272,10 +267,6 @@ class Annotation_Player(Annotation_Capture, Analysis_Plugin_Base):
             self.export_annotations(notification["range"], notification["export_dir"])
 
     def cleanup(self):
-        """called when the plugin gets terminated.
-        This happens either voluntarily or forced.
-        if you have a GUI or glfw window destroy it here.
-        """
         with fm.PLData_Writer(self.cache_dir, "annotations") as writer:
             for ts, annotation in zip(self.annotations.timestamps, self.annotations):
                 writer.append_serialized(ts, "notify.annotation", annotation.serialized)
--- a/pupil_src/shared_modules/annotations.py
+++ b/pupil_src/shared_modules/annotations.py
@@ -12,213 +12,292 @@ See COPYING and COPYING.LESSER for license details.
 import csv
 import logging
 import os
+from abc import ABCMeta, abstractmethod
+from collections import namedtuple
 
 from pyglui import ui
 
 import file_methods as fm
 import player_methods as pm
+import zmq_tools
 from plugin import Analysis_Plugin_Base, Plugin
 
 logger = logging.getLogger(__name__)
 
 
-class Annotation_Capture(Plugin):
+def create_annotation(label, timestamp, duration=0.0, **custom_fields):
     """
-    Pupil Player plugin to view, edit, and add annotations.
+    Returns a dictionary in the format needed to send annotations
+    to an annotation plugin via the ICP.
+    
+    See python/remote_annotations.py in pupil-helpers for an example.
+
+    :param custom_fields:
+    """
+    return {
+        "topic": "annotation",
+        "label": label,
+        "timestamp": timestamp,
+        "duration": duration,
+        **custom_fields,
+    }
+
+
+AnnotationDefinition = namedtuple("AnnotationDefinition", "label hotkey")
+
+
+class AnnotationPlugin(Plugin, metaclass=ABCMeta):
+    """
+    Base for player and capture plugins that support adding and removing
+    annotations and the corresponding quickbar buttons
     """
 
-    icon_chr = chr(0xe866)
+    _AnnotationButtons = namedtuple("_AnnotationButtons", "quickbar menu")
+
+    icon_chr = chr(0xE866)
     icon_font = "pupil_icons"
 
-    def __init__(self, g_pool, annotation_definitions=(("My annotation", "E"),)):
+    def __init__(self, g_pool, annotation_definitions=None):
         super().__init__(g_pool)
         self.menu = None
-        self.sub_menu = None
-        self.buttons = []
+        self._annotation_list_menu = None
 
-        self.annotation_definitions = list(annotation_definitions)
+        if annotation_definitions is None:
+            annotation_definitions = [["My annotation", "E"]]
+        self._initial_annotation_definitions = annotation_definitions
+        self._definition_to_buttons = {}
 
-        self.new_annotation_name = "new annotation name"
-        self.new_annotation_hotkey = "E"
+        self._new_annotation_label = "new annotation label"
+        self._new_annotation_hotkey = "E"
 
-        self.current_frame = -1
+    def get_init_dict(self):
+        annotation_definitions = list(self._definition_to_buttons.keys())
+        return {"annotation_definitions": annotation_definitions}
 
     def init_ui(self):
         self.add_menu()
-        self.menu.label = "Annotations"
-        self.menu.append(ui.Text_Input("new_annotation_name", self))
-        self.menu.append(ui.Text_Input("new_annotation_hotkey", self))
-        self.menu.append(ui.Button("add annotation type", self.append_annotation))
-        self.sub_menu = ui.Growing_Menu("Events - click to remove")
-        self.menu.append(self.sub_menu)
-        self.update_buttons()
+        self.customize_menu()
+        self.menu.append(
+            ui.Text_Input("_new_annotation_label", self, label="New Label")
+        )
+        self.menu.append(
+            ui.Text_Input("_new_annotation_hotkey", self, label="New Hotkey")
+        )
+        self.menu.append(
+            ui.Button("Add Annotation Type", self._on_add_annotation_clicked)
+        )
+        self._annotation_list_menu = ui.Growing_Menu(
+            "Annotation List - Click to Remove"
+        )
+        self.menu.append(self._annotation_list_menu)
+        self._create_initial_annotation_list()
 
-    def update_buttons(self):
-        for b in self.buttons:
-            self.g_pool.quickbar.remove(b)
-            self.sub_menu.elements[:] = []
-        self.buttons = []
+    def _create_initial_annotation_list(self):
+        for label, hotkey in self._initial_annotation_definitions:
+            self._add_annotation_definition(label, hotkey)
 
-        for e_name, hotkey in self.annotation_definitions:
-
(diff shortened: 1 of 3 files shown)


pupil_src/shared_modules/player_methods.py
```diff
@@ -269,6 +271,12 @@ def update_meta_info(rec_dir, meta_info):
         csv_utils.write_key_value_file(csvfile,meta_info)
 
 
+def _update_info_version_to(new_version, rec_dir):
```
papr: Do you have a specific reason for the order of the arguments? Intuitively, I would have reversed the order... I cannot tell why.

ckbaumann: I know what you mean and I thought about this as well. I went for this order because something like _update_info_version_to('1.9', rec_dir) reads so nicely ("update info version to 1.9")

pupil_src/shared_modules/annotations.py
```diff
@@ -7,239 +7,267 @@
...
-    def remove_annotation(self, annotation):
-        try:
-            self.annotation_definitions.remove(annotation)
-        except ValueError:
-            print(annotation, self.annotation_definitions)
-        self.update_buttons()
 
+    def _update_quickbar_buttons(self):
+        self._clear_quickbar_buttons()
+        self._add_new_quickbar_buttons()
```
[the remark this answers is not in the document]
ckbaumann: I will change this as discussed in person.

pupil_src/shared_modules/annotations.py
```diff
@@ -7,239 +7,329 @@
...
+        pass
+
+    def _add_annotation_clicked(self):
+        # new_annotation_name and hotkey are set by the ui input fields
+        self._add_annotation_definition(
+            self._new_annotation_name, self._new_annotation_hotkey
+        )
+
+    def _add_annotation_definition(self, annotation_name, hotkey):
+        annotation_definition = AnnotationDefinition(
```
marc-tonsen: Is the existence of `AnnotationDefinition` necessary? It is almost identical to `_AnnotationInfo`, is there a reason why `_create_button_quickbar` could not receive a `_AnnotationInfo`-object too instead? `_AnnotationInfo` and `AnnotationDefinition` are kinda confusing/ambiguous.

pupil_src/shared_modules/annotations.py
```diff
@@ -7,239 +7,329 @@
...
+        return ui.Thumb(
+            annotation_definition.name,
+            setter=make_fire(annotation_definition),
+            getter=lambda: False,
+            label=annotation_definition.hotkey,
+            hotkey=annotation_definition.hotkey,
+        )
+
+    def _create_button_menu(self, annotation_info):
+        def make_remove(info):
```
marc-tonsen: wouldn't `make_remove = lambda: self._remove_annotation_info(annotation_info)` be simpler? And later 
`return ui.Button(label=name + " <" + hotkey + ">", function=make_remove)`

pupil_src/shared_modules/annotations.py
```diff
@@ -7,239 +7,329 @@
...
+            annotation_definition
+        )
+        annotation_info.button_menu = self._create_button_menu(annotation_info)
+        self._annotation_infos.append(annotation_info)
+
+        self._append_button_quickbar(annotation_info)
+        self._append_button_menu(annotation_info)
+
+    def _create_button_quickbar(self, annotation_definition):
+        def make_fire(definition):
```
marc-tonsen: Same as in line 134

pupil_src/shared_modules/annotations.py
```diff
@@ -7,239 +7,329 @@
...
+            return lambda _: self.fire_annotation(definition)
```
marc-tonsen:
```suggestion
            return lambda: self.fire_annotation(definition)
```

pupil_src/shared_modules/annotations.py
```diff
@@ -7,239 +7,329 @@
...
-        self.notify_all(notification)
+    def _clear_button_quickbars(self):
+        for annotation_info in self._annotation_infos:
+            self._remove_button_quickbar(annotation_info)
+
+    @abstractmethod
+    def fire_annotation(self, annotation_definition):
+        pass
+
+    def _add_annotation_clicked(self):
```
marc-tonsen: To me this naming suggests that you add a "clicked annotation", confused me a little at first. Would something like `_add_annotation_callback` or `_on_add_annotation` be better? Alternatively you can make changes to allow you to call  `_add_annotation_definition` directly.

pupil_src/shared_modules/annotations.py
```diff
@@ -7,239 +7,329 @@
...
 
-    def on_notify(self, notification):
-        if notification['subject'] == 'annotation':
-            logger.info('Received {} annotation'.format(notification['label']))
+    def _append_button_menu(self, annotation_info):
+        self._annotation_list_menu.append(annotation_info.button_menu)
 
-    def get_init_dict(self):
-        return {'annotation_definitions': self.annotation_definitions}
+    def _remove_annotation_info(self, annotation_info):
```
marc-tonsen: Should this be called `_remove_annotation_definition` instead to match `_add_annotation_definition`? Or change both to `Info`.

pupil_src/shared_modules/annotations.py
```diff
@@ -7,239 +7,329 @@
...
+        self._remove_button_quickbar(annotation_info)
+        self._remove_button_menu(annotation_info)
+        self._annotation_infos.remove(annotation_info)
+
+    def _remove_button_quickbar(self, annotation_info):
+        if annotation_info.button_quickbar:
```
marc-tonsen: Can this ever be `None` when this function is called?

pupil_src/shared_modules/annotations.py
```diff
@@ -7,239 +7,329 @@
...
-        with open(os.path.join(export_dir, 'annotations.csv'), 'w', encoding='utf-8', newline='') as csvfile:
-            csv_writer = csv.writer(csvfile)
-            csv_writer.writerow(csv_keys)
-            for annotation, idx in zip(annotation_section['data'], annotation_idc):
-                csv_row = [idx]
-                csv_row.extend((annotation.get(k, '') for k in csv_keys[1:]))
-                csv_writer.writerow(csv_row)
-            logger.info("Created 'annotations.csv' file.")
+        ts = self.last_frame_ts
+        logger.info('"{}"@{}'.format(annotation_label, ts))
```
marc-tonsen: Why is this different than the log message in the Capture version in line 219?

pupil_src/shared_modules/annotations.py
```diff
@@ -7,239 +7,329 @@
...
+    def _remove_button_menu(self, annotation_info):
+        if annotation_info.button_menu:
+            self._annotation_list_menu.remove(annotation_info.button_menu)
+            annotation_info.button_menu = None
 
-class Annotation_Player(Annotation_Capture, Analysis_Plugin_Base):
-    """Describe your plugin here
-    View,edit and add Annotations.
+
+def create_annotation(label, timestamp):
```
marc-tonsen: Is there a reason why this is not a static method of `AnnotationPlugin`?

pupil_src/shared_modules/annotations.py
```diff
@@ -7,239 +7,329 @@
...
-                return lambda: self.remove_annotation((e_name, hotkey))
-
-            thumb = ui.Thumb(e_name, setter=make_fire(e_name, hotkey),
-                             getter=lambda: False, label=hotkey, hotkey=hotkey)
-            self.buttons.append(thumb)
-            self.g_pool.quickbar.append(thumb)
-            self.sub_menu.append(ui.Button(e_name+" <"+hotkey+">", make_remove(e_name, hotkey)))
+        self.customize_menu()
+        self.menu.append(ui.Text_Input("_new_annotation_name", self, label="New Name"))
+        self.menu.append(ui.Text_Input("_new_annotation_hotkey", self, label="New Hotkey"))
```
marc-tonsen: Line over 88 characters long

```suggestion
        self.menu.append(
            ui.Text_Input("_new_annotation_hotkey", self, label="New Hotkey")
        )
```

pupil_src/shared_modules/annotations.py
```diff
@@ -7,239 +7,329 @@
...
+            ui.Info_Text(
+                "Annotations recorded with capture are displayed when this "
+                "plugin is loaded. New annotations can be added with the "
+                "interface below."
+            )
+        )
+        self.menu.append(
+            ui.Info_Text(
+                "If you want to revert annotations to the recorded state, "
+                "stop player, delete the annotation_player.pldata file in the recording "
```
marc-tonsen: Line over 88 characters long

pupil_src/shared_modules/annotations.py
papr: This would not be the same! This is a so called [closure](https://www.programiz.com/python-programming/closure).

pupil_src/shared_modules/annotations.py
papr: We should discuss this.

pupil_src/shared_modules/annotations.py
papr: I like the idea

pupil_src/shared_modules/annotations.py
papr: This is the annotation being fired, line 219 is the annotation being received. I think we should change this one to a debug message.

pupil_src/shared_modules/player_methods.py
papr: Agreed

pupil_src/shared_modules/annotations.py
```diff
@@ -7,239 +7,333 @@
...
+
+    # class _AnnotationUI:
+    #     # no namedtuple because we cannot set the menu button at creation
+    #     # (this button references the _AnnotationInfo for deletion)
+    #     def __init__(self, definition, button_quickbar=None, button_menu=None):
+    #         self.definition = definition
+    #         self.button_quickbar = button_quickbar
+    #         self.button_menu = button_menu
+
+    _AnnotationButtons = namedtuple("_AnnotationUI", "quickbar menu")
```
marc-tonsen: `"_AnnotationUI"` is a little non-descriptive. Maybe use `sidebar` and `quickbar` as names?

pupil_src/shared_modules/annotations.py
```diff
@@ -7,239 +7,333 @@
...
 
-    def get_init_dict(self):
-        return {'annotation_definitions': self.annotation_definitions}
+    def _remove_annotation_info(self, annotation_definition):
+        buttons = self._definition_to_buttons[annotation_definition]
+        self._remove_button_quickbar(buttons.quickbar)
+        self._remove_button_menu(buttons.menu)
+        del self._definition_to_buttons[annotation_definition]
 
+    def _remove_button_quickbar(self, button_quickbar):
```
marc-tonsen: does it make sense to wrap these one-liners in a separate function? The one-liner is almost identical to the function names.

pupil_src/shared_modules/annotations.py
marc-tonsen: Did you forget to push your changes here @ckbaumann ?

pupil_src/shared_modules/annotations.py
ckbaumann: ups, thanks for catching that, the name should be `_AnnotationButtons`

pupil_src/shared_modules/annotations.py
ckbaumann: No, just forgot to change this as well ;)

pupil_src/shared_modules/annotations.py
```diff
@@ -10,220 +10,297 @@
...
     """
 
-    icon_chr = chr(0xE866)
+    # class _AnnotationUI:
+    #     # no namedtuple because we cannot set the menu button at creation
+    #     # (this button references the _AnnotationInfo for deletion)
+    #     def __init__(self, definition, button_quickbar=None, button_menu=None):
+    #         self.definition = definition
+    #         self.button_quickbar = button_quickbar
+    #         self.button_menu = button_menu
```
papr: Can the comments above be deleted?

pupil_src/shared_modules/annotations.py
```diff
@@ -10,220 +10,297 @@
...
+
+    _AnnotationButtons = namedtuple("_AnnotationButtons", "quickbar menu")
+
+    icon_chr = chr(0xe866)
     icon_font = "pupil_icons"
 
-    def __init__(self, g_pool, annotation_definitions=(("My annotation", "E"),)):
+    def __init__(self, g_pool, annotation_definitions=None):
```
papr: What is the reasoning behind this change?

pupil_src/shared_modules/annotations.py
```diff
@@ -10,220 +10,297 @@
...
+            self._add_annotation_definition(label, hotkey)
+
+    @abstractmethod
+    def customize_menu(self):
+        pass
 
     def deinit_ui(self):
-        self.menu.remove(self.sub_menu)
-        self.sub_menu = None
+        self._clear_button_quickbars()
```
papr:
```suggestion
        self._clear_quickbar_buttons()
```

pupil_src/shared_modules/annotations.py
```diff
@@ -10,220 +10,297 @@
...
-            self.annotation_definitions.remove(annotation)
-        except ValueError:
-            print(annotation, self.annotation_definitions)
-        self.update_buttons()
+    def _add_annotation_definition(self, annotation_label, hotkey):
+        annotation_definition = AnnotationDefinition(
+            label=annotation_label, hotkey=hotkey
+        )
+        if annotation_definition in self._definition_to_buttons:
+            return
```
papr: Can this be triggered by the user? If yes, there should be a warning.

pupil_src/shared_modules/annotations.py
```diff
@@ -10,220 +10,297 @@
...
-            "record": True,
-        }
-        self.notify_all(notification)
+    def _create_button_menu(self, annotation_definition):
+        def make_remove():
+            self._remove_annotation_info(annotation_definition)
+
+        label = annotation_definition.label
+        hotkey = annotation_definition.hotkey
+        return ui.Button(label=label + " <" + hotkey + ">", function=make_remove)
```
papr:
```suggestion
        return ui.Button(label="{} <{}>".format(label, hotkey), function=make_remove)
```
We usually prefer string-formatting over manual formatting.

I am looking forward to not having to support Python 3.5 anymore. Then we can do this:
```python
f"{label} <{hotkey}>"
```

pupil_src/shared_modules/annotations.py
```diff
@@ -10,220 +10,297 @@
...
+
+    def _append_button_quickbar(self, button_quickbar):
+        current_buttons = self.g_pool.quickbar.elements
+        index_of_last_button = -1
+        for definition, buttons in self._definition_to_buttons.items():
+            if not buttons.quickbar:
+                continue
+            try:
```
papr: Why not remove the `not buttons.quickbar` check and handle eventual exceptions in the try/except?

pupil_src/shared_modules/annotations.py
ckbaumann: Thanks :)

pupil_src/shared_modules/annotations.py
ckbaumann: This is the suggested way to add list default arguments, right?

pupil_src/shared_modules/annotations.py
ckbaumann: uhh lovely Yoda speach, it should be `_clear_buttons_quickbar`, this order is strange but consistent with reverse notation, right?

pupil_src/shared_modules/annotations.py
ckbaumann: Right, thanks!

pupil_src/shared_modules/annotations.py
ckbaumann: Thanks, I forgot this when introducing the namedtuple `AnnotationButtons`

pupil_src/shared_modules/annotations.py
marc-tonsen: In my understanding the reverse notation only makes sense if some kind of topology is defined, i.e. if there were multiple `_clear_button_*` methods, this notation would be appropriate. Since there is only this function, there is no reason to prefer on order over the other. `_clear_quickbar_buttons` is however easier to read.

pupil_src/shared_modules/annotations.py
marc-tonsen: I am not sure if this improves readability, but multiple usage is a good reason!

pupil_src/shared_modules/annotations.py
papr: Important is the plural/singular here, order is only a secondary thought.

ckbaumann: I wanted to be consistent with the other methods (add_button_x and create_button_x). I'm not a huge fan of reverse annotation, though, so in this case my personal preference would be create/add/clear_x_button everywhere

ckbaumann:
As Marc didn't have additional comments and is now not available to review, I now finished the branch with a rebase.

:neckbeard: :arrow_right: :princess:

ckbaumann:
I get an error when starting Capture after the rebase. Let's test this before merging to ensure that it was not caused by the rebase.

ckbaumann:
I tested it again and found no issues. Ready to merge imo.

PR #1306, opened 2018-09-27
Fixed background logging issue by replacing zmq handler with a StreamHandler

marc-tonsen:
Fixes #1305

PR #1342, opened 2018-10-25
Background Logging: Set push_url via class attribute instead of passing as init arg

papr:
Follow-up PR for #1306


pupil_src/shared_modules/background_helper.py
```diff
@@ -64,6 +66,10 @@ def _wrapper(self, pipe, _should_terminate_flag, generator, *args, **kwargs):
             pipe.close()
             logger.debug("Exiting _wrapper")
 
+    @abc.abstract_method
```
marc-tonsen: For this to have an effect your also have to add `metaclass=abc.ABCMeta` to the class definition, i.e. 
`class Task_Proxy(object, metaclass=abc.ABCMeta):`

pupil_src/shared_modules/background_helper.py
```diff
@@ -112,27 +118,19 @@ def __del__(self):
...
-
-    def _enforce_IPC_logging(self, ipc_push_url):
+    push_url = None
+
+    def _change_logging_behavior(self):
         """
         ZMQ_handler sockets from the foreground thread are broken in the background.
         Solution: Remove all potential broken handlers and replace by new oneself.
 
         Caveat: If a broken handler is present is incosistent across environments.
```
marc-tonsen:
```suggestion
        Caveat: If a broken handler is present it is incosistent across environments.
```

pupil_src/shared_modules/background_helper.py
```diff
@@ -112,27 +118,19 @@ def __del__(self):
...
         """
+        assert self.push_url, "`push_url` was not set by foreground process"
```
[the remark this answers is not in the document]
marc-tonsen: I think this as a `TypeError` this would be weird enough to justify this assertion! I like it.

pupil_src/shared_modules/background_helper.py
papr: I usually prefer way (1) but I see the annoyance.

pupil_src/shared_modules/background_helper.py
```diff
@@ -112,28 +122,28 @@ def __del__(self):
...
-        Caveat: If a broken handler is present is incosistent across environments.
+        Caveat: If a broken handler is present it is incosistent across environments.
         """
+        assert self.push_url, "`push_url` was not set by foreground process"
         del logger.root.handlers[:]
         zmq_ctx = zmq.Context()
-        handler = zmq_tools.ZMQ_handler(zmq_ctx, ipc_push_url)
+        handler = zmq_tools.ZMQ_handler(zmq_ctx, self.push_url)
         logger.root.addHandler(handler)
+        logger.root.setLevel(logging.NOTSET)
```
papr: Default `logger.root.level` is set to `logging.WARNING`. This filters logs before sending them onto the IPC. There is already filtering happening on the log-receiving side.

pupil_src/shared_modules/background_helper.py
```diff
@@ -45,9 +46,10 @@ def __init__(self, name, generator, args=(), kwargs={}):
 
     def _wrapper(self, pipe, _should_terminate_flag, generator, *args, **kwargs):
         """Executed in background, pipes generator results to foreground"""
-        logger.debug("Entering _wrapper")
-
         try:
+            self._change_logging_behavior()
+            logger.debug("Entering _wrapper")
+
```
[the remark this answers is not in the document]
ckbaumann: Imo, this should be a comment.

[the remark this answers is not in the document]
papr: This allows explicit testing of the logging behaviour though, if the debug log level is enabled. 🤔

PR #1361, opened 2018-10-31
World: Add set_min_calibration_confidence notification

papr:
Requires `value` field, unit: confidence


pupil_src/launchables/world.py
```diff
@@ -392,6 +393,8 @@ def handle_notifications(n):
                     "mode": g_pool.detection_mapping_mode,
                 }
                 ipc_pub.notify(n)
+            elif subject == "set_min_calibration_confidence":
```
ckbaumann: Maybe a chance to rename `n` to `noti` or `notification`? Would add some readability and at least `noti` is not very long either.

PR #1365, opened 2018-11-01
Rs2 record rgb

emredog:
Depth recording disabled on purpose.
Tested with different resolutions and frame rates.


papr:
I will close this PR since its changes are also included in #1366

PR #1367, opened 2018-11-02
Refactor calibration routines

euryalus:
-Corrected few typos
-Deleted unnecessary imports
-Deleted unused functions fit_error_angle and calculate_residual_3D_Points from calibrate.py
-Moved function match_data from finish_calibration.py to calibrate.py


pupil_src/shared_modules/calibration_routines/calibrate.py
```diff
@@ -17,6 +17,9 @@
 
 logger = logging.getLogger(__name__)
 
+not_enough_data_error_msg = (
```
marc-tonsen: This variable is only used inside of `match_data`. Is there a reason for it to be defined globally rather than inside of `match_data`?

pupil_src/shared_modules/calibration_routines/calibrate.py
```diff
@@ -398,6 +392,48 @@ def fn(pt_0, pt_1):
...
+    else:
+        logger.error(not_enough_data_error_msg)
+        return {
+            "subject": "calibration.failed",
+            "reason": not_enough_data_error_msg,
+            "timestamp": g_pool.get_timestamp(),
+            "record": True,
+        }
+
+    # match eye data and check if biocular and or monocular
```
marc-tonsen:
```suggestion
    # match eye data and check if binocular and or monocular
```

marc-tonsen: Does this comment make sense and/or is it needed? In what sense is eye data being matched? Isn't it rather gaze and pupil data being matched? 
What is checked for being monocular or binocular? We are always computing both versions anyway.

pupil_src/shared_modules/calibration_routines/calibrate.py
euryalus: Only historical ones. I have moved it as you proposed.

pupil_src/shared_modules/calibration_routines/calibrate.py
euryalus: Yeah, I agree. I have removed the comment.

pupil_src/shared_modules/calibration_routines/finish_calibration.py
```diff
@@ -512,6 +473,40 @@ def select_calibration_method(g_pool, pupil_list, ref_list):
             )
 
 
+def dump_calibration_data(binocular, eye0, eye1, result):
```
papr: This function was probably used for testing purposes and should not be part of this PR. Even if it should stay, I would recommend to use `g_pool.user_dir` instead of a hard-coded path.

pupil_src/shared_modules/calibration_routines/finish_calibration.py
```diff
@@ -10,14 +10,18 @@
 """
 
 import os
-import numpy as np
 
-from . import calibrate
```
papr: Isn't `calibrate` used further down? e.g. https://github.com/pupil-labs/pupil/pull/1367/files#diff-6325cc3346b4eb436ce02ce0aa3a38aaR348

pupil_src/shared_modules/math_helper/transformations.py
```diff
@@ -1980,6 +1983,33 @@ def _import_module(name, package=None, warn=True, prefix="_py_", ignore="_"):
         return True
 
 
+def find_rigid_transform(A, B):
+    # we expect the shape to be of length 2
+    assert len(A.shape) == len(B.shape) == 2
```
marc-tonsen: Assertions should generally not be used to make sure function parameters are given the correct arguments. Wrong parameters should throw e.g. a `ValueError` or in this case numpy should anyway throw an error in line 1999 because arrays of different lengths can not be multiplied like that. 
Please check e.g. [this](https://github.com/emre/notes/blob/master/python/when-to-use-assert.md) resource on when to use assertions and evaluate if they make sense at this point.

pupil_src/shared_modules/math_helper/transformations.py
```diff
@@ -1980,6 +1983,33 @@ def _import_module(name, package=None, warn=True, prefix="_py_", ignore="_"):
...
+    assert A.shape[0] == B.shape[0]
+
+    centroid_A = np.mean(A, axis=0)
+    centroid_B = np.mean(B, axis=0)
+
+    # centre the points
```
marc-tonsen: Please iterate over comments and variable names in this function. Some comments are not necessary. Please also solve the todo that is mentioned in one comment.

pupil_src/shared_modules/math_helper/transformations.py
```diff
@@ -1980,6 +1983,33 @@ def _import_module(name, package=None, warn=True, prefix="_py_", ignore="_"):
...
+    A -= centroid_A
+    B -= centroid_B
+
+    # dot is matrix multiplication for array
+    H = A.T @ B
+    U, S, Vt = np.linalg.svd(H)
+    R = Vt.T @ U.T
+    # special reflection case
```
marc-tonsen: Why is this case special and when does it occur? Should we throw an exception in this case? Forwarding this information without context to the end-user by logging it on the info level seems confusing/inappropriate to me.

pupil_src/shared_modules/calibration_routines/finish_calibration.py
euryalus: Totally! I was not aware that when I push a commit, that it would go to "Needs review" automatically.  I thought, I would need to trigger that manually. And I pushed something which was not completely finished.

pupil_src/shared_modules/calibration_routines/finish_calibration.py
euryalus: In the meantime, I have renamed it to calibrate_2d and import the only function needed in finish_calibration explicitly.

pupil_src/shared_modules/calibration_routines/gaze_mappers.py
```diff
@@ -236,7 +236,10 @@ def deinit_ui(self):
 
     def _map_binocular(self, p0, p1):
         if self.multivariate:
-            gaze_point = self.map_fn(p0["norm_pos"], p1["norm_pos"])
+            try:
+                gaze_point = self.map_fn([*p0["norm_pos"]]+[*p1["norm_pos"]]) #todo: Change call signature
+            except:
```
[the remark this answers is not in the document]
euryalus: Yeah, this was due to this not being finished and I wanted to fall back on the old way of calling the function when necessary. Will be gone in the next commit.

pupil_src/shared_modules/math_helper/transformations.py
euryalus: True fact. I will finish the rest of the calibration code first and put it on the list of things to do. This was in the code since ages and I just moved it to a more appropriate place. But I agree, this function needs some work.

pupil_src/shared_modules/math_helper/transformations.py
euryalus: See above.

pupil_src/shared_modules/math_helper/transformations.py
euryalus: See above.

euryalus:
I think, I am pretty much done with the refactor now. In particular, calibrate_2d has changed dramatically. Would be great to have your feedback on that!

pupil_src/shared_modules/calibration_routines/calibrate_2d.py
```diff
@@ -0,0 +1,264 @@
...
+
+        if binocular:
+            polynomial_type = PolynomialBinocular
+        else:
+            polynomial_type = PolynomialMonocular
+
+        self.map_to_world_cam_x = polynomial_type(params_x, dof, degree)
+        self.map_to_world_cam_y = polynomial_type(params_y, dof, degree)
+
+    def __call__(self, pt):
```
marc-tonsen: I am not sure about this, so @papr correct me if I am wrong, but if all your classes only implement `__call__` and nothing else it might make more sense to use a factory instead. A factory is a function that takes parameters and returns a function, which is specified by these parameters. For example:
```
def add_factory(b):
    def add(a):
        return a+b
    return add
```

The difference between the function returned by the factory and the `__call__` function of your class is that the parameters that define what happens in `__call__` are technically still mutable, while the function returned by the factory is not. In your case you do not want the function to be mutable, so the factory pattern might be more appropriate.

pupil_src/shared_modules/calibration_routines/calibrate_2d.py
```diff
@@ -0,0 +1,264 @@
...
+def fit_mapping_polynomials(
+    calibration_data,
+    binocular,
+    degree=2,
+    ignored_terms=(),
+    regularization=0.0,
+    loss_scale=-1,
+):
+    """
+    Fit two polynomials, one for each of the last two columns of the provided calibration data.
```
marc-tonsen: Please change style to [Google dosctring style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). You can also set this style as default in PyCharm, so it auto-creates the correct stump.

marc-tonsen: Also make sure the maximum line length is kept.

pupil_src/shared_modules/calibration_routines/calibrate_2d.py
```diff
@@ -0,0 +1,264 @@
...
+
+    :param calibration_data: Array of floats. The last two columns correspond to screen coordinates of the
+    calibration targets. The remaining columns can contain any data that we want to fit on. In the binocular case,
+    the number of columns of the array must be divisble by two.
+    :param binocular: Boolean. Indicates whether the data is binocular or not.
+    :param degree: Int. The degree of the polynomials used.
+    :param ignored_terms: List of tuples. Specifies terms of the polynomials which should be ignored in the fit.
+    Example: (1,2) ignores term of the form x*y**2.
```
marc-tonsen: I do not understand why this tuple maps to that term.

pupil_src/shared_modules/calibration_routines/calibrate_2d.py
```diff
@@ -0,0 +1,264 @@
...
+    :param regularization: Float. Scales a weight decay term in the total calibration cost.
+    :param loss_scale: Float. Outliers are taken care of by a Cauchy loss function. Roughly speaking, squared
+    residuals which are larger than the given loss scale are not influencing the outcome of the fit.
+    :return: Tuple. First entry is a Boolean, indicating whether the fit was successful. Second entry is a tuple
+    containing the parameters and signature of the fitted polynomials (to be used to initialize MappingFunction2D).
+    """
+
+    calibration_data = np.asarray(calibration_data)
```
marc-tonsen: In your documentation you state that `calibration_data` is already an array. Is this conversion needed?

pupil_src/shared_modules/calibration_routines/calibrate_2d.py
```diff
@@ -0,0 +1,264 @@
...
+
+    if binocular:
+        dof_multiplier = 2
+    else:
+        dof_multiplier = 1
+
+    dof = (calibration_data.shape[1] - 2) // dof_multiplier
+    n_params = dof_multiplier * ((degree + 1) ** dof - len(ignored_terms))
```
marc-tonsen: The way you compute `dof` and `n_params` is a bit difficult to follow. Can you insert a few meaningful variable name or split into more obvious steps to make this more clear?

pupil_src/shared_modules/calibration_routines/calibrate_2d.py
```diff
@@ -0,0 +1,264 @@
...
+
+    :param calibration_data: Array of floats. The last two columns correspond to screen coordinates of the
+    calibration targets. The remaining columns can contain any data that we want to fit on. In the binocular case,
+    the number of columns of the array must be divisble by two.
+    :param binocular: Boolean. Indicates whether the data is binocular or not.
+    :param degree: Int. The degree of the polynomials used.
+    :param ignored_terms: List of tuples. Specifies terms of the polynomials which should be ignored in the fit.
+    Example: (1,2) ignores term of the form x*y**2.
+    :param regularization: Float. Scales a weight decay term in the total calibration cost.
+    :param loss_scale: Float. Outliers are taken care of by a Cauchy loss function. Roughly speaking, squared
```
marc-tonsen: Can `loss_scale` be renamed to be more specific? Something like `cauchy_scale` or `outlier_scale`? The loss is not scaled by this exact value id understand correctly.

pupil_src/shared_modules/calibration_routines/calibrate_2d.py
```diff
@@ -0,0 +1,264 @@
...
+    Assumes the array to have the same size in all dimensions.
+    """
+    dims = len(multi_idx)
+    linear_idx = 0
+    for k, idx in enumerate(multi_idx):
+        linear_idx += idx * (size_per_dim ** (dims - 1 - k))
+    return linear_idx
+
+
+def extend_params(params, degree, ignored_terms=(), fill_value=0, binocular=False):
```
marc-tonsen: I feel like you should rename this function to something that make the first sentence of the docstring unnecessary.

pupil_src/shared_modules/calibration_routines/calibrate_2d.py
```diff
@@ -0,0 +1,264 @@
...
+    """
+    Insert a given fill value into a parameter list at places specified by ignored terms.
+
+    This is needed to comply with the API of numpy.polynomial.polynomial when polynomials are considered, that do not
```
marc-tonsen: Line length.

pupil_src/shared_modules/calibration_routines/calibrate_2d.py
```diff
@@ -0,0 +1,264 @@
...
+    contain all terms.
+    """
+    params = list(params)
+
+    if binocular:
+        n_params = len(params)
+        augmented_params_0 = extend_params(
```
marc-tonsen: Can you reuse `MappingFunction2D` here?

pupil_src/shared_modules/calibration_routines/calibrate_2d.py
```diff
@@ -0,0 +1,264 @@
...
+    pupil_positions,
+    targets,
+    binocular,
+    dof=2,
+    degree=2,
+    ignored_terms=(),
+    regularization=0.0,
+    loss_scale=-1,
+):
+
```
marc-tonsen: blank line

pupil_src/shared_modules/calibration_routines/calibrate_2d.py
```diff
@@ -0,0 +1,264 @@
...
+    cost += regularization * np.sum(params ** 2)
+
+    return cost
+
+
+def cauchy_loss(residuals, loss_scale):
+    return loss_scale ** 2 * np.log(1 + residuals / loss_scale ** 2)
+
+
+def convert_to_linear_index(multi_idx, size_per_dim):
```
marc-tonsen: This function is used only in `extend_params`, which is defined *after* this function. Usually the caller should be defined first, so the overall structure of methods is `general -> specific`.

pupil_src/shared_modules/calibration_routines/calibrate_2d.py
```diff
@@ -0,0 +1,264 @@
...
+        cost = 1 / 2 * np.sum(cauchy_loss(squared_residuals, loss_scale))
+    else:
+        cost = 1 / 2 * np.sum(squared_residuals)
+
+    cost += regularization * np.sum(params ** 2)
+
+    return cost
+
+
+def cauchy_loss(residuals, loss_scale):
```
marc-tonsen: Python supports private methods to some extend. If you prepend a function name with `_`, e.g. `def _foo(): ...`, than this method is marked private (actually more similar to protected). Such methods are e.g. not imported if you import the parent module. Functions and methods which are not supposed to be used from the outside should be marked private to reduce the apparent complexity of your module/class. Please prepend your functions with `_` appropriately.

pupil_src/shared_modules/calibration_routines/calibrate_2d.py
```diff
@@ -0,0 +1,264 @@
...
+
+        self.map_to_world_cam_x = polynomial_type(params_x, dof, degree)
+        self.map_to_world_cam_y = polynomial_type(params_y, dof, degree)
+
+    def __call__(self, pt):
+
+        return self.map_to_world_cam_x(pt), self.map_to_world_cam_y(pt)
+
+
+def fit_mapping_polynomials(
```
marc-tonsen: The name kinda suggests that you return polynomials, but you return polynomial parameters. Mayb rename to `fit_polynomial_parameters` or `fit_polynomial_coefficients`. Alternatively, depending on how this function is used, could it make sense that this function in fact does return a polynomial function?

pupil_src/shared_modules/calibration_routines/calibrate_2d.py
```diff
@@ -0,0 +1,264 @@
...
+    binocular,
+    degree=2,
+    ignored_terms=(),
+    regularization=0.0,
+    loss_scale=-1,
+):
+    """
+    Fit two polynomials, one for each of the last two columns of the provided calibration data.
+
+    :param calibration_data: Array of floats. The last two columns correspond to screen coordinates of the
```
marc-tonsen: If `calibration_data` is such a complex object, would it make sense to split it up into separate arguments? Into e.g. `calibration_targets` and `calibration_data`?

pupil_src/shared_modules/calibration_routines/calibrate_2d.py
marc-tonsen: I like the overall structure of your functions and classes. They are both short and serve a clear purpose! :+1:

pupil_src/shared_modules/calibration_routines/data_processing.py
```diff
@@ -0,0 +1,200 @@
+import numpy as np
+
+# logging
```
marc-tonsen: remove comment

pupil_src/shared_modules/calibration_routines/data_processing.py
```diff
@@ -0,0 +1,200 @@
...
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def filter_confidence(pupil_list, threshold):
+
```
marc-tonsen: blank line

pupil_src/shared_modules/calibration_routines/data_processing.py
```diff
@@ -0,0 +1,200 @@
...
+            return idx - 1
+        else:
+            return idx
+    except IndexError:
+        return idx - 1
+
+
+def match_data(g_pool, pupil_list, ref_list):
+    """
+    Returns binocular and monocular matched pupil datums and ref points. Uses a dispersion criterion to
```
marc-tonsen: Line length

pupil_src/shared_modules/calibration_routines/data_processing.py
```diff
@@ -0,0 +1,200 @@
...
+    dismiss matches which are too far apart.
+    """
+    if pupil_list and ref_list:
```
marc-tonsen: Invert the condition and only test else-condition explicitly.

pupil_src/shared_modules/calibration_routines/data_processing.py
```diff
@@ -0,0 +1,200 @@
...
+        pupil1,
+    )
+
+
+def closest_matches_binocular(ref_pts, pupil_pts, max_dispersion=1 / 15.0):
+    """
+    get pupil positions closest in time to ref points.
+    return list of dict with matching ref, pupil0 and pupil1 data triplets.
+    """
+    pupil0 = [p for p in pupil_pts if p["id"] == 0]
```
marc-tonsen: This split has already been computed in `match_data`. You could user `pupil0` and `pupil1` as arguments already.

pupil_src/shared_modules/calibration_routines/data_processing.py
```diff
@@ -0,0 +1,200 @@
...
+        logger.info(
+            "Dismissing {:.2f}% pupil data due to confidence < {:.2f}".format(
+                dismissed_percentage, threshold
+            )
+        )
+
+    return pupil_list
+
+
+def find_nearest_idx(array, value):
```
marc-tonsen: Move definition behind callers.

pupil_src/shared_modules/calibration_routines/data_processing.py
```diff
@@ -0,0 +1,200 @@
...
+    idx = np.searchsorted(array, value, side="left")
+    try:
+        if abs(value - array[idx - 1]) < abs(value - array[idx]):
```
marc-tonsen: Why is this needed? Maybe deserves a comment.

pupil_src/shared_modules/calibration_routines/data_processing.py
```diff
@@ -0,0 +1,200 @@
...
+
+    if pupil0 and pupil1:
+        for r in ref_pts:
+
+            closest_p0_idx = find_nearest_idx(pupil0_ts, r["timestamp"])
+            closest_p0 = pupil0[closest_p0_idx]
+            closest_p1_idx = find_nearest_idx(pupil1_ts, r["timestamp"])
+            closest_p1 = pupil1[closest_p1_idx]
+
+            dispersion = max(
```
marc-tonsen: Would be a bit nicer when split into multiple commands/lines.

pupil_src/shared_modules/calibration_routines/data_processing.py
```diff
@@ -0,0 +1,200 @@
...
+        matched_monocular_data,
+        matched_pupil0_data,
+        matched_pupil1_data,
+        pupil0,
+        pupil1,
+    )
+
+
+def closest_matches_binocular(ref_pts, pupil_pts, max_dispersion=1 / 15.0):
+    """
```
marc-tonsen: Please make the comment a bit more verbose

pupil_src/shared_modules/calibration_routines/data_processing.py
```diff
@@ -0,0 +1,200 @@
...
+
+    return matched
+
+
+def closest_matches_monocular(ref_pts, pupil_pts, max_dispersion=1 / 15.0):
+    """
+    get pupil positions closest in time to ref points.
+    return list of dict with matching ref and pupil datum.
+
+    if your data is binocular use:
```
marc-tonsen: This comment is inappropriately placed.

pupil_src/shared_modules/calibration_routines/data_processing.py
```diff
@@ -0,0 +1,200 @@
...
+                matched.append({"ref": r, "pupil": closest_p0, "pupil1": closest_p1})
+            else:
+                logger.debug(
+                    "Binocular match rejected due to time dispersion criterion"
+                )
+
+    return matched
+
+
+def closest_matches_monocular(ref_pts, pupil_pts, max_dispersion=1 / 15.0):
```
marc-tonsen: The code of this function is very redundant with the binocular version. They can probably be merged or redundant code can be extracted.

pupil_src/shared_modules/calibration_routines/data_processing.py
```diff
@@ -0,0 +1,200 @@
...
+                matched.append({"ref": r, "pupil": closest_p0})
+            else:
+                logger.debug(
+                    "Monocular match rejected due to time dispersion criterion"
+                )
+
+    return matched
+
+
+def preprocess_2d_data_monocular(matched_data):
```
marc-tonsen: `preprocess` is a strong/very general word for what happens here. Maybe something like `extract_calibration_data` or `extract_relevant_fields` or so would be more appropriate.

pupil_src/shared_modules/calibration_routines/data_processing.py
```diff
@@ -0,0 +1,200 @@
...
+    Takes matched pupil data and returns list of tuples, keeping only the positions in normalized screen coordinates
+    for pupil and ref.
+    """
+    cal_data = [
+        (*pair["pupil"]["norm_pos"], *pair["ref"]["norm_pos"]) for pair in matched_data
+    ]
+    return cal_data
+
+
+def preprocess_2d_data_binocular(matched_data):
```
marc-tonsen: Again this is very redundant.

pupil_src/shared_modules/calibration_routines/finish_calibration.py
```diff
@@ -12,23 +12,30 @@
...
-from math_helper import *
-from file_methods import load_object, save_object
+import math_helper
+from file_methods import save_object
 
 from .optimization_calibration import bundle_adjust_calibration
-from .calibrate import find_rigid_transform
+from math_helper import find_rigid_transform
+from .data_processing import (
```
marc-tonsen: Consider only using `import .data_processing` and later user `data_processing.match_data`. This makes it explicit that these functions are defined somewhere else and the prepended namespace also gives an additional hint at what the function is doing.

Alongside this change I also wonder if `data_processing` is the most descriptive name.

pupil_src/shared_modules/calibration_routines/finish_calibration.py
```diff
@@ -12,23 +12,30 @@
...
+    match_data,
+    filter_confidence,
+    preprocess_2d_data_binocular,
+    preprocess_2d_data_monocular,
+    preprocess_3d_data,
+)
+from .calibrate_2d import fit_mapping_polynomials
+
 
 # logging
```
marc-tonsen: I don't know why we have this comment everywhere...

pupil_src/shared_modules/calibration_routines/calibrate_2d.py
papr: I think one could use [`np.ravel_multi_index`](https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.ravel_multi_index.html) as well.

pupil_src/shared_modules/calibration_routines/calibrate_2d.py
```diff
@@ -0,0 +1,264 @@
...
+        n_params = len(params)
+        augmented_params_0 = extend_params(
+            list(params[: n_params // 2]), degree, ignored_terms
+        )
+        augmented_params_1 = extend_params(
+            list(params[n_params // 2 :]), degree, ignored_terms
+        )
+        augmented_params = np.hstack((augmented_params_0, augmented_params_1))
+    else:
+        list(ignored_terms).sort(
```
papr: Bug: This line does not save the sorted result.
```suggestion
        ignored_terms = sorted(
            ignored_terms
```

romanroibu:
@papr and @marc-tonsen could you please review this PR? I merged the latest `master` and tried to resolve the conflicts as sensibly as I can. I also tried to incorporate your feedback. Please let me know if I've missed something important.

The Travis built is currently also failing on `master`, but @patrickFuerst is attempting to fix that.

pupil_src/shared_modules/math_helper/transformations.py
```diff
@@ -1980,6 +1983,36 @@ def _import_module(name, package=None, warn=True, prefix="_py_", ignore="_"):
...
+    if np.linalg.det(R) < 0:
+        logger.info("Reflection detected")
```
papr:
```suggestion
        logger.debug("Reflection detected")
```

pupil_src/shared_modules/calibration_routines/hmd_calibration.py
```diff
@@ -141,7 +143,7 @@ def finish_calibration(self):
 
         if matched_pupil0_data:
             cal_pt_cloud = calibrate.preprocess_2d_data_monocular(matched_pupil0_data)
-            map_fn0, inliers0, params0 = calibrate.calibrate_2d_polynomial(
+            map_fn0, inliers0, params0 = calibrate.calibrate_polynomial(
```
papr:
```suggestion
            map_fn0, inliers0, params0 = calibrate_2d.calibrate_polynomial(
```

pupil_src/shared_modules/calibration_routines/gaze_mappers.py
```diff
@@ -236,7 +236,7 @@ def deinit_ui(self):
 
     def _map_binocular(self, p0, p1):
         if self.multivariate:
-            gaze_point = self.map_fn(p0["norm_pos"], p1["norm_pos"])
+            gaze_point = self.map_fn([*p0["norm_pos"]] + [*p1["norm_pos"]])
```
papr: Please reevaluate if change is necessary. If yes, replace by `list(pX["norm_pos"])`

romanroibu:
@papr 

|                                    | 2d accuracy | 2d precision | 3d accuracy | 3d precision |
|:----------------------------|:---------------:|:----------------:|:---------------:|:----------------:|
| `master`                     | 0.2&deg; (3208/3247) | 0.1&deg; (3208/3246) | 3.3&deg; (2205/2238) | 0.1&deg; (2189/2237) |
| `calibration_refactor` | 0.8&deg; (3207/3247) | 0.1&deg; (3163/3246) | 3.6&deg; (2140/2238) | 0.1&deg; (2191/2237) |

papr:
Will be replaced with Gazer Refactor.

PR #1389, opened 2018-11-16
New Observable Class

ckbaumann:
BTW: this is going to a feature branch, not master

### DIFF
--- a/pupil_src/shared_modules/observable.py
+++ b/pupil_src/shared_modules/observable.py
@@ -1,0 +1,314 @@
+"""
+(*)~---------------------------------------------------------------------------
+Pupil - eye tracking platform
+Copyright (C) 2012-2018 Pupil Labs
+
+Distributed under the terms of the GNU
+Lesser General Public License (LGPL v3.0).
+See COPYING and COPYING.LESSER for license details.
+---------------------------------------------------------------------------~(*)
+"""
+
+import functools
+import inspect
+import weakref
+
+
+class ObserverError(Exception):
+    pass
+
+
+class Observable:
+    """
+    An object that allows to add observers to any of its methods. See
+    the docstring for add_observer for more information.
+    """
+
+    def add_observer(self, method_name, observer):
+        """
+        Adds an observer to a method. An observer is an arbitrary callable. Every time
+        the method is invoked, the observer will be called.
+
+        The observer is called with the same arguments as the method, so make sure
+        that the observer can handle these arguments!
+
+        You cannot add observers to static and class methods.
+
+        It's possible to add multiple observers to the same method.
+
+        Observers are called AFTER the method completed. Even if the method raises an
+        exception, observers are still called.
+        Observers should NOT raise exceptions. If they do, exceptions are wrapped
+        in an ObserverError and raised to the caller of the method.
+
+        You only add the observer to this instance of the class. The observer is not
+        triggered for calls to the same method of a different instance of the class.
+
+        If the observer is a bound (i.e. non-static) method, it is safe to delete the
+        corresponding object the observer is bound to. In this case, the observer
+        will be ignored automatically. No need to remove it before deletion.
+
+        Never use a bound method as an observer unless the corresponding object is
+        still referenced somewhere else. Observables will only keep weak references
+        to bound methods s.t. the objects can get deleted safely (see above).
+        However, this means that the object will get garbage collected if its only
+        referenced from an Observable. Use a callable class or lambda in this case.
+
+        Args:
+            method_name (String): The name of a bound method to which the observer
+                will be added. Unbound methods (i.e. static and class methods) are NOT
+                supported.
+            observer (Callable): Will be called every time the method is invoked.
+
+        Raises:
+            AttributeError: The attribute specified by method_name is not found in
+                the class
+            TypeError: The attribute specified by method_name is
+                1) no method, or
+                2) a class or static method.
+
+        """
+        add_observer(self, method_name, observer)
+
+    def remove_observer(self, method_name, observer):
+        """
+        Removes a specific observer from a method. You have to specify the same
+        callable as you added before.
+
+        Caveat: Removing lambdas does not work with this function. For example,
+        the following will fail:
+        > add_observer("foo", lambda: 1)
+        > remove_observer("foo", lambda: 1)
+        The two lambdas do the same but are two distinct objects.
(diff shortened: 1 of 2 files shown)


ckbaumann:
Please don't review yet, @papr proposed something better (and more evil)

pupil_src/shared_modules/observable_test.py
```diff
@@ -0,0 +1,309 @@
...
+    def class_method(cls):
+        pass
+
+
+@pytest.fixture()
+def observable():
+    return FakeObservable()
+
+
+class TestObservableMethods:
```
marc-tonsen:
```suggestion
class TestObservabilityOfMethods:
```
More accurate?

pupil_src/shared_modules/observable.py
```diff
@@ -0,0 +1,320 @@
...
+
+    Usually you should not use this directly. Instead, make your class inherit from
+    Observable and use the corresponding method there.
+
+    """
+    observable = _get_wrapper_and_create_if_not_exists(obj, method_name)
+    observable.add_observer(observer)
+
+
+def _get_wrapper_and_create_if_not_exists(obj, method_name):
```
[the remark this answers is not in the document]
marc-tonsen: @ckbaumann and I discussed this and in the end agreed that this change makes sense using the following argument: The problem with the `and` is that it does in fact not do both, `get` *and* `create`. It either returns a previously created wrapper or if there is none, it creates one. It does always return a wrapper, that is correct, but from a function that is called `create_wrapper` I would also expect to get a wrapper back, i.e. the get is not needed on top of the create in the create case.

pupil_src/shared_modules/observable.py
```diff
@@ -0,0 +1,320 @@
...
+
+        It's possible to add multiple observers to the same method.
+
+        Observers are called AFTER the method completed. Even if the method raises an
+        exception, observers are still called.
+        Observers should NOT raise exceptions. If they do, exceptions are forwarded
+        to the caller of the method.
+
+        You only add the observer for this object. The observer is not triggered for
+        the same method of a different object.
```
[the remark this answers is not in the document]
ckbaumann: My feeling is when you just state a method and a callback, then people could think you are patching it for everywhere, so I just added it to make it clearer. You are right that this will be clear to most people, but I think this short sentence is ok here.

pupil_src/shared_modules/observable.py
```diff
@@ -0,0 +1,316 @@
...
+
+def _get_wrapper_and_create_if_not_exists(obj, method_name):
+    observed_method = getattr(obj, method_name)
+    method_already_wrapped = isinstance(observed_method, _ObservableMethodWrapper)
+    # this case has to be tested first, because if it is already wrapped,
+    # its not a method anymore and the next case would raise an exception!
+    if method_already_wrapped:
+        return observed_method
+    elif not inspect.ismethod(observed_method):
+        raise ValueError(
```
papr:
```suggestion
        raise TypeError(
```

pupil_src/shared_modules/observable.py
```diff
@@ -0,0 +1,316 @@
...
+            "Attribute {} of object {} is not a method but {} and, thus, "
+            "cannot be observed!".format(method_name, obj, observed_method)
+        )
+    elif _is_classmethod(obj, method_name):
+        raise ValueError(
```
papr:
```suggestion
        raise TypeError(
```

pupil_src/shared_modules/observable.py
```diff
@@ -0,0 +1,316 @@
...
+
+    """
+    observable = _get_wrapper_or_raise_if_not_exists(obj, method_name)
+    observable.remove_all_observers()
+
+
+def _get_wrapper_or_raise_if_not_exists(obj, method_name):
+    observable_wrapper = getattr(obj, method_name)
+    if not isinstance(observable_wrapper, _ObservableMethodWrapper):
+        raise ValueError(
```
papr:
```suggestion
        raise TypeError(
```

pupil_src/shared_modules/observable.py
```diff
@@ -0,0 +1,316 @@
...
+                "its method. Your wrapper was probably referenced "
+                "elsewhere and called via this reference after you "
+                "removed the wrapper!"
+            )
+        try:
+            return self._original_method(*args, **kwargs)
+        except Exception:
+            raise
+        finally:
+            self.call_all_observers(args, kwargs)
```
papr: Isn't this nice :)

pupil_src/shared_modules/observable.py
```diff
@@ -0,0 +1,316 @@
...
+        for observer in self._observers:
+            try:
+                observer(*args, **kwargs)
+            except _ReferenceNoLongerValid:
+                self._observers.remove(observer)
+            except Exception as e:
+                raise ObserverError("An observer raised an exception.") from e
+
+
+class _ReferenceNoLongerValid(Exception):
```
papr:
```suggestion
class _ReferenceNoLongerValidError(Exception):
```

pupil_src/shared_modules/observable.py
```diff
@@ -0,0 +1,316 @@
...
+        self._obj_ref = weakref.ref(observer.__self__)
+        self._func_ref = weakref.ref(observer.__func__)
+
+    def __call__(self, *args, **kwargs):
+        obj_deref = self._obj_ref()
+        func_deref = self._func_ref()
+        try:
+            method = getattr(obj_deref, func_deref.__name__)
+        except AttributeError:
+            raise _ReferenceNoLongerValid
```
papr:
```suggestion
            raise _ReferenceNoLongerValidError
```

pupil_src/shared_modules/observable.py
```diff
@@ -0,0 +1,316 @@
...
+
+    def __eq__(self, other_observer):
+        if inspect.ismethod(other_observer):
+            obj_deref = self._obj_ref()
+            func_deref = self._func_ref()
+            if self._method_still_exists(func_deref, obj_deref):
+                return (
+                        obj_deref == other_observer.__self__
+                        and func_deref == other_observer.__func__
+                )
```
papr: Maybe refactor into 3 lines: 2x boolean value assignments, 1x return

pupil_src/shared_modules/observable.py
ckbaumann: Yes :)

PR #1400, opened 2018-11-26
New Background Tasks Module

ckbaumann:
depends on https://github.com/pupil-labs/pupil/pull/1389

### DIFF
--- a/pupil_src/shared_modules/tasklib/__init__.py
+++ b/pupil_src/shared_modules/tasklib/__init__.py
@@ -1,0 +1,12 @@
+"""
+(*)~---------------------------------------------------------------------------
+Pupil - eye tracking platform
+Copyright (C) 2012-2018 Pupil Labs
+
+Distributed under the terms of the GNU
+Lesser General Public License (LGPL v3.0).
+See COPYING and COPYING.LESSER for license details.
+---------------------------------------------------------------------------~(*)
+"""
+
+from tasklib.observers import raise_exception
--- a/pupil_src/shared_modules/tasklib/background/__init__.py
+++ b/pupil_src/shared_modules/tasklib/background/__init__.py
@@ -1,0 +1,12 @@
+"""
+(*)~---------------------------------------------------------------------------
+Pupil - eye tracking platform
+Copyright (C) 2012-2018 Pupil Labs
+
+Distributed under the terms of the GNU
+Lesser General Public License (LGPL v3.0).
+See COPYING and COPYING.LESSER for license details.
+---------------------------------------------------------------------------~(*)
+"""
+
+from tasklib.background.create import create
--- a/pupil_src/shared_modules/tasklib/background/create.py
+++ b/pupil_src/shared_modules/tasklib/background/create.py
@@ -1,0 +1,67 @@
+"""
+(*)~---------------------------------------------------------------------------
+Pupil - eye tracking platform
+Copyright (C) 2012-2018 Pupil Labs
+
+Distributed under the terms of the GNU
+Lesser General Public License (LGPL v3.0).
+See COPYING and COPYING.LESSER for license details.
+---------------------------------------------------------------------------~(*)
+"""
+
+import inspect
+
+from tasklib.background.task import BackgroundGeneratorFunction
+from tasklib.background.task import BackgroundRoutine
+
+
+def create(
+    name,
+    routine_or_generator_function,
+    pass_shared_memory=False,
+    args=None,
+    kwargs=None,
+    patches=None,
+):
+    """
+    Creates the right background task for your type of task.
+
+    Normally, you would not use this directly, but use a PluginTaskManager and the
+    create_background_task() method there.
+    If you don't use a manager, you need to regularly call update() on the returned task
+    to trigger callbacks etc.
+
+    See the docstring for PluginTaskManager.create_background_task()
+    (in tasklib.manager.py) for information about the different parameters!
+    """
+    if args is None:
+        args = ()
+    if kwargs is None:
+        kwargs = {}
+    if patches is None:
+        patches = []
+
+    if inspect.isgeneratorfunction(routine_or_generator_function):
+        return BackgroundGeneratorFunction(
+            name,
+            routine_or_generator_function,
+            pass_shared_memory,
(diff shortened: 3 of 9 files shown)


pupil_src/shared_modules/background_task/task.py
```diff
@@ -0,0 +1,175 @@
...
+        import traceback
+
+        pipe_send.send(_TaskExceptionSignal(e, traceback.format_exc()))
+    else:
+        pipe_send.send(_TaskCompletedSignal(return_value=None))
+    finally:
+        pipe_send.close()
+
+
+class BackgroundRoutine(BackgroundTaskBase):
```
marc-tonsen: Could it make sense to overwrite extend `cancel_gracefully` and throw an error since a routine can not be cancelled gracefully, respectively it would not make sense to call this?

pupil_src/shared_modules/background_task/abstract_task.py
```diff
@@ -0,0 +1,181 @@
...
+Lesser General Public License (LGPL v3.0).
+See COPYING and COPYING.LESSER for license details.
+---------------------------------------------------------------------------~(*)
+"""
+
+import abc
+from observable import Observable
+
+
+class AbstractBackgroundTask(Observable, metaclass=abc.ABCMeta):
```
marc-tonsen:
```suggestion
class BackgroundTaskInterface(Observable, metaclass=abc.ABCMeta):
```

pupil_src/shared_modules/background_task/task.py
ckbaumann: Well, it makes sense to cancel_gracefully() routines. Even if they continue normally, no on_completed will be called if it was canceled, so you can safely just start the task again with different parameters.

Additionally, now you can cancel routines by passing shared memory and looking into the `should_terminate_flag` (see latest commit).

marc-tonsen: Makes sense!

pupil_src/shared_modules/task_lib/background/shared_memory.py
```diff
@@ -0,0 +1,48 @@
...
+"""
+
+
+class SharedMemory:
+    """
+    Contains information shared between the foreground and a background process.
+    All properties are process safe.
+    """
+
+    def __init__(self, should_terminate_flag, progress):
```
papr: On init `should_terminate_flag` and `progress` are shared cvalue objects while the equally-named properties are accessors for the `.value` attribute. This seems to be inconsistent. Maybe, SharedMemory should create the shared cvalue objects on its own and assign the arguments as values.

pupil_src/shared_modules/task_lib/background/task.py
```diff
@@ -0,0 +1,190 @@
...
+from collections import namedtuple
+from ctypes import c_bool, c_double
+
+from task_lib.interface import TaskInterface
+from task_lib.background.shared_memory import SharedMemory
+
+_TaskYieldSignal = namedtuple("_TaskYieldSignal", "datum")
+_TaskCompletedSignal = namedtuple("_TaskCompletedSignal", "return_value")
+_TaskCanceledSignal = namedtuple("_TaskCanceledSignal", "")
+_TaskExceptionSignal = namedtuple("_TaskExceptionSignal", "exception traceback")
```
papr: I would prefer passing a tuple of strings instead of a string with spaces. It feels more explicit to me.

pupil_src/shared_modules/task_lib/background/task.py
```diff
@@ -0,0 +1,190 @@
...
+            raise ValueError(
+                "Received unknown signal {} from background " "process".format(signal)
+            )
+
+    def _handle_signal_if_canceled(self, signal):
+        if (
+            isinstance(signal, _TaskCanceledSignal)
+            or isinstance(signal, _TaskCompletedSignal)
+            or isinstance(signal, _TaskExceptionSignal)
+        ):
```
papr: This can be replaced with a single line: `isinstance(signal, (SignalCls, ...))`

pupil_src/shared_modules/task_lib/interface.py
```diff
@@ -0,0 +1,190 @@
...
+        - The foreground process blocks until the process terminates or the grace
+          period is over.
+        - You likely break resources shared between your background process and the
+          foreground.
+        - You can be sure that on_canceled_or_killed() gets called immediately and not
+          during the next update().
+
+        Args:
+            grace_period (Number): Seconds to wait for the process to shut itself
+                down. If None, there is no grace period.
```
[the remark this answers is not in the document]
ckbaumann: Technically yes, with `0` it will trigger a `cancel_gracefully`, with `None` it doesn't. Practically, it doesn't make a difference. I thought that `kill(grace_period=None)` reads better.

pupil_src/shared_modules/task_lib/manager.py
```diff
@@ -0,0 +1,126 @@
...
+            routine_or_generator_function,
+            pass_shared_memory,
+            args,
+            kwargs,
+            patches,
+        )
+        self._scheduled_tasks.append(task)
+        return task
+
+    def add_task(self, task):
```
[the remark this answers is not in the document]
ckbaumann: Well, you can already start it before you call this. Yes, `add` is a little ambiguous, but I hoped it would be not too surprising that a manager is starting a task you added if you didn't. Thinking about this, I should probably rename `scheduled_tasks` to `fresh_tasks` or what do you think?

pupil_src/shared_modules/task_lib/background/shared_memory.py
ckbaumann: Yes, good point!

pupil_src/shared_modules/task_lib/manager.py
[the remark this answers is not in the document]
papr: `waiting_tasks`, `recently_added_tasks`, `recently_scheduled_tasks`?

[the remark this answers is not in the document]
ckbaumann: Thanks, good suggestions!

PR #1412, opened 2018-12-07
Add bundle.bat

papr:


willpatera:
@papr I would have to test this myself on a Windows bundling env VM and will not be able to do so until Tuesday at earliest. Perhaps @pointcontrols would be the better reviewer for this PR.

pointcontrols:
@papr Maybe not hardcode 7zip path like that ?

papr:
I agree, but without the full path, the command was not found. Not sure if one has to modify some PATH variables first...

pointcontrols:
Yes. The 7zip dir needs to be added to the system path on each machine
where bundling is going to be done.

papr:
Ok, I will make this change tomorrow.

PR #1417, opened 2018-12-12
bpo-1405:Let Pupil Player support play video from multiple resource v…

Windsooon:
Except for the feature itself, I refactor some functions for easy understanding.  In test_file_backend.py, the test didn't pass because the makefile function can't create a valid mp4 file, so the av.open() function will raise an error. We can choose 

1. @patch the av.open() function as well as mock the return object. 
2. Use real-world data for testing. 

The first option takes more time to write tests code but less time to run tests.

### DIFF
--- a/pupil_src/shared_modules/video_capture/file_backend.py
+++ b/pupil_src/shared_modules/video_capture/file_backend.py
@@ -121,6 +121,9 @@ class Decoder(ABC):
             int(self.video_stream.format.height),
         )
 
+    def cleanup(self):
+        pass
+
 
 class BufferedDecoder(Decoder):
     def __init__(self, container, video_stream):
@@ -139,6 +142,9 @@ class BufferedDecoder(Decoder):
 
     def next_frame(self):
         return self.buffered_decoder.get_frame()
+
+    def cleanup(self):
+        self.buffered_decoder.stop_buffer_thread()
 
 
 class OnDemandDecoder(Decoder):
@@ -220,6 +226,10 @@ class File_Source(Playback_Source, Base_Source):
         return rec, set_name
 
     def setup_video(self, container_index):
+        try:
+            self.video_stream.cleanup()
+        except AttributeError:
+            pass
         self.current_container_index = container_index
         self.container = self.videoset.containers[container_index]
         self.video_stream, self.audio_stream = self._get_streams(
@@ -531,7 +541,7 @@ class File_Source(Playback_Source, Base_Source):
 
     def cleanup(self):
         try:
-            self.video_stream.buffered_decoder.stop_buffer_thread()
+            self.video_stream.cleanup()
         except AttributeError:
             pass
         super().cleanup()


pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -111,78 +112,16 @@ def __init__(
...
-            logger.debug("No audiostream found in media container")
-
-        if not self.video_stream and not self.audio_stream:
-            logger.error(
-                "Init failed. Could not find any video or audio stream in the given source file."
-            )
-            self._initialised = False
-            return
+        self.containers_path = self._get_conatiners_path(source_path)
+        self.current_index = 0
```
papr: Is this the current frame or container index?

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -111,78 +112,16 @@ def __init__(
...
+        self.container = self._get_containers(self.current_index)
+        self.video_stream, self.audio_stream = self._get_streams(
+            self.container)
+        self.timestamps = self._get_timestamps_lst(source_path)
```
papr: `_get_timestamps_lst` returns `None`

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -220,6 +159,112 @@ def run_func(self, *args, **kwargs):
...
+            logger.info("The vide is broken")
+            self.current_index += 1
+            return self._get_containers(self.current_index)
+        except IndexError:
+            logger.info("No More Container Found")
+            raise EndofVideoError("Reached end of video file")
+        else:
+            return container
+
+    def _get_timestamps_lst(self, source_path):
```
papr:
```suggestion
    def _set_timestamps_lst(self, source_path):
```
The function returns `None`

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -220,6 +159,112 @@ def run_func(self, *args, **kwargs):
...
+        return self._get_pattern_lst(source_path)
+
+    def _get_containers(self, index):
+        '''
+        Get container by index, find the next one if the prev one is broken.
+        '''
+        try:
+            container = av.open(self.containers_path[index])
+        except av.AVError:
+            logger.info("The vide is broken")
```
papr:
```suggestion
            logger.info("The video is broken")
```
I think it would be nice if we could log that error once, including the file name, and then skip the broken files afterwards.

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -220,6 +159,112 @@ def run_func(self, *args, **kwargs):
...
+            self.current_index += 1
+            return self._get_containers(self.current_index)
+        except IndexError:
+            logger.info("No More Container Found")
+            raise EndofVideoError("Reached end of video file")
+        else:
+            return container
```
papr: What do you think about placing this line in the `try` block. Else the actual action of this function (opening and returning a single container) is split by a huge exception handling clause.

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -220,6 +159,112 @@ def run_func(self, *args, **kwargs):
 
         return decorator
 
+    def _get_pattern_lst(self, source_path, timestamps=False):
+        '''
+        Get an order list follow the name pattern
+        '''
+        if timestamps:
+            pattern, ext = os.path.basename(
+                os.path.normpath(source_path)).split(".")
```
papr: I would highly recommend to use https://docs.python.org/3/library/os.path.html#os.path.splitext to split extensions.

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -220,6 +159,112 @@ def run_func(self, *args, **kwargs):
...
+            suffix = "_timestamps.npy"
+        else:
+            pattern, suffix = os.path.basename(
+                os.path.normpath(source_path)).split(".")
+            suffix = "." + suffix
+        path = []
+        for file in os.listdir(os.path.dirname(source_path)):
+            if fnmatch.fnmatch(file, pattern + "*" + suffix):
+                path.append(file)
```
papr: I would recommend to use the https://docs.python.org/3/library/glob.html#module-glob package for this kind of work

pupil_src/shared_modules/video_capture/file_backend.py
Windsooon: Yes, I will update the log later.

pupil_src/shared_modules/video_capture/file_backend.py
Windsooon: Yes, you are right.

pupil_src/shared_modules/video_capture/file_backend.py
Windsooon: Ok, I will update it.

Windsooon:
This version works on non-split data now, and we can run the tests with

    python -m pytest video_capture/tests

under pupil/pupil_src/shared_modules dir

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -187,31 +181,36 @@ def _get_containers(self, index):
...
 
         if float(avg_rate) % 1 != 0.0:
             logger.error(
-                "Videofile pts are not evenly spaced, pts to index conversion may fail and be inconsitent."
+                "Videofile pts are not evenly spaced, pts to" +
+                "index conversion may fail and be inconsitent."
             )
         try:
-            self.timestamps = np.load(timestamps[0])
+            self.timestamps = np.load(self.timestamps_lst[0])
```
papr: I have just realised that we will require the timestamps of all partial timestamp files in `self.timestamps`.

Windsooon: Yes, for displays the whole time of the bar at the bottom of the screen, right? I will use other var to store the timestamp of the current video later.

papr: Yes, and some other places, too.

Windsooon: Ok, I see. so self.timestamps should contain all timestamp from the different part of videos. And we use another var to store the timestamp of the current video.

papr: Yes, I think this is a good solution.

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -220,6 +162,115 @@ def run_func(self, *args, **kwargs):
...
+    def _get_pattern_lst(self, source_path, timestamps=False):
+        '''
+        return an order list sort by file name pattern
+        '''
+        if timestamps:
+            # [world_timestamps.npy, world_001_timestamps.npy, world_002_timestamps.npy]
+            pattern, _ = os.path.splitext(source_path)
+            suffix = "_timestamps.npy"
+            tmp = sorted(glob.glob(pattern + '*' + suffix))
+            return tmp[-1:] + tmp[:-1]
```
papr: Please add a comment to facilitate the understanding of what this line does and why it is required.

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -220,6 +162,115 @@ def run_func(self, *args, **kwargs):
...
+            avg_rate = Fraction(0, 1)
+
+        if float(avg_rate) % 1 != 0.0:
+            logger.error(
+                "Videofile pts are not evenly spaced, pts to" +
+                "index conversion may fail and be inconsitent."
+            )
+        self.timestamps = np.array([])
+        try:
+            for timestamp in self.timestamps_lst:
```
papr:
```suggestion
             for ts_filename in self.timestamps_lst:
```

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -307,43 +387,35 @@ def get_frame(self):
...
-        try:
-            frame = frame or self.get_frame()
-        except EndofVideoError:
-            logger.info("Video has ended.")
-            self.g_pool.seek_control.play = False
-            frame = frame or self._recent_frame.copy()
-
+        # Normla Case to get next frame
+        if not self.play:
+            frame = self._recent_frame
```
papr: We might need to make a copy here since there are plugins that modify the decoded image buffer. I will try to confirm this in my tests later.

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -372,6 +444,21 @@ def recent_events_own_timing(self, events):
...
+        container_index = 0
+        # self.frame_count contain frame_count from every video
+        # like [50, 100, 150], so when we get a seek_pos 120, we
+        # calculate which video we should play
+
+        for i in range(len(self.frame_count)-1, 0, -1):
+            if seek_pos >= sum(self.frame_count[:i]):
+                    seek_pos = seek_pos - sum(self.frame_count[:i])
+                    container_index = i
+                    break
```
papr: This block got indented once too much :)

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -372,6 +444,21 @@ def recent_events_own_timing(self, events):
...
     def seek_to_frame(self, seek_pos):
         # frame accurate seeking
+        ori_pos = seek_pos
+        container_index = 0
+        # self.frame_count contain frame_count from every video
+        # like [50, 100, 150], so when we get a seek_pos 120, we
+        # calculate which video we should play
+
+        for i in range(len(self.frame_count)-1, 0, -1):
+            if seek_pos >= sum(self.frame_count[:i]):
```
papr: This seems a bit too complicated to find the correct index. I suggest using `np.cumsum` + `bisect`/`np.searchsorted` to find the correct container index.

pupil_src/shared_modules/video_capture/file_backend.py
Windsooon: Which one we should make a copy here?

papr: `frame = self._recent_frame.copy()`

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -220,6 +172,142 @@ def run_func(self, *args, **kwargs):
...
+        ori_timestamps = self.timestamps.copy()
+        self.timestamps = self._interpolate_timestamps(self.timestamps)
+        self.timestamps_mask = np.in1d(self.timestamps, ori_timestamps)
+
+    def _interpolate_timestamps(self, timestamps):
+        first = np.load(self.timestamps_lst[0])
+        frame_rate = (first[-1] - first[0]) / len(first)
+        for i, t in np.ndenumerate(timestamps[:-1]):
+            if timestamps[i[0]+1] - timestamps[i[0]] > 0.5:
+                timestamps = np.append(
```
papr: `np.append()` is very slow. I would recommend using a list to aggregate timestamps:
```python
ts_aggregate = [timestamps]
for gap in gaps:
    filling_timestamps = interpolate_gap(gap)
    ts_aggregate.append(filling_timestamps)
timestamps = np.concatenate(ts_aggregate)
timestamps.sort()
```

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -340,6 +339,8 @@ def get_next_frame(self):
     def get_frame(self):
         target_entry = self.videoset.lookup[self.target_frame_idx]
         if target_entry.container_idx == -1:
+            self.current_frame_idx = self.target_frame_idx
+            self.target_frame_idx += 1
```
papr: Ah nice catch!

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -30,11 +31,13 @@
...
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
-
-assert av.__version__ >= "0.4.2", "pyav is out-of-date, please update"
 av.logging.set_level(av.logging.ERROR)
 logging.getLogger("libav").setLevel(logging.ERROR)
 
+assert av.__version__ >= "0.4.2", "pyav is out-of-date, please update"
+assert sys.version_info >= (3, 7), \
+    ("We need python 3.7 or later to run Pupil player")
```
papr: TODO: Make PR compatible with Python 3.5, i.e. remove dataclasses and type annotations.

pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -282,7 +282,7 @@ def _loaded_ts_sorted(self) -> np.ndarray:
     def _fill_gaps(self, timestamps: np.ndarray) -> np.ndarray:
         time_diff = np.diff(timestamps)
         median_time_diff = np.median(time_diff)
-        gap_start_idc = np.flatnonzero(time_diff > 2 * median_time_diff)
+        gap_start_idc = np.flatnonzero(time_diff > 5 * median_time_diff)
```
papr: This change would result in only gaps with 4 or more dropped frames being filled, assuming them being filled with `median_time_diff` distance.

pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -155,3 +165,144 @@ def check_slice(self, frame_gray):
...
+        all_ts = np.concatenate(loaded_ts)
+        return all_ts
+
+    def _fill_gaps(self, timestamps: np.ndarray) -> np.ndarray:
+        time_diff = np.diff(timestamps)
+        median_time_diff = np.median(time_diff)
+        gap_start_idc = np.flatnonzero(time_diff > 5 * median_time_diff)
+        gap_stop_idc = gap_start_idc + 1
+        gap_fill_starts = timestamps[gap_start_idc] + median_time_diff
+        gap_fill_stops = timestamps[gap_stop_idc]
```
Windsooon: We should use `timestamps[gap_stop_idc]` instead of `timestamps[gap_stop_idc] - median_time_diff` here if I understand correctly

pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -155,3 +165,144 @@ def check_slice(self, frame_gray):
...
+
+    def save_lookup(self):
+        np.save(self.lookup_loc, self.lookup)
+
+    def load_lookup(self):
+        self.lookup = np.load(self.lookup_loc).view(np.recarray)
+
+    def _loaded_ts_sorted(self) -> np.ndarray:
+        loaded_ts = [vid.timestamps for vid in self.videos]
+        all_ts = np.concatenate(loaded_ts)
```
Windsooon: We don't have the sort all_ts here, right?

pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -155,3 +165,144 @@ def check_slice(self, frame_gray):
...
+        time_diff = np.diff(timestamps)
+        median_time_diff = np.median(time_diff)
+        gap_start_idc = np.flatnonzero(time_diff > 5 * median_time_diff)
+        gap_stop_idc = gap_start_idc + 1
+        gap_fill_starts = timestamps[gap_start_idc] + median_time_diff
+        gap_fill_stops = timestamps[gap_stop_idc]
+
+        all_ts = [timestamps]
+        for start, stop in zip(gap_fill_starts, gap_fill_stops):
+            all_ts.append(np.arange(start, stop, median_time_diff))
```
Windsooon: Should we use numpy's array instead of list for speed up?

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -307,64 +318,59 @@ def idx_to_pts(self, idx):
...
+        elif target_entry.container_idx != self.current_container_index:
+            self.setup_video(target_entry.container_idx)
+        try:
+            av_frame = next(self.next_frame)
+            if not av_frame:
+                raise EndofVideoError
+            index = self._convert_frame_index(av_frame.pts)
+        except StopIteration as stop:
+            return self._get_fake_frame_and_advance(target_entry.timestamp)
+            raise EndofVideoError from stop
```
papr: Do we still need this line? And should be returning a fake frame? What if `fill_gaps==False`?

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -373,6 +371,11 @@ def recent_events_own_timing(self, events):
             frame = self.get_frame()
         except EndofVideoError:
             logger.info("Video has ended.")
+            if self.loop:
```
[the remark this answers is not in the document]
papr: OPD should not loop. Loop is exposed as a ui switch to the user when the file source is used in Capture.

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -307,64 +318,59 @@ def idx_to_pts(self, idx):
...
+            frame = self._recent_frame.copy()
+        elif ts_idx < last_index or ts_idx > last_index + 1:
             self.seek_to_frame(ts_idx)
 
         # Normal Case to get next frame
-        if not self.play:
+        try:
+            frame = frame or self.get_frame()
+        except EndofVideoError:
+            if self.loop:
```
papr: `loop` does not need to be supported in `recent_events_external_timing`. I should have made this more clear before.

pupil_src/shared_modules/video_capture/utils.py
papr: We do since `_fill_gaps()` as well as the following code expect a sorted list.

pupil_src/shared_modules/video_capture/utils.py
papr: This is correct since we use `np.arange` instead of `np.linspace`.

pupil_src/shared_modules/video_capture/utils.py
papr: Not in this case, since appending numpy arrays is very slow in comparison to adding entries to python lists. `np.concatenate` appends all single numpy arrays at once in the end.

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -22,21 +22,15 @@
...
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 av.logging.set_level(av.logging.ERROR)
 logging.getLogger("libav").setLevel(logging.ERROR)
 
 assert av.__version__ >= "0.4.2", "pyav is out-of-date, please update"
-assert sys.version_info >= (3, 7), \
-    ("We need python 3.7 or later to run Pupil player")
+assert sys.version_info >= (3, 7), "We need python 3.7 or later to run Pupil player"
```
papr: I think this line can be removed by now

pupil_src/shared_modules/video_capture/file_backend.py
Windsooon: Did you mean 
    `return self._get_fake_frame_and_advance(target_entry.timestamp)`
Yes, I think we don't need it anymore.

pupil_src/shared_modules/video_capture/file_backend.py
Windsooon: Oh, I missed understand, so we can delete all the loop part, right?

pupil_src/shared_modules/video_capture/file_backend.py
Windsooon: > We do since _fill_gaps() as well as the following code expect a sorted list.

Since 

```
    @property
    def videos(self) -> Sequence[Video]:
        if self._videos is None:
            self._videos = sorted(self.fetch_videos(), key=lambda v: v.path)
        return self._videos
```
self._videos supposed been sorted already, so `loaded_ts = [vid.timestamps for vid in self.videos]` should also sorted by default if I understand it correctly

pupil_src/shared_modules/video_capture/file_backend.py
papr: Correct, that is the line that I meant.

pupil_src/shared_modules/video_capture/file_backend.py
papr: Yes, I did not have this in mind. Please write a specific test that makes sure that the assumption holds true for a given recording.

pupil_src/shared_modules/video_capture/file_backend.py
Windsooon: > OPD should not loop.
But we still have to keep this `loop` code in `recent_events_own_timing`, right?

papr: Correct, that is required for File_Source usage in Capture.

pupil_src/shared_modules/pupil_producers.py
```diff
@@ -301,23 +302,23 @@ def start_eye_process(self, eye_id):
             for ext in (".mjpeg", ".mp4", ".mkv")
         ]
         existing_locs = [loc for loc in potential_locs if os.path.exists(loc)]
-        timestamps_path = os.path.join(
-            self.g_pool.rec_dir, "eye{}_timestamps.npy".format(eye_id)
-        )
+        timestamps_path_lst = glob.glob(os.path.join(
+            self.g_pool.rec_dir, "eye{}*_timestamps.npy".format(eye_id)))
+        timestamp_len = sum(len(np.load(t)) for t in timestamps_path_lst)
```
papr: We should load the lookup table and filter it as if `fill_gaps==False`. Loading the single timestamp files does not let you know if there are broken containers.

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -82,16 +80,16 @@ class FakeFrame:
...
 
-    static_img = np.ones((720, 1280, 3), dtype=np.uint8) * 128
-
-    def __init__(self, timestamp, index):
+    def __init__(self, shape, timestamp, index):
+        self.shape = shape
         self.yuv_buffer = None
-        self.img = self.bgr = self.gray = self.static_img.copy()
+        static_img = np.ones(self.shape, dtype=np.uint8) * 128
+        self.img = self.bgr = self.gray = static_img.copy()
```
papr: There is no need to create a copy here since `static_img` is newly instantiated everytime.

pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -218,7 +218,15 @@ def videos(self) -> Sequence[Video]:
 
     @property
     def containers(self) -> Sequence[Video]:
-        return [av.open(video.path) for video in self.videos]
+        lst = []
+        for video in self.videos:
+            try:
+                cont = av.open(video.path)
```
papr: Did you mean to use `check_container` here? We could change `check_container` to handle the failure case and return `None` in these cases.

```python
# Video Class
def check_container(self):
    try:
        cont = av.open(self.path)
        cont.streams.video[0]
        frame = next(cont.decode(video=0))
    except av.AVError:
        return None
    return cont

# VideoSet Class
@property
def containers(self):
    return [video.check_container() for video in self.videos]
```

pupil_src/shared_modules/video_export/plugin_base/isolated_frame_exporter.py
```diff
@@ -78,7 +78,10 @@ def _convert_video_file(
     timestamp_export_format,
 ):
     yield "Export video", 0.0
-    input_source = File_Source(EmptyGPool(), input_file)
+    input_source = File_Source(
+        EmptyGPool(), input_file,
+        buffered_decoding=True,
```
papr: Exporters do not need buffering.

pupil_src/shared_modules/video_export/plugins/world_video_exporter.py
```diff
@@ -170,7 +170,11 @@ def _export_world_video(
             )
         except StopIteration:
             raise FileNotFoundError("No Video world found")
-        cap = init_playback_source(g_pool, source_path=video_path, timing=None)
+        cap = init_playback_source(
+            g_pool, source_path=video_path,
+            buffered_decoding=True,
```
papr: Exporters do not need buffering.

pupil_src/shared_modules/pupil_producers.py
```diff
@@ -302,21 +301,24 @@ def start_eye_process(self, eye_id):
...
             self.detection_status[eye_id] = "No eye video found."
             return
-        if not timestamps_path_lst:
+        rec, file_ = os.path.split(existing_locs[0])
+        set_name = os.path.splitext(file_)[0]
+        self.videoset = VideoSet(rec, set_name, fill_gaps=False)
+        try:
+            self.videoset.load_lookup()
+        except FileNotFoundError:
+            self.videoset.build_lookup()
```
papr: Since these for lines seem to repeat every time we want a lookup table: Let's add a `load_or_build_lookup()` function that handles this internally.

pupil_src/shared_modules/pupil_producers.py
```diff
@@ -302,21 +301,24 @@ def start_eye_process(self, eye_id):
...
-            self.g_pool.rec_dir, "eye{}*_timestamps.npy".format(eye_id)))
-        timestamp_len = sum(len(np.load(t)) for t in timestamps_path_lst)
-
         if not existing_locs:
             logger.error("no eye video for eye '{}' found.".format(eye_id))
             self.detection_status[eye_id] = "No eye video found."
             return
-        if not timestamps_path_lst:
+        rec, file_ = os.path.split(existing_locs[0])
+        set_name = os.path.splitext(file_)[0]
```
papr: What do you think about a utility function to extract the set name for a given path?

pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -173,8 +173,14 @@ def __init__(self, path: str) -> None:
...
-        cont = av.open(self.path)
-        frame = next(cont.decode(video=0))
+        try:
+            cont = av.open(self.path)
+            n = cont.decode(video=0)
+            _ = next(n)
+        except av.AVError:
+            return
+        else:
+            return av.open(self.path)
```
papr: Why do you call `open()` a second time? Isn't it faster to seek back to the zero'th frame and return `cont`?

pupil_src/shared_modules/pupil_producers.py
Windsooon: Yes, that would be better.

pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -155,3 +165,152 @@ def check_slice(self, frame_gray):
...
+            self.load_ts()
+        return self.ts
+
+
+class VideoSet:
+    def __init__(self, rec: str, name: str, fill_gaps: bool):
+        self.rec = rec
+        self.name = name
+        self.fill_gaps = fill_gaps
+        self.video_exts = ("mp4", "mjpeg", "h264")
```
papr:
```suggestion
        self.video_exts = ("mp4", "mjpeg", "h264", "mkv", "avi")
```
We need to add some more legacy ext. They do not require specific tests.

pupil_src/shared_modules/video_capture/file_backend.py
```diff
@@ -104,6 +104,43 @@ def gray(self):
         return self.img[:, :, 0]  # return first channel
 
 
+class Buffering:
```
papr: I like this abstraction but I think we did not unlock its full potential.

`Buffering.state` is basically a wrapper for `File_Source.video_stream`. Therefore, my suggestion would be modify `File_Source._get_streams()` as follows:

- Abstract `Decoder` class that defines decoding interface: `seek()`, `next_frame()`, `frame_size`
- Rename `Buffer -> BufferedDecoder` and `NotBuffered -> OnDemandDecoder`, both inherting from `Decoder`
- `File_Source._get_streams()` receives an additional argument: `should_buffer`
- It returns a 2-element tuple including an audio stream as before. But instead of an `av.Video_Stream`, it returns a `Decoder` instance (depending on `should_buffer`)

Consequences:
- We do not need `File_Source.buffer` anymore
- All `File_Source.buffer` api calls are called on the (new) `File_Source.video_stream` (Decoder instance)
- `File_Source.frame_size` simply returns `File_Source.video_stream.frame_size`

This way we can hide the `av.Video_Stream` completely in the `Decoder` classes.

pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -19,6 +20,9 @@
 
 logger = logging.getLogger(__name__)
 
+VIDEO_EXTS = ("mp4", "mjpeg", "h264", "mkv", "avi")
+VIDEO_TIME_EXTS = ("mp4", "mjpeg", "h264", "mkv", "avi", "time")
```
papr: Suggestion to prevent duplication: `VIDEO_TIME_EXTS = VIDEO_EXTS + ("time",)`

pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -233,12 +244,16 @@ def lookup_loc(self) -> str:
...
+        videos = (
             Video(loc)
             for ext in self.video_exts
             for loc in glob.iglob(
                 os.path.join(self.rec, f"{self.name}*.{ext}"))
         )
+        # If not self.fill_gaps, we skip the broken videos
+        if not self.fill_gaps:
+            return [v for v in videos if v.is_valid]
+        return videos
```
papr: Previously, `fetch_videos()` always returned a generator. Now it only does if `fill_gaps==True`. Therefore, I suggest to `yield from` in both cases. Additionally, I would convert `[v for v in videos if v.is_valid]` from a list comprehension to a generator as well.

pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -259,6 +274,15 @@ def build_lookup(self):
...
         container, load it if necessary, and calculate the target PTS
+
+        Case 1: all video is_valid and self._fill_gaps
+            base case
+        Case 2: all video is_valid and not self._fill_gaps
+            skip to the next video, use for detection
+        Case 3: some videos is broken and self._fill_gaps
+            return gray frame for the broken video
+        Case 4: some videos is broken and not self._fill_gaps
+            skip to the next valid video, use for detection
```
papr: Great summary 👍

pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -321,3 +341,61 @@ def _setup_lookup(self, timestamps: np.ndarray) -> np.recarray:
...
         return lookup
+
+
+class RenameSet:
+    class RenameFile:
+        def __init__(self, path):
+            self.path = path
+
+        @property
+        def is_exists(self):
```
papr: Naming suggestion: `is_exists -> exists`. This would mirror `os.path.exists` on which it is based.

pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -321,3 +341,61 @@ def _setup_lookup(self, timestamps: np.ndarray) -> np.recarray:
...
+        def rewrite_time(self, destination_name):
+            timestamps = np.fromfile(self.path, dtype=">f8")
+            timestamp_loc = os.path.join(
+                os.path.dirname(self.path), "{}_timestamps.npy".format(self.name))
+            np.save(timestamp_loc, timestamps)
+
+    def __init__(self, rec_dir, pattern, exts=VIDEO_TIME_EXTS):
+        self.rec_dir = rec_dir
+        self.pattern = os.path.join(rec_dir, pattern)
+        self.exists_files = self.get_exists_files(self.pattern, exts)
```
papr: General naming suggestion due to grammar: `exists_ -> existing_`.

pupil_src/shared_modules/video_capture/tests/utils/test_utils.py
```diff
@@ -0,0 +1,47 @@
...
+    # Rename to eye0
+    rename_set.rename("Pupil Cam(0|1) ID0", "eye0")
+    assert all(
+        elem in os.listdir(tmp_dir) for elem in
+        ["eye0.mp4", "eye0_001.mp4"])
+    assert not any(
+        elem in os.listdir(tmp_dir) for elem in
+        ["Pupil Cam1 ID0.mp4", "Pupil Cam1 ID0_001.mp4"])
+    # Rename to world
+    rename_set.rename("Pupil Cam(0|1) ID2", "world")
```
papr: Is this kind of RE compatible with `glob`?

Windsooon: No, glob use fnmatch which only support Unix shell-style wildcards ([implement details](https://github.com/python/cpython/blob/master/Lib/fnmatch.py#L74))

PR #1434, opened 2019-01-28
Gaze Producer Improvements

ckbaumann:


pupil_src/shared_modules/gaze_producer/model/calibration_storage.py
```diff
@@ -37,15 +38,24 @@ def _add_default_calibration(self):
     def create_default_calibration(self):
         return model.Calibration(
             unique_id=model.Calibration.create_new_unique_id(),
-            name="Default Calibration",
+            name=make_unique.by_number_at_end("Default Calibration", self.item_names),
             recording_uuid=self._recording_uuid,
             mapping_method="2d",
```
papr: The default mapping method should be "3d"

PR #1442, opened 2019-02-19
Fixed #1421:Flip image display of eye0 by default

### LINKED ISSUE #1421: Flip image display of eye0 by default

Windsooon:


papr:
I talked to our support team, who originally requested this feature. We decided to not merge this feature, since the eye0 videos would not be recorded in a flipped fashion but in their original orientation. Currently, customers see a preview of what is being recorded. If we flip by default, customers that care about the orientation of their eye videos would only notice after the recording.

The second case would be much more frustrating for the customer than the current state.

Issue #1419, opened 2018-12-14
Test msgpack 0.6.0 compatibility


Windsooon:
We use [check_code](https://github.com/pupil-labs/pupil/blob/master/pupil_src/shared_modules/file_methods.py#L23)

```
assert (
    msgpack.version[1] == 5
), "msgpack out of date, please upgrade to version (0, 5, 6 ) or later."
```
which doesn't work when msgpack version is 0.6.0

we can use:

```
from distutils.version import LooseVersion, StrictVersion
assert StrictVersion(msgpack.version) >= StrictVersion("0.5.6")
```
instead

papr:
I did not know that there was a new version. From a quick look, there are backwards in compatible changes that might effect us. This needs to be tested thoroughly.

papr:
There are few changes, that might have an effect on how Pupil runs. For now, we will require version 0.5.6.

Use `pip3 install msgpack==0.5.6` to install the correct version.

PR #1447, opened 2019-02-23
msgpack version assertion == --> >=

cboulay:
The error message says version 0.5.6 or greater. Presumably this means 0.6 is OK too!?


papr:
We do not know if there is 100% compatibility, see #1419

cboulay:
OK thanks. After I cleanup the Windows from-source changes then I'll make a new PR with a changed error message.

PR #1467, opened 2019-03-27
unet: Replace `torch.nn.functional` functions with their respective `torch` counterparts

papr:
pytorch 1.0 shows a deprecation warning for `torch.nn.functional.tanh()`

@ChingT please review as well.

### DIFF
--- a/pupil_src/shared_modules/calibration_routines/fingertip_calibration/models/unet.py
+++ b/pupil_src/shared_modules/calibration_routines/fingertip_calibration/models/unet.py
@@ -13,7 +13,6 @@ See COPYING and COPYING.LESSER for license details.
 
 import torch
 import torch.nn as nn
-import torch.nn.functional as F
 
 
 def conv3x3(in_channels, out_channels, stride=1, padding=1, bias=True, groups=1):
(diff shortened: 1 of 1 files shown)


pupil_src/shared_modules/calibration_routines/fingertip_calibration/models/unet.py
```diff
@@ -64,8 +63,8 @@ def __init__(self, in_channels, out_channels, pooling=True):
             self.pool = nn.MaxPool2d(kernel_size=2, stride=2)
 
     def forward(self, x):
-        x = F.relu(self.conv1(x))
```
ChingT: I think it is not necessary to replace `F.relu` with `torch.relu`.

papr: You are right that it is not necessary. I did it for code consistency reasons. Do you know if there is any difference between `F.relu` and `torch.relu`?

ChingT: I just found this: https://github.com/pytorch/pytorch/blob/81e030d9a6e8d1d5c3d83628f37f2d49bbed597d/torch/nn/functional.py#L929
So then in our case, no difference between `F.relu` and `torch.relu`.

ChingT:
thanks!

pupil_src/shared_modules/calibration_routines/fingertip_calibration/models/unet.py
papr: Great find!

PR #1468, opened 2019-03-27
OpenCV 4 compatibility changes

papr:


pupil_src/shared_modules/circle_detector.py
```diff
@@ -426,10 +426,15 @@ def find_concentric_circles(
...
+
+        # OpenCV version compatibility note:
+        # - Opencv3 returns three results: image, contours, hierarchy
+        # - Opencv4 returns two results: contours, hierarchy
+        # We do not use `image` in any case. Therefore, we can ensure
+        # compatibility by using `*_`
+        *_, contours, hierarchy = cv2.findContours(
             edge, mode=cv2.RETR_TREE, method=cv2.CHAIN_APPROX_TC89_KCOS
-        )
+        )  # CHAIN_APPROX_TC89_KCOS does not store absolutely all the contour points
```
ckbaumann: I would argue that this comment explaining `CHAIN_APPROX_TC89_KCOS` is not really necessary. "What" it does can be easily found in the opencv docs, interesting would be "why" you use this mode here (e.g. `use CHAIN_APPROX_TC89_KCOS because its faster than the default XXX`).
Anyway, feel free to keep the comment :)

papr: @ChingT Can you say something regarding the reason _why_ we are using this option?

ChingT: `use CHAIN_APPROX_TC89_KCOS because it is faster than the default method CV_CHAIN_APPROX_NONE`

papr: Thanks!

PR #1278, opened 2018-08-23
Camera auto selection

papr:
### Goal
Have a button to automatically select world and eye cameras based on selected backend.

Selected cameras:
- `eye0`: `Pupil Cam1/2 ID0`
- `eye1`: `Pupil Cam1/2 ID1`
- `world`: `Pupil Cam1 ID2`

Note: The NDSI backend needs to respect the selected host.

PR #1470, opened 2019-03-28
1278 camera auto selection

emredog:
Implements feature, as requested in #1278

### DIFF
--- a/pupil_src/shared_modules/video_capture/base_backend.py
+++ b/pupil_src/shared_modules/video_capture/base_backend.py
@@ -175,34 +175,29 @@ class Base_Manager(Plugin):
     def __init__(self, g_pool):
         super().__init__(g_pool)
 
+        from . import manager_classes
+
+        self.manager_classes = {m.__name__: m for m in manager_classes}
+
     def on_notify(self, notification):
         """
         Reacts to notification:
-            ``notify.auto_select_manager``: Changes the current Manager to one that's emitted
-            ``notify.auto_activate``: Activates the current source via self.auto_activate()
+            ``backend.auto_select_manager``: Changes the current Manager to one that's emitted
+            ``backend.auto_activate_source``: Activates the current source via self.auto_activate_source()
 
         Emmits notifications (indirectly):
             ``start_plugin``: For world thread
-            ``notify.auto_activate``
+            ``backend.auto_activate_source``
         """
 
-        def get_class_by_name(name):
-            from . import manager_classes
-
-            manager_classes_lut = {}
-            for m in manager_classes:
-                manager_classes_lut[m.__name__] = m
-
-            return manager_classes_lut[name]
-
-        if notification["subject"].startswith("notify.auto_select_manager"):
-            target_manager_class = get_class_by_name(notification["name"])
+        if notification["subject"].startswith("backend.auto_select_manager"):
+            target_manager_class = self.manager_classes[notification["name"]]
             self.replace_backend_manager(target_manager_class, auto_activate=True)
         if (
-            notification["subject"].startswith("notify.auto_activate")
+            notification["subject"].startswith("backend.auto_activate_source")
             and notification["name"] == self.g_pool.process
         ):
-            self.auto_activate()
+            self.auto_activate_source()
 
     def replace_backend_manager(self, manager_class, auto_activate=False):
         if self.g_pool.process.startswith("eye"):
@@ -218,10 +213,10 @@ class Base_Manager(Plugin):
                 )
         if auto_activate:
             self.notify_all(
-                {"subject": "notify.auto_activate", "name": self.g_pool.process}
+                {"subject": "backend.auto_activate_source", "name": self.g_pool.process}
             )
 
-    def auto_activate(self):
+    def auto_activate_source(self):
         """This function should be implemented in *_Manager classes 
             to activate the corresponding source with following preferences:
                 eye0: Pupil Cam1/2 ID0
@@ -234,7 +229,14 @@ class Base_Manager(Plugin):
 
     def auto_select_manager(self):
         self.notify_all(
-            {"subject": "notify.auto_select_manager", "name": self.__class__.__name__}
+            {"subject": "backend.auto_select_manager", "name": self.class_name}
+        )
+
+    def add_auto_select_button(self):
+        from pyglui import ui
+
+        self.menu.append(
+            ui.Button("Auto select in all processes", self.auto_select_manager)
         )
 
     def add_menu(self):
--- a/pupil_src/shared_modules/video_capture/base_backend.py
+++ b/pupil_src/shared_modules/video_capture/base_backend.py
@@ -239,7 +239,7 @@ class Base_Manager(Plugin):
         from pyglui import ui
 
         self.menu.append(
-            ui.Button("Auto select in all processes", self.auto_select_manager)
+            ui.Button("Start with default devices", self.auto_select_manager)
         )
 
     def add_menu(self):
--- a/pupil_src/shared_modules/video_capture/base_backend.py
+++ b/pupil_src/shared_modules/video_capture/base_backend.py
@@ -195,7 +195,7 @@ class Base_Manager(Plugin):
             self.replace_backend_manager(target_manager_class, auto_activate=True)
         if (
             notification["subject"].startswith("backend.auto_activate_source")
-            and notification["name"] == self.g_pool.process
+            and notification["proc_name"] == self.g_pool.process
         ):
             self.auto_activate_source()
 
@@ -213,7 +213,10 @@ class Base_Manager(Plugin):
                 )
         if auto_activate:
             self.notify_all(
-                {"subject": "backend.auto_activate_source", "name": self.g_pool.process}
+                {
+                    "subject": "backend.auto_activate_source",
+                    "proc_name": self.g_pool.process,
+                }
             )
 
     def auto_activate_source(self):
--- a/pupil_src/shared_modules/video_capture/base_backend.py
+++ b/pupil_src/shared_modules/video_capture/base_backend.py
@@ -176,35 +176,25 @@ class Base_Manager(Plugin):
         super().__init__(g_pool)
 
     def on_notify(self, notification):
+        def get_class_by_name(name):
+            from . import manager_classes
+
+            manager_classes_lut = {}
+            for m in manager_classes:
+                manager_classes_lut[m.__name__] = m
+
+            return manager_classes_lut[name]
+
         if notification["subject"].startswith("notify.auto_select_manager"):
-            if notification["name"] == "Fake_Manager":
-                from .fake_backend import Fake_Manager
+            target_manager_class = get_class_by_name(notification["name"])
+            self.replace_backend_manager(target_manager_class, auto_activate=True)
+        if (
+            notification["subject"].startswith("notify.auto_activate")
+            and notification["name"] == self.g_pool.process
+        ):
+            self.auto_activate()
 
-                target_manager_class = Fake_Manager
-            elif notification["name"] == "File_Manager":
-                from .file_backend import File_Manager
-
-                target_manager_class = File_Manager
-            elif notification["name"] == "NDSI_Manager":
-                pass
-            elif notification["name"] == "Realsense_Manager":
-                pass
-            elif notification["name"] == "UVC_Manager":
-                from .uvc_backend import UVC_Manager
-
-                target_manager_class = UVC_Manager
-            elif notification["name"] == "Realsense2_Manager":
-                pass
-            else:
-                raise NotImplementedError(
-                    "Manager auto-selection is not implemented for {}".format(
-                        notification["name"]
-                    )
-                )
-
-            self.replace_backend_manager(target_manager_class)
-
-    def replace_backend_manager(self, manager_class):
+    def replace_backend_manager(self, manager_class, auto_activate=False):
         if self.g_pool.process.startswith("eye"):
             self.g_pool.capture_manager.deinit_ui()
             self.g_pool.capture_manager.cleanup()
@@ -212,6 +202,26 @@ class Base_Manager(Plugin):
             self.g_pool.capture_manager.init_ui()
         else:
             self.notify_all({"subject": "start_plugin", "name": manager_class.__name__})
+        if auto_activate:
+            self.notify_all(
+                {"subject": "notify.auto_activate", "name": self.g_pool.process}
+            )
+
+    def auto_activate(self):
+        """This function should be implemented in *_Manager classes 
+            to activate the corresponding source with following preferences:
+                eye0: Pupil Cam1/2 ID0
+                eye1: Pupil Cam1/2 ID1
+                world: Pupil Cam1 ID2
+
+            See issue #1278 for more details.
+        """
+        pass
+
+    def auto_select_manager(self):
+        self.notify_all(
+            {"subject": "notify.auto_select_manager", "name": self.__class__.__name__}
+        )
 
     def add_menu(self):
         super().add_menu()
@@ -219,14 +229,6 @@ class Base_Manager(Plugin):
         from pyglui import ui
 
         self.menu_icon.order = 0.1
-
-        def auto_select_manager():
-            self.notify_all(
-                {
-                    "subject": "notify.auto_select_manager",
-                    "name": self.__class__.__name__,
-                }
-            )
 
         # We add the capture selection menu
         manager_classes.sort(key=lambda x: x.gui_name)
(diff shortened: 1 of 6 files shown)


papr:
@emredog
- Backend: Pupil Mobile
- Feature request: Log a warning if the selected host does not have any cameras available
- Use-case: Pupil Mobile running without a connected headset

pupil_src/shared_modules/video_capture/ndsi_backend.py
```diff
@@ -408,76 +433,83 @@ def re_build_ndsi_menu(self):
...
                 label="Remote host",
             )
         )
 
         self.menu.extend(ui_elements)
+
+        self.menu.append(
+            ui.Button("Auto select in all processes", self.auto_select_manager)
+        )
+
```
papr: Can this be abstracted into a `base_backend` function? This way would not have to change the button text n multiple locations if needed to.

pupil_src/shared_modules/video_capture/ndsi_backend.py
```diff
@@ -408,76 +433,83 @@ def re_build_ndsi_menu(self):
...
+        self.notify_all(
+            {
+                "subject": "notify.selected_host",
+                "selected_host": self.selected_host,
+                "delay": 0.4,
+            }
+        )
+
+    def auto_activate(self):
+        cam_selection_lut = {"eye0": "ID0", "eye1": "ID1", "world": "ID2"}
```
papr: This could be a class attribute for efficiency.

pupil_src/shared_modules/video_capture/ndsi_backend.py
```diff
@@ -536,13 +571,36 @@ def on_notify(self, n):
...
+            ``notify.selected_host``: Switches to selected host from other process
 
         Emmits notifications:
             ``capture_manager.source_found``
         """
+
+        super().on_notify(n)
+
         if (
             n["subject"].startswith("capture_manager.source_found")
```
papr: This notification could use a better name, e.g. `backend.ndsi_source_found`

pupil_src/shared_modules/video_capture/ndsi_backend.py
```diff
@@ -536,13 +571,36 @@ def on_notify(self, n):
...
             and isinstance(self.g_pool.capture, NDSI_Source)
             and not self.g_pool.capture.sensor
         ):
             self.recover()
+
+        if n["subject"].startswith("notify.selected_host"):
```
papr:
```suggestion
        if n["subject"].startswith("backend.ndsi_host_selected"):
```

pupil_src/shared_modules/video_capture/ndsi_backend.py
```diff
@@ -536,13 +571,36 @@ def on_notify(self, n):
...
+        if selected_host in host_sel:
+            self.view_host(selected_host)
+            self.should_select_host = None
+            self.notify_all(
+                {
+                    "subject": "notify.auto_select_manager",
+                    "name": self.__class__.__name__,
+                }
+            )
+            # self.auto_activate()
```
papr: Please remove the comment if it is not needed anymore.

pupil_src/shared_modules/video_capture/ndsi_backend.py
emredog: My train of thought was "this function is to be seldomly called, so it's better not the bloating the memory up".

But you're right, we should favor performance over memory.

pupil_src/shared_modules/video_capture/ndsi_backend.py
```diff
@@ -408,76 +434,82 @@ def re_build_ndsi_menu(self):
...
+        logger.info(src_sel_labels)
+        if src_sel is None or src_sel_labels is None:
+            logger.warning("here?")
+            return
+
+        source_id = [
+            src_sel[i] for i, lab in enumerate(src_sel_labels) if cam_id in lab
+        ]
+
+        if len(source_id) != 1:
```
ckbaumann: Could it happen that there are more than one s.t. this should rather be `if len(source_id) == 0`?

pupil_src/shared_modules/video_capture/ndsi_backend.py
```diff
@@ -408,76 +434,82 @@ def re_build_ndsi_menu(self):
...
+            return
+
+        source_id = source_id[0]
```
ckbaumann: I'm not a particular fan of changing the type of a variable in a function (replacing a list with one of its elements in this case). For longer methods, it makes it difficult to judge what a variable means. If I see just `source_id` I always need to check if it is before or after the change, to know its meaning. I would prefer something like `first_source_id = source_id[0]` instead. However, in this case I don't see any reason against `self.activate(source_id[0])` which is super clear.

pupil_src/shared_modules/video_capture/uvc_backend.py
```diff
@@ -773,42 +774,58 @@ def dev_selection_list():
...
+    def auto_activate_source(self):
+        if not self.devices or len(self.devices) == 0:
+            logger.warning("No UVC source available.")
+            return
+
+        cam_selection_lut = {"eye0": "ID0", "eye1": "ID1", "world": "ID2"}
+        cam_id = cam_selection_lut[self.g_pool.process]
+        source_id = [d["uid"] for d in self.devices if cam_id in d["name"]]
+        if len(source_id) != 1:
+            logger.warning("Unexpected process name / camera ID pair")
```
ckbaumann: Maybe include the `cam_id` you are looking for in the warning?

pupil_src/shared_modules/video_capture/ndsi_backend.py
ckbaumann: Thinking more about this, why not do something like
```
try:
    source_id = next(src_sel[i] for i, lab in enumerate(src_sel_labels) if cam_id in lab)
except StopIteration:
    source_id = None
return source_id
```

pupil_src/shared_modules/video_capture/base_backend.py
```diff
@@ -175,30 +175,83 @@ class Base_Manager(Plugin):
...
+
+    def replace_backend_manager(self, manager_class, auto_activate=False):
+        if self.g_pool.process.startswith("eye"):
+            if not isinstance(self.g_pool.capture_manager, manager_class):
                 self.g_pool.capture_manager.deinit_ui()
                 self.g_pool.capture_manager.cleanup()
                 self.g_pool.capture_manager = manager_class(self.g_pool)
                 self.g_pool.capture_manager.init_ui()
-            else:
+        else:
```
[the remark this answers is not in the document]
emredog: I find it more readable this way, here's my reasoning:
It's a two-fold (hierarchical) condition. First one is to decide whether it's an `eye` or `world` process, and *then* it checks if there's need for action for that process type.

**Or**, we could just formulate it this way:
```
        if (self.g_pool.process.startswith("eye") and 
            not isinstance(self.g_pool.capture_manager, manager_class)):
            self.g_pool.capture_manager.deinit_ui()
            ...
        elif not isinstance(self, manager_class):
                self.notify_all(
                    {"subject": "start_plugin", "name": manager_class.__name__}
                )
```

What do  you think?

pupil_src/shared_modules/video_capture/base_backend.py
```diff
@@ -175,30 +175,83 @@ class Base_Manager(Plugin):
...
+            if not isinstance(self, manager_class):
                 self.notify_all(
                     {"subject": "start_plugin", "name": manager_class.__name__}
                 )
+        if auto_activate:
+            self.notify_all(
+                {"subject": "backend.auto_activate_source", "name": self.g_pool.process}
+            )
+
+    def auto_activate_source(self):
```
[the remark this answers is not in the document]
emredog: There are backends (e.g. `realsense` and `realsense2`) that are not supposed to implement this feature.

Do you think it would be better to have a 
```python
def auto_activate_source(self):
    pass
````
in those classes?

File: pupil_src/shared_modules/video_capture/ndsi_backend.py [hunk 11]
```diff
@@ -408,76 +434,82 @@ def re_build_ndsi_menu(self):
...
+        )
+
+    def auto_activate_source(self):
+        cam_id = self.cam_selection_lut[self.g_pool.process]
+
+        src_sel, src_sel_labels = self.source_selection_list()
+        logger.info(src_sel)
+        logger.info(src_sel_labels)
+        if src_sel is None or src_sel_labels is None:
+            logger.warning("here?")
```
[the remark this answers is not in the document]
emredog: Yeah sorry, this was just a temporary thing, it will be changed. Good catch! :+1:

pupil_src/shared_modules/video_capture/ndsi_backend.py
emredog: Not sure. Would any single host would have more than one headset connected?

We can have 
```python
        if len(source_id) == 0:
            return

        source_id = source_id[0]
```
i.e. choosing the first camera that matches the criteria.

pupil_src/shared_modules/video_capture/ndsi_backend.py
emredog: This makes perfect sense, and it's definitely more compact. (except that I change `return source_id` with `self.activate(source_id)`)

Thanks for the suggestion!

pupil_src/shared_modules/video_capture/uvc_backend.py
emredog: Better yet, I incorporated what you suggested for the `ndsi_backend` for both classes:

```python
        try:
            source_id = next(
                uid for uid, name in self.devices.items() if cam_id in name
            )
        except StopIteration:
            logger.warning("No camera found with ID: {}".format(cam_id))
            source_id = None
        self.activate(source_id)
```

pupil_src/shared_modules/video_capture/ndsi_backend.py
emredog: Just ignore this. See comments below. I'm resolving this.

pupil_src/shared_modules/video_capture/base_backend.py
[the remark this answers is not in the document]
ckbaumann: Yeah, Pablo already pointed this out. Then I think its ok to keep it the way it is. Maybe you can add some info in the docstring like "Implement this if your data source ....."

pupil_src/shared_modules/video_capture/base_backend.py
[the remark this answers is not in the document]
ckbaumann: You are right, keep the original version, you've got a proper reason, in my opinion. The `and` would make things hard to read.

pupil_src/shared_modules/video_capture/ndsi_backend.py
```diff
@@ -492,23 +492,23 @@ def auto_select_manager(self):
...
-        cam_id = self.cam_selection_lut[self.g_pool.process]
+        if not self.selected_host:
+            return
 
         src_sel, src_sel_labels = self.source_selection_list()
-        logger.info(src_sel)
-        logger.info(src_sel_labels)
-        if src_sel is None or src_sel_labels is None:
-            logger.warning("here?")
+        if len(src_sel) <= 1 and src_sel[0] is None:
```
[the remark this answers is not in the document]
ckbaumann: ok, the list will never be empty, but anyway the changed version raises a little less suspicion.

pupil_src/shared_modules/video_capture/ndsi_backend.py
```diff
@@ -486,29 +489,34 @@ def auto_select_manager(self):
...
-        source_id = source_id[0]
-        self.activate(source_id)
+        for cam_id in cam_ids:
+            try:
+                source_id = next(
+                    src_sel[i] for i, lab in enumerate(src_sel_labels) if cam_id in lab
+                )
+            except StopIteration:
+                source_id = None
+                logger.warning("No camera was found with ID: {}".format(cam_id))
```
papr: In accordance with the button label change, I would suggest to rename the warning to `No default device found for {process name}`.

Additionally, I would move the error handling/warning out of the exception into the [`else`-part of the for-loop](http://book.pythontips.com/en/latest/for_-_else.html).

pupil_src/shared_modules/video_capture/uvc_backend.py
```diff
@@ -790,9 +793,15 @@ def dev_selection_list():
...
         if not source_uid:
             return
-        if not uvc.is_accessible(source_uid):
-            logger.error("The selected camera is already in use or blocked.")
+
+        try:
+            if not uvc.is_accessible(source_uid):
+                logger.error("The selected camera is already in use or blocked.")
+                return
+        except ValueError as ve:
```
papr: When and by what is this `ValueError` triggered?

pupil_src/shared_modules/video_capture/uvc_backend.py
```diff
@@ -812,16 +821,22 @@ def activate(self, source_uid):
...
-        source_id = source_id[0]
+        cam_ids = self.cam_selection_lut[self.g_pool.process]
 
-        self.activate(source_id)
+        for cam_id in cam_ids:
+            try:
+                source_id = next(d["uid"] for d in self.devices if cam_id in d["name"])
+            except StopIteration:
+                logger.warning("No camera found with ID: {}".format(cam_id))
+                source_id = None
```
papr: As above, I would move this code into the `else`-part of the `for`-loop.

pupil_src/shared_modules/video_capture/uvc_backend.py
emredog: For the following scenario:
1. switch from Test Image to UVC backend
2. hit Auto-select
3. (all cameras active now)
4. unplug headset, hit the button again shortly after


Here, `UVCManager.auto_activate_source` calls the `self.activate(source_id)` with a valid argument. But when in `UVCManager.activate`, the `source_uid` is not `None`, but also invalid somehow. 

This raises a `ValueError` in `if not uvc.is_accessible(source_uid)`.

pupil_src/shared_modules/video_capture/ndsi_backend.py
emredog: Good suggestions. Thanks!

pupil_src/shared_modules/video_capture/ndsi_backend.py
```diff
@@ -501,22 +501,20 @@ def auto_activate_source(self):
...
-                )
-            except StopIteration:
-                source_id = None
-                logger.warning("No camera was found with ID: {}".format(cam_id))
+            for i, label in enumerate(src_sel_labels):
+                if cam_id in label:
+                    source_id = src_sel[i]
+                    self.activate(source_id)
+                    break
             else:
```
papr: Sorry for not being clear. :-/ I think the `next()` call is still the best way to find a match. What I was referring to was to move the log message to the outer `for cam_id in cam_ids` loop. This way the user does not get a warning for each condition that was not met but only one in case no condition was met at all.

pupil_src/shared_modules/video_capture/uvc_backend.py
```diff
@@ -823,20 +823,20 @@ def activate(self, source_uid):
...
+            for d in self.devices:
+                if cam_id in d["name"]:
+                    source_id = d["uid"]
+                    self.activate(source_id)
+                    break
             else:
-                self.activate(source_id)
-                break
+                logger.warning("The default device is not found.")
+                source_id = None
```
papr: Same as above applies here.

pupil_src/shared_modules/video_capture/uvc_backend.py
emredog: I hope I get it right this time. Both `StopIteration` and `for/else` are fairly new to me.

pupil_src/shared_modules/video_capture/ndsi_backend.py
emredog: I hope I get it right this time. Both StopIteration and for/else are fairly new to me.

pupil_src/shared_modules/video_capture/uvc_backend.py
papr: Ah, thank you. I guess `uvc.is_accessible()` expects a value that is part of `uvc.Device_List()`

pupil_src/shared_modules/video_capture/ndsi_backend.py
papr: Looks good! 

- A `StopIteration` is raised if a generator `return`s. See this tutorial on  https://realpython.com/introduction-to-python-generators/
- Related to `for-else`: https://realpython.com/quizzes/python-while-loop This quiz starts simple but is actually a bit tricky ;)

pupil_src/shared_modules/video_capture/ndsi_backend.py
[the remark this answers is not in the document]
papr: I actually think that you raise a good point, @ckbaumann 
```suggestion
        if len(src_sel) < 1 or src_sel[0] is None:
```

PR #1476, opened 2019-04-04
Allow any audio*.mp4 to be loaded for audio playback

### LINKED ISSUE #1475: Player v1.11: Pupil Mobile audio files are not renamed to `audio.mp4`
Two possible solutions:

1) Fix the renaming and add recording upgrade method that renames non-renamed audio files retrospectively.
2) Change the audio playback plugin's file name constraint from `audio.mp4` to `audio*.mp4`

Solution 2) is simpler to implement and comes with the advantage that we do not have to change any files retrospectively.

papr:
Fixes #1475 by allowing any `audio*.mp4` to be loaded for audio playback.

- Uses `glob` to match potential audio files
- Prioritizes `audio.mp4` over `audio_xxxxx.mp4`
- Extracts loading of audio container, stream and timestamps into separate function and returns them as `LoadedAudio` namedtuple
- `Audio_Playback` and `Audio_Viz_Transform` make direct use of `LoadedAudio`

### DIFF
--- a/pupil_src/shared_modules/av_writer.py
+++ b/pupil_src/shared_modules/av_writer.py
@@ -15,23 +15,22 @@ requires:
     -
 """
 
-import os, sys, platform
+# logging
+import logging
+import multiprocessing as mp
+import os
+import platform
+import sys
+from fractions import Fraction
+from threading import Event, Thread
+from time import time
+
+import numpy as np
+
 import av
 from av.packet import Packet
 
-import numpy as np
-from time import time
-from fractions import Fraction
-
-# logging
-import logging
-
 logger = logging.getLogger(__name__)
-
-
-from threading import Thread
-from threading import Event
-import multiprocessing as mp
 
 
 """
--- a/pupil_src/shared_modules/av_writer.py
+++ b/pupil_src/shared_modules/av_writer.py
@@ -147,23 +147,32 @@ class AV_Writer(object):
 
     def write_video_frame(self, input_frame):
         if not self.configured:
-            self.video_stream.height = input_frame.height
-            self.video_stream.width = input_frame.width
-            self.configured = True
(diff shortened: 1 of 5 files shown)


pupil_src/shared_modules/av_writer.py
```diff
@@ -190,35 +184,39 @@ def write_video_frame(self, input_frame):
...
+                    / self.audio_export_stream.time_base
                 )
                 audio_packet.pts = audio_pts
                 audio_packet.dts = audio_pts
-                audio_packet.stream = self.audio_export
+                audio_packet.stream = self.audio_export_stream
                 self.audio_packets_decoded += 1
 
-                if audio_pts * self.audio_export.time_base < 0:
+                if audio_pts * self.audio_export_stream.time_base < 0:
```
ckbaumann: I don't get what `audio_pts * self.audio_export_stream.time_base` is and you have this computation three times. Could you think of an explanatory variable name where you could store the value?

pupil_src/shared_modules/av_writer.py
```diff
@@ -190,35 +184,39 @@ def write_video_frame(self, input_frame):
             self.container.mux(packet)
         self.current_frame_idx += 1
         self.timestamps.append(input_frame.timestamp)
-        if self.audio_export:
-            for audio_packet in self.audio_rec.demux():
-                if self.audio_packets_decoded >= len(self.audio_ts):
+        if self.audio:
```
ckbaumann: The content of the `if` would like great as an extracted method ;) Ok, in this case it still looks ok because it's the last block in the method, but in general such high numbers of indentation can be hard to follow.

papr: Good point. I extracted the other functionality while I was at it.

PR #1481, opened 2019-04-09
Allow plugins to consume events 

papr:
### Consumed events
Consumed events will not be propagated to other plugins.

Consumable events:
- `on_click`
- `on_key`
- `on_char`
- `on_drop`

### `on_drop` support in Player
Add `on_drop` plugin callback support in Player. This allows plugins to react to files being dropped onto the Player window.

When dropping a valid recording onto player the previous behaviour (opening the dropped recording) will take preference over executing the `on_drop` plugin callbacks.


papr:
There are plugins that do use the callback but do not return anything aka. `None`. This results in equivalent as the previous version. You are right though, that I have to check in which cases it makes sense to consume the event. It is not always necessary to do that.

pupil_src/shared_modules/vis_eye_video_overlay.py
```diff
@@ -339,12 +339,16 @@ def on_click(self, pos, button, action):
             if not self.eye1.on_click(
```
ckbaumann: if you wanted to, you could change the `action == 1` above to `action == glfw.GLFW_PRESS`

papr: Adapted that in the new version

PR #1482, opened 2019-04-11
Eye movement detector

romanroibu:
This PR adds eye movement detector based on the NSLR-HMM library.

Paper: https://www.nature.com/articles/s41598-017-17983-x
Library: https://gitlab.com/nslr/nslr-hmm


pupil_src/shared_modules/eye_movement_detector.py
```diff
@@ -0,0 +1,937 @@
...
+    SMOOTH_PURSUIT = "smooth_pursuit"
+
+    @classmethod
+    def from_nslr_class(cls, nslr_class):
+        return {
+            nslr_hmm.FIXATION: Segment_Class.FIXATION,
+            nslr_hmm.SACCADE: Segment_Class.SACCADE,
+            nslr_hmm.PSO: Segment_Class.POST_SACCADIC_OSCILLATIONS,
+            nslr_hmm.SMOOTH_PURSUIT: Segment_Class.SMOOTH_PURSUIT,
+        }[nslr_class]
```
papr: How about making this a class attribute. This would change the calling syntax only slightly different and the dict would be instantiated only once.

pupil_src/shared_modules/eye_movement_detector.py
```diff
@@ -0,0 +1,937 @@
...
+        yellow = (241 / 255, 196 / 255, 15 / 255, 1.0)  # sun flower
+        green = (39 / 255, 174 / 255, 96 / 255, 1.0)  # nephritis
+        blue = (41 / 255, 128 / 255, 185 / 255, 1.0)  # belize hole
+        purple = ((142 / 255, 68 / 255, 173 / 255, 1.0),)  # wisteria
+        return {
+            Segment_Class.FIXATION: yellow,
+            Segment_Class.SACCADE: green,
+            Segment_Class.POST_SACCADIC_OSCILLATIONS: blue,
+            Segment_Class.SMOOTH_PURSUIT: purple,
+        }.get(self, grey)
```
papr: Same argument as above to make this dict a class attribute. Also, I would not catch the case where `self` is not one of the possible values.

pupil_src/shared_modules/eye_movement_detector.py
```diff
@@ -0,0 +1,937 @@
...
+
+
+Gaze_Time = typing.Iterable[float]
+
+
+MsgPack_Serialized_Segment = typing.Type[bytes]
+
+
+class Classified_Segment:
+    def from_attrs(
```
papr: Is this supposed to be a `@staticmethod`?

pupil_src/shared_modules/eye_movement_detector.py
```diff
@@ -0,0 +1,937 @@
...
+GFR = typing.TypeVar("GFR")  # Generator function return type
+
+On_Started_Observer = typing.Callable[[], None]
+On_Yield_Observer = typing.Callable[[GFY], None]
+On_Completed_Observer = typing.Callable[[GFR], None]
+On_Ended = typing.Callable[[], None]
+On_Exception = typing.Callable[[Exception], None]
+On_Canceled_Or_Killed = typing.Callable[[], None]
+
+Eye_Movement_Detection_Task_Generator = typing.Generator[GFY, GFS, GFR]
```
papr: How about
> Alternatively, annotate your generator as having a return type of either Iterable[YieldType] or Iterator[YieldType]:

https://docs.python.org/3/library/typing.html#typing.Generator

PR #1492, opened 2019-04-27
Even running from source, do not build modules unless required.

cboulay:
Currently, when running from source, the built modules will be rebuilt (no-op) every time. This is problematic for a couple reasons. First, it means that the RUN environment has to have all of the includes and libs required to build the modules, even though they aren't all required to run the modules. Second, PyCharm debug mangles in-line Cython commands; see the bug report [here](https://youtrack.jetbrains.com/issue/PY-27833).

There is a work around available to the PyCharm-specific issue, but this disables automatic attaching to sub-processes, and this is a desirable feature given how pupil makes use of multiprocessing.

So this PR simply tries to import the compiled modules then if it fails, and if it's not the bundle, it then tries to build the required modules.

### DIFF
--- a/pupil_src/shared_modules/pupil_detectors/__init__.py
+++ b/pupil_src/shared_modules/pupil_detectors/__init__.py
@@ -9,18 +9,15 @@ See COPYING and COPYING.LESSER for license details.
 ---------------------------------------------------------------------------~(*)
 """
 
-# when running from source compile cpp extension if necessary.
-import sys
-
-if not getattr(sys, "frozen", False):
+try:
+    from .detector_2d import Detector_2D
+except ModuleNotFoundError:
+    # when running from source compile cpp extension if necessary.
     from .build import build_cpp_extension
-
     build_cpp_extension()
+    from .detector_2d import Detector_2D
 
-from .detector_2d import Detector_2D
 from .detector_3d import Detector_3D
 from .detector_dummy import Detector_Dummy
-
-
 # explicit import here for pyinstaller because it will not search .pyx source files.
 from .visualizer_3d import Eye_Visualizer
--- a/pupil_src/shared_modules/calibration_routines/optimization_calibration/__init__.py
+++ b/pupil_src/shared_modules/calibration_routines/optimization_calibration/__init__.py
@@ -9,13 +9,10 @@ See COPYING and COPYING.LESSER for license details.
 ---------------------------------------------------------------------------~(*)
 """
 
-# when running from source compile cpp extension if nessesary.
-
-import sys
-
-if not getattr(sys, "frozen", False):
+try:
+    from .calibration_methods import bundle_adjust_calibration
+except ModuleNotFoundError:
+    # when running from source compile cpp extension if necessary.
     from .build import build_cpp_extension
-
     build_cpp_extension()
-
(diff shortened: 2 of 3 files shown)


papr:
Where does the PyCharm-specific issue come into play? We do not use any Cython in-line code as far as I know.

pupil_src/shared_modules/pupil_detectors/__init__.py
```diff
@@ -9,18 +9,17 @@
...
 """
 
-# when running from source compile cpp extension if necessary.
-import sys
+try:
+    from .detector_2d import Detector_2D
+except ModuleNotFoundError:
+    # when running from source compile cpp extension if necessary.
+    import sys
+    if not getattr(sys, "frozen", False):
```
papr: This check can be removed since the build detector should always be present in the bundle and therefore not fail. Same applies to the `optimization_calibration/` changes.

cboulay:
> Where does the PyCharm-specific issue come into play? We do not use any Cython in-line code as far as I know.

To be honest I don't know the internals of PyCharm's debugger, but this line:

`ret = sp.check_output(build_cmd).decode(sys.stdout.encoding)`

spawns a sys process with the command `python setup.py install --install-lib=../`. That generates more Cython commands to build the module, and it's those generated commands that are not parsed correctly when run in debug mode.

pupil_src/shared_modules/pupil_detectors/__init__.py
cboulay: Updated and rebased.

brettbrocato:
I'm not a pupil user, but seeking somewhere to ask a question since it looks like PyCharm is not open source.  I'm having an issue that looks like this, where my mulitprocessing code runs in Jupyter notebook on WinPython 3.6, but not in Pycharm with the same interpreter, where it locks up 'forever'.  Can you provide any insight?  Thanks!

https://stackoverflow.com/questions/47267077/multiprocessing-process-locking-up-forever-in-pycharm

https://stackoverflow.com/questions/51335467/python-multiprocessing-cant-create-process-on-windows-in-pycharm

papr:
@cboulay Maybe a bit off-topic to the repo but did you try VS Code debugging?
https://gist.github.com/papr/cee69604af30e97ce5104a9273ca421d

cboulay:
> Maybe a bit off-topic to the repo but did you try VS Code debugging?

No, PyCharm only. Am I missing out on something?

papr:
> No, PyCharm only. Am I missing out on something?
Depends on who you ask 🙈 The experiences are mixed.

papr:
I just realized that this pr comes with the disadvantage that the cython code will not automatically recompile if it was changed... 🤔

cboulay:
@papr I guess that is a disadvantage but I still think it's wrong to always try to build.

How about a sys.arg like `--build` ([added here](https://github.com/pupil-labs/pupil/blob/master/pupil_src/main.py#L21-L22)) when set will call `build_cpp_extension()` for each of the cython modules? Cython-level developers can set this argument in their shortcut/ide/whatever.

papr:
It's not Cython developers that I am worried about but normal users that run from source.  It might result in hidden problems if changes were made but not applied. Alternatively, we could explicitly check the version and require a rebuild if it changed.

PR #1516, opened 2019-05-30
 Export eye movements by gaze and by segment 

papr:
Instead of having to map segments to surfaces, we can just rely on mapping gaze data to surfaces and then correlating them to the `eye_movement_by_gaze.csv` output.


papr:
@romanroibu please test the real-time by-gaze classification. For some reason accessing the `gaze_classification` result from `nslr_hmm.classify_gaze()` sometimes hangs/crashes on macOS.

PR #1518, opened 2019-06-06
Pupil Invisible Gaze Preview

papr:
Plugin that subscribes to the gaze data published by the Pupil Invisible Companion app.

---

Requires: https://github.com/pupil-labs/pi_preview


pupil_src/shared_modules/pi_preview/sensor.py
```diff
@@ -0,0 +1,118 @@
...
+        self._ndsi_sensor = None
+        self.network = network
+        self.host_uuid = linked_device.uuid
+        self.host_name = linked_device.name
+        self.offset = [0, 0]
+        if self.linked:
+            self.activate()
+
+    @property
+    def linked(self):
```
romanroibu: nitpick: I find prefixing `bool` properties with `is_XYZ` much more readable, especially since in some contexts it even makes sense to have both `XYZ()` and `is_XYZ` (e.g. `my_array.sorted()` returns a new, sorted copy of `my_array`, while `my_array.is_sorted` returns `True` if `my_array` is already sorted).

PR #1529, opened 2019-06-19
Add gaze data input validation for NSLR-HMM classification

romanroibu:
Calling `nslr_hmm.classify_gaze` function with invalid eye positions and/or timestamps causes the function to hang in an invalid state. This PR prevents calling this function with invalid data, and shows the user a message with the cause of the validation error.

This addresses the same problem as #1528 but specifically targets the manifestation of the problem.

### DIFF
--- a/pupil_src/shared_modules/eye_movement/eye_movement_detector_offline.py
+++ b/pupil_src/shared_modules/eye_movement/eye_movement_detector_offline.py
@@ -106,7 +106,7 @@ class Offline_Eye_Movement_Detector(Observable, Eye_Movement_Detector_Base):
 
     def on_task_exception(self, exception: Exception):
         error_message = f"{exception}"
-        logger.exception(error_message)
+        logger.error(error_message)
         self.menu_content.update_error_text(error_message)
 
     def on_task_completed(self):
--- a/pupil_src/shared_modules/eye_movement/eye_movement_detector_offline.py
+++ b/pupil_src/shared_modules/eye_movement/eye_movement_detector_offline.py
@@ -58,6 +58,7 @@ class Offline_Eye_Movement_Detector(Observable, Eye_Movement_Detector_Base):
         self.offline_controller = controller.Eye_Movement_Offline_Controller(
             plugin=self,
             storage=self.storage,
+            on_started=self.on_task_started,
             on_status=self.on_task_status,
             on_progress=self.on_task_progress,
             on_exception=self.on_task_exception,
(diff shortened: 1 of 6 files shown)


pupil_src/shared_modules/eye_movement/utils.py
```diff
@@ -67,4 +67,30 @@ def gaze_data_to_nslr_data(capture, gaze_data, use_pupil: bool):
...
+    def is_monotonic(arr: np.ndarray):
+        return np.all(arr[:-1] <= arr[1:])
+
+    def is_unique(arr: np.ndarray):
+        return arr.shape == np.unique(arr, axis=0).shape
+
+    if has_nan(eye_positions):
+        raise ValueError("Gaze data contains NaN values")
+    if not is_monotonic(eye_timestamps):
+        raise ValueError("Gaze timestamps contains NaN values")
```
papr:
```suggestion
        raise ValueError("Gaze timestamps contain NaN values")
```

pupil_src/shared_modules/eye_movement/utils.py
```diff
@@ -67,4 +67,30 @@ def gaze_data_to_nslr_data(capture, gaze_data, use_pupil: bool):
...
+    if not is_monotonic(eye_timestamps):
+        raise ValueError("Gaze timestamps are not monotonic")
+    if not is_unique(eye_timestamps):
+        raise ValueError("Gaze timestamps are not unique. Please recalculate gaze mapping with only 1 mapper enabled")
```
papr:
```suggestion
        raise ValueError("Gaze timestamps are not unique. If you are using Offline Calibration, please disable all but one gaze mapping section.")
```

pupil_src/shared_modules/eye_movement/eye_movement_detector_offline.py
```diff
@@ -91,14 +95,19 @@ def trigger_recalculate(self):
...
     def on_task_progress(self, progress: float):
         self.menu_content.update_progress(progress)
 
     def on_task_status(self, status: str):
         self.menu_content.update_status(status)
 
     def on_task_exception(self, exception: Exception):
-        raise exception
+        error_message = f"{exception}"
+        logger.exception(error_message)
```
papr:
```suggestion
        logger.error(error_message)
```
Even though `logger.exception()` is appropriate here, I do not know if the zmq log handler is able to handle the included exception correctly.

Issue #1493, opened 2019-04-27
Binocular gaze mapper sometimes has negative time jumps


cboulay:
Current mapping logic:
* If eye0 (right) cache has a low confidence sample then use that monocular
* elif the eye1 (right) cache has a low confidence sample then use that.
* elif both caches have data and the timestamps aren't too far then do binocular
* elif the timestamps are far apart then do monocular with the oldest
* elif only 1 cache, was high confidence, and overflowing, then use the oldest sample

These last 2 paths can lead to negative time jumps in the gaze stream.

I've implemented a different mapping logic, as follows:
* High-confidence samples will attempt binocular mapping if the other eye also has a high-confidence sample in its cache that is within time-delta temporal_cutoff. Otherwise monocular mapping will occur.
* We add the sample to the cache and process the cache, doing monocular mapping on any (low-confidence) samples that are newer than the previous last_timestamp, but older than the new last_timestamp.
* Added GUI elements for confidence, temporal cutoff, buffer length, and whether or not to process low confidence samples.

Key differences:
* Never returns a datum with a timestamp older than the last returned datum.
* The same timestamp can be sent multiple times. e.g., if we have a very recent high-conf timestamp in eye1, and eye0 is sending multiple samples in a row all of which are older than our eye1 sample, then that same eye1 sample will be combined with each eye0 sample that has high confidence.
* Upon a new high-confidence sample in eyeY, any samples in eyeX older than the previously sent timestamp are discarded without ever being processed. This can happen when eyeX is accumulating low-conf samples while eyeY gets multiple high-conf samples in a row.

cboulay:
The implementation is here:
https://github.com/cboulay/pupil/commit/c8e1aa96b3790598c5f679d926d7adb2c929a7e9

It is currently based on my 2 outstanding PRs. If you want me to make a PR out of this then I can rebase it against master, though I can't test it.

cboulay:
A couple of required adjustments...

* If it's using timestamps then it'll have to monitor time-resetting notifications then reset its last timestamp (to -inf) when that happens. This is easy enough to add (I think).
* If every timestamp from eyeX is always older than the cached timestamp from eyeY (e.g., eyeX processing is ~ 2 frames behind), but eyeY is low-confidence, binocular mapping will fail and the frame from eyeX will be ignored even if high-confidence. If this continues then the gaze will always be monocular from the low-confidence eyeY and eyeX cannot help.

[...]

PR #1455, opened 2019-02-28
Simpler Windows build using conda

cboulay:
New install instructions:

#### Developer Tools Requirements

If you don't already have them installed, you'll need git, conda (miniconda or anaconda), and MS Visual Studio 2017.

#### pupil dependencies

Get your Python environment ready. Install miniconda (or use Anaconda if you already have it installed), open an Anaconda Prompt and enter the following:

* `conda create -n pupil python=3.6 opencv boost eigen ceres-solver ffmpeg glfw glew`
* `conda activate pupil`
* `pip install pyzmq msgpack==0.5.6 pyopengl numpy pyaudio cython scipy psutil pyrealsense2`

There are some additional dependencies that are not found on pypi. The URLs below were correct at the time of writing but they may need to be updated. For the pupil derived packages, you can find them on their GitHub release pages. For pytorch, you can generate the pip command on the [pytorch website](https://pytorch.org/).

* `pip install --upgrade https://github.com/pupil-labs/pyglui/releases/download/v1.25/pyglui-1.25-cp36-cp36m-win_amd64.whl https://github.com/pupil-labs/PyAV/releases/download/v0.4.2/av-0.4.2.dev0-cp36-cp36m-win_amd64.whl https://github.com/pupil-labs/pyndsi/releases/download/v0.5/ndsi-0.5-cp36-cp36m-win_amd64.whl https://github.com/pupil-labs/pyuvc/releases/download/v0.13/uvc-0.13-cp36-cp36m-win_amd64.whl`
* `pip install https://download.pytorch.org/whl/cu100/torch-1.1.0-cp36-cp36m-win_amd64.whl torchvision`
* `pip install git+https://github.com/zeromq/pyre.git`
* `pip install git+https://github.com/pupil-labs/nslr`
* `pip install git+https://github.com/pupil-labs/nslr-hmm`

Finally we need to make sure conda has all the latest packages, because the ceres-solver fix is quite new:
* `conda update --all`

##### Get the source

The following steps assume you are still in an Anaconda Prompt with your pupil environment activated.

* `git clone https://github.com/pupil-labs/pupil.git`
* `cd pupil/pupil_src`

Note: If you want to test these changes before this is merged into pupil, you'll have to use Chad's branch:
* `git remote add -f cboulay https://github.com/cboulay/pupil.git`
* `git checkout cboulay/win_vcpkg_build`

##### Build the compiled modules

[...]

PR #1541, opened 2019-06-27
Binocular mapper -- no temporal reverse

cboulay:
I decided to make a pull request out of #1493 to facilitate discussion and monitor changes/commits. 

The current binocular gaze-mapping logic will sometimes use cached monocular samples that are older than the previous pushed-through sample, leading to negative time jumps in the gaze data stream. This is problematic for fixation/saccade detection algorithms.

I wrote a new binocular gaze mapper that takes a different approach. Please refer to the description in #1493 for a basic outline of the new logic.

There's only 1 file that has changes relevant to this PR: pupil_src/shared_modules/calibration_routines/gaze_mappers.py

The rest of the changes are the same as in #1455 , because I need those changes to run from source.

If you have a recording that includes calibration at the beginning then you can test these changes "online" without collecting data following [this gist](https://gist.github.com/cboulay/7b13d074730d454c9a3e29e91c0a8956).


cboulay:
I've reworked the binocular mapper a couple more times since the last update. This latest version works pretty well.

There's only one "feature" that I'm not sure about. Previously, in the original binocular gaze mapper...
* a sample would be held if it was high confidence and there wasn't a sample from the other eye that was also high confidence.
* if the cache length surpassed a limit (10 samples) then it would always output the first sample in the cache.
* only one mapped-sample could be output per iteration.

The result was that if you had a series of low-confidence samples in a row then the cache would fill up and stay full, and it would output exactly 1 sample per iteration for the rest of the run time.

In my updated mapper, I can push through a chunk of mapped samples that meet some criteria, almost clearing out the cache. This means that on some iterations multiple sample are output, and on other iterations there isn't anything to output: I end up returning `[]`. This creates a bit of a flicker in the cursor in the World or in hmd-eyes.

So, @papr or others, is it acceptable to sometimes return `[]`? Or should I make sure there is a mapped sample returned on every iteration? I think it's impossible to do that while guaranteeing no negative time jumps, but we can get close if we are confident that we will never get more than X samples from eye 0 in a row before we get a single sample from eye 1.

papr:
I think this is a great analysis of the previous behavior. I am in favor of pushing the chunks out instead of blocking the queue. Later, in player, it won't matter since player correlates everything by pupil detection time, not mapping time. I will be able to review the code in the coming week

cboulay:
Use #1557 instead.

PR #1552, opened 2019-07-16
Remove boost dependency

pfaion:
Boost was not actually used for anything in the singleeyefitter but for dead code. Additionally boost logging was used in the PyCppLogger, which again was not used anywhere. Besides that boost was a huge compile obstacle on windows, as especially boost.python is very old and outdated.

I removed all dependencies and adjusted the setup scripts.
In the process I also updated the explicit opencv dependency on windows from 3.2.0 to 3.4.5 as this is the currently stable 3.x release. We should check if we can actually update to 4.x in the future and maybe even remove the explicit version dependency for windows setup.


pfaion:
This is tested on windows, but still needs to be tested on Linux and OSX.

pfaion:
Tested on Linux as well: working!

PR #1555, opened 2019-07-18
Surface tracker bugfix

marc-tonsen:
- Fixed a bug where the offline surface tracker did not use robust marker detection correctly, which lead to a worse marker detection performance. 

- Also set the minimum marker detection confidence to zero to mirror behavior of previous version of surface tracker.

- Fixed a bug where the heatmap export had switched horizontal and vertical resolution.


papr:
@romanroibu Please integrate these fixes into #1540

marc-tonsen:
Update: Reverted a previous commit that removed mapping of fixations to the surface during export.

romanroibu:
@papr by "integrate" do you mean "apply the same diff to the `surface-tracker-apriltag` branch"? or just pull the latest `master` once this PR is merged? I think the last option is better, and it would be straight-forward to fix any conflicts

papr:
@romanroibu Please review the changes I made some further changes regarding mapped fixation export in https://github.com/pupil-labs/pupil/pull/1555/commits/6cb24844c89231c58b5153dfdda30725d0397175 and https://github.com/pupil-labs/pupil/pull/1555/commits/5fd35e08efeb894b969d41361d15b9ffcb349a39

pupil_src/shared_modules/surface_tracker/background_tasks.py
```diff
@@ -479,3 +510,48 @@ def _export_gaze_on_surface(self, gazes_on_surface, surface, surface_name):
...
+            encoding="utf-8",
+            newline="",
+        ) as csv_file:
+            csv_writer = csv.writer(csv_file, delimiter=",")
+            csv_writer.writerow(
+                (
+                    "fixation_id",
+                    "start_timestamp",
+                    "duration [ms]",
+                    "dispersion [deg]",
```
romanroibu: @papr any particular reason to have `[ms]` and `[deg]` here? are there any backwards-compatibility or consistency reasons? I'm asking because all the other labels follow the rules for identifier, so these 2 look out of place a bit.

papr: Agreed. I thought it would be nice to add some self-documentation to the headers but you are right that it would be inconsistent. The correct place to put this information is the official documentation.

PR #1557, opened 2019-07-20
binoc mapper - force monotonic timestamps

cboulay:
The previous algorithm to combine samples from 2 eyes had a couple problems.
* Every time a high-confidence sample was followed by a low-confidence sample, there would be no output and the cache would be extended by 1 up to a limit of 10. But the mapper only ever output up to 1 sample per input so it couldn't catch up. Effectively the mapper was always 10 samples behind.
* Sometimes there would be a negative time difference when switching from binocular mapping to monocular mapping on low confidence samples, because the monocularly mapped sample may have an older timestamp than the previous binocularly mapped sample.

In this pull request, a different algorithm is used to select pupil samples and combine them for the binocular mapper. Some highlights:
* Only a single cache is used (instead of one per eye) and new samples are inserted into the cache according to its timestamp to preserver order.
* The cache is processed from oldest-to-newest. A sample can only be mapped and pushed if there is a _newer_ sample from the _other_ eye. Otherwise it is possible the next sample from the other eye would be inserted before the sample about to be mapped and pushed.
* If the cache builds up (up to limit `sample_cutoff`) while waiting for a sample from the other eye to appear, multiple samples can be cleared out of the cache in one iteration if they meet criteria.
* The timestamp associated with a binocularly mapped sample comes from the timestamp of the oldest sample in the pair. This is the easiest way to maintain monotonic timestamps while keeping the latency low. (The alternative would be to maintain a separate cache for mapped timestamps and have a separate rule for outputting those in order, adding delay).
* Some of the parameters are tunable via the GUI.

This is my 4th attempt at fixing this algorithm, and replaces #1493 and #1541 . To make it easier to evaluate, this PR is rebased on master and DOES NOT include changes in my easy-build PR. Further, to make it easier to test out and debug, you can modify and use [this gist](https://gist.github.com/cboulay/7b13d074730d454c9a3e29e91c0a8956) to replay previously acquired data and still trigger the gaze mapper.

Below are a couple figures produced from LabStreamingLayer streams for Pupil and OpenVR while using a demo scene from hmd-eyes beta. The red lines are the head angles reported by OpenVR, and the blue lines are the gaze position reported by pupil. These were recorded while I kept fixation on a target and rotated my head left-right a few times, then up-down a few times. This may be of interest to @performlabrit

In the first image, gaze_normal_ channels are used, but sometimes these are `nan` so there are discontinuities in the data. In the second image, `norm_pos` channels are used, but sometimes during low-confidence there's a bit of noise as the data switches between binocular estimates and monocular estimates. Overlaid on top of these noisy lines in black are lines with low-confidence samples excluded.

openvr_and_pupil_gaze_normals

openvr_and_pupil_norm_pos


pupil_src/shared_modules/calibration_routines/gaze_mappers.py
```diff
@@ -76,68 +76,128 @@ class Binocular_Gaze_Mapper_Base(Gaze_Mapping_Plugin):
...
+        self.menu.append(ui.Slider("temporal_cutoff", self._cfg, step=0.002, min=0.0, max=0.2,
+                                   label="Max binocular time diff. (s)",))
+        self.menu.append(ui.Slider("sample_cutoff", self._cfg, step=1, min=2, max=20,
+                                   label="Max queue length", ))
 
     def map_batch(self, pupil_list):
-        current_caches = self._caches
-        self._caches = (deque(), deque())
+        backup_state = self._state
+        self._reset()
```
papr: Unfortunately, this also resets `backup_state`  since it's `deque`s are only referenced, not deep-copied.

pupil_src/shared_modules/calibration_routines/gaze_mappers.py
```diff
@@ -76,68 +76,128 @@ class Binocular_Gaze_Mapper_Base(Gaze_Mapping_Plugin):
...
         super().__init__(g_pool)
+        self._cfg = {'min_pupil_confidence': 0.6,
+                     'temporal_cutoff': 0.03,
+                     'timestamp_reset_threshold': 5.0,
+                     'process_low_confidence': True,
+                     'sample_cutoff': 10}
+        self._state = {'cache': deque(),
+                       'cache_ts': deque(),
+                       'cache_id': deque(),
+                       'cache_conf': deque()}
```
papr: I would recommend to extract the state into its own class. Additionally, instead of managing 4 separate univariate deques, I would recommend to use a single queue with multi-variate values (if at all necessary). The only real reason to have more than just the datum's deque is `in_ix = np.searchsorted(self._state["cache_ts"], p["timestamp"])`, and even there one could argue that one could replace the extra deque with a list comprehension, building the list on demand.

The main argument here is that keeping 4 lists coherent is much more work (in the sense of maintenance) than managing a single list.

pupil_src/shared_modules/calibration_routines/gaze_mappers.py
```diff
@@ -76,68 +76,128 @@ class Binocular_Gaze_Mapper_Base(Gaze_Mapping_Plugin):
 
     def __init__(self, g_pool):
         super().__init__(g_pool)
+        self._cfg = {'min_pupil_confidence': 0.6,
+                     'temporal_cutoff': 0.03,
+                     'timestamp_reset_threshold': 5.0,
+                     'process_low_confidence': True,
+                     'sample_cutoff': 10}
```
papr: What is the reason for storing the configuration in `_cfg` instead of the class itself?

cboulay: As far as I can tell, that's what the GUI expects. These variables are all exposed and can be modified via the GUI.

pupil_src/shared_modules/calibration_routines/gaze_mappers.py
cboulay: Thanks. I don't use `deque` very much. It's fixed in the new commit.

pupil_src/shared_modules/calibration_routines/gaze_mappers.py
cboulay: Agreed. They were left over from when I was pulling my hair out trying to figure out why I was generating garbage data. The state is simplified to a single cache of data samples.

cboulay:
Feel free to squash the commits before merging.

papr:
We have evaluated several gaze matching algorithms and have decided to go for #1731.

You can find the evaluation here: https://nbviewer.jupyter.org/github/pupil-labs/pupil-matching-evaluation/blob/master/gaze_mapper_evaluation.ipynb

@cboulay We sincerely appreciate the work that you have put into this PR. I want to apologize for us taking to long for properly evaluate this and other PRs.

With https://github.com/pupil-labs/pupil-matching-evaluation, we hope that we have created a platform that makes it easier to evaluate and improve Pupil's gaze matching algorithms.

cboulay:
Does that mean the new mapper output (and thus the LSL stream) does not have monotonic timestamps? If so then I'll need to adjust my downstream processing.

cboulay:
> Unfortunately, the implementation matches low and high confidence pupil data which potentially can deteriorate the spatial gaze mapping accuracy.

This could have been a solvable issue either by you, or by me if you kept me in the loop.

papr:
> This could have been a solvable issue either by you, or by me if you kept me in the loop.
We have been working on this the last two days. There was no long loop to keep anyone in. :) Feel free to evaluate any fixes/changes to your algorithm using the repository. Ideally, include solutions from lessons learnt so far, e.g. adaptive temporal cutoff.

> If so then I'll need to adjust my downstream processing
Yes, basically you just need to adjust to sort the entries by `time_stamp`.

This can be easily done using Pandas in Python:

```python
import pyxdf
import pandas as pd

path = ...

stream = pyxdf.load_xdf(path, select_streams=[{"name": "pupil_capture"}])[0]
data = stream["time_series"]
channels = stream["info"]["desc"][0]["channels"][0]["channel"]
headers = [chan["label"][0] for chan in channels]
index = stream["time_stamps"]
df = pd.DataFrame(data, index=index, columns=headers)
df.sort_index(inplace=True)  # Sorts by ts
```
The current implementation is a trade-off between mapping low confidence data as soon as possible and binocularly matching high confidence data as much as possible.

papr:
We can also add a further channel to the pupil capture lsl relay to facilitate filtering monocular and binocular data.

papr:
Nonetheless, I would like to apologize for the lack of communication. I could have notified you earlier, that is true. I will do so in the future.

cboulay:
> Yes, basically you just need to adjust to sort the entries by time_stamp.

It's not that simple. One use case is to do online saccade event detection. These algorithms require the data to be monotonic, especially if we're interpolating, smoothing, taking the diff to get velocity, etc.

If the timestamps are not monotonic then we can't do this. We can maintain a buffer >= maximum temporal cutoff (what is that now?) and only process data that old. This adds unnecessary delay and is less than ideal.

papr:
Temporal cutoff is set to twice the estimated frame rate.

PR #1562, opened 2019-07-30
Improve head pose tracker

ChingT:
1. Improve the effectiveness of optimization of 3d markers model
1. Filter duplicate detected markers

### DIFF
--- a/pupil_src/shared_modules/head_pose_tracker/worker/detection_worker.py
+++ b/pupil_src/shared_modules/head_pose_tracker/worker/detection_worker.py
@@ -9,11 +9,15 @@ See COPYING and COPYING.LESSER for license details.
 ---------------------------------------------------------------------------~(*)
 """
 
+import logging
+from types import SimpleNamespace
+
+import apriltag
 import file_methods as fm
 import video_capture
-import apriltag
 from methods import normalize
-from types import SimpleNamespace
+
+logger = logging.getLogger(__name__)
 
 apriltag_detector = apriltag.Detector()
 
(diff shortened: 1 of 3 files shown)


pupil_src/shared_modules/head_pose_tracker/worker/detection_worker.py
```diff
@@ -34,7 +35,7 @@ def get_markers_data(detection, img_size, timestamp):
 def _detect(frame):
     image = frame.gray
     apriltag_detections = apriltag_detector.detect(image)
-    apriltag_detections = _filter_duplicate_markers(apriltag_detections)
+    apriltag_detections = unique(apriltag_detections, key=lambda marker: marker.tag_id)
```
romanroibu: @papr maybe I'm wrong, but it seems to me that @ChingT 's implementation of `_filter_duplicate_markers` was not only filtering out repeated occurrences of duplicate markers (but leaving the first occurrence), but was rejecting all occurrences of a marker that has duplicates. not sure if you're commit is an intentional change of this behavior; just an observation.

romanroibu: if you do want to keep the behavior, the following should do the trick:
```
apriltag_detections = unique(apriltag_detections, key=lambda marker: marker.tag_id, select=lambda _, _: None)
apriltag_detections = filter(lambda tag_id: tag_id is not None, apriltag_detections)
```

but this of course won't warn the user that some tags are ignored due to duplicates.

papr: @romanroibu  Good catch! @ChingT what do you think? Which behaviour would be preferable?

ChingT: I would prefer rejecting all occurrences of duplicate markers and having the warning so that the user could be aware of it and remove the duplicates. Since duplicate markers may break 3d markers model.

papr: Resolved by removing markers with smaller perimeter.

PR #1576, opened 2019-08-09
Allow a change of the ROI by network

Christopher22:
This pull request extends the modification of pupil detector properties by network. After the change, it is possible to modify the position and size of the ROI, too.

A minimum working example for testing may look like this:
```python
import zmq
import msgpack as serializer

ctx = zmq.Context()

ip = "localhost"
port = 50020

requester.connect("tcp://%s:%s" % (ip, port))
requester.send_string("SUB_PORT")
sub_port = requester.recv_string()

notification = {
    "subject": "pupil_detector.set_property",
    "target": "eye0",
    "name": "roi",
    "value": [0, 100, 0, 200],
}

topic = "notify." + notification["subject"]
payload = serializer.dumps(notification)
requester.send_string(topic, flags=zmq.SNDMORE)
requester.send(payload)

print(requester.recv_string())
```


papr:
@Christopher22 I updated your description since the example was not consistent with the implementation.

papr:
@Christopher22 Actually, your example was correct, and I misinterpreted the implementation. Apologies for that. I am reverting your description to the original.

PR #1584, opened 2019-08-14
Surface Tracker Heatmap Resolution

marc-tonsen:
If no gaze data is available on a surface (either there is none or it is still being mapped in the background) the generated heatmap used to have resolution 1x1px. This PR changes this to always use the resolution implied by the smoothness value.

### DIFF
--- a/pupil_src/shared_modules/surface_tracker/surface.py
+++ b/pupil_src/shared_modules/surface_tracker/surface.py
@@ -528,7 +528,7 @@ class Surface(abc.ABC):
             hist *= (255.0 / hist_max) if hist_max else 0.0
             hist = hist.astype(np.uint8)
         else:
-            self.within_surface_heatmap = self.get_uniform_heatmap()
+            self.within_surface_heatmap = self.get_uniform_heatmap(grid)
             return
 
         color_map = cv2.applyColorMap(hist, cv2.COLORMAP_JET)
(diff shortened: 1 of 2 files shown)


marc-tonsen:
Looks like my git history on this branch is not clean. I'll fix this asap.

pupil_src/shared_modules/surface_tracker/surface.py
```diff
@@ -538,8 +538,8 @@ def update_heatmap(self, gaze_on_surf):
             self.within_surface_heatmap[:, :, 3] = 125
         self.within_surface_heatmap[:, :, :3] = color_map
 
-    def get_uniform_heatmap(self):
-        hm = np.zeros((1, 1, 4), dtype=np.uint8)
+    def get_uniform_heatmap(self, resolution):
+        hm = np.zeros((*resolution, 4), dtype=np.uint8)
         hm[:, :, :3] = cv2.applyColorMap(hm[:, :, :3], cv2.COLORMAP_JET)
```
papr: This should ensure that resolution has the correct number of elements or fail right there.

pupil_src/shared_modules/surface_tracker/surface_tracker_offline.py
```diff
@@ -337,6 +337,21 @@ def _start_gaze_buffer_filler(self, all_gaze_events, all_world_timestamps, secti
             mp_context,
         )
 
+    def _start_fixation_buffer_filler(
```
papr: I removed that because I could not find any usage of this function. In case this function is needed, fixations_on_surf_buffer_filler needs to be defined somewhere outside of this function or the if condition raises an attribute error.

marc-tonsen:
@papr I incorporated your two comments and cleaned up the git history.

PR #1585, opened 2019-08-14
Improve invalid recording errors

### LINKED ISSUE #1428: Player: Improve "Invalid Recording" error message
Currently, one gets the same error message for different cases:
- User drops a video file instead of the recording folder (wrong usage, message: "No valid dir supplied")
- User drops a folder that does not contain (actual invalid recording)
    - an `info.csv` (message: "Could not read info.csv file: Not a valid Pupil recording.")
    - any video files (message: "Could not generate world timestamps from eye timestamps. This is an invalid recording.")
- `info.csv` does not contain `Recording Name` (message: "Could not read info.csv file: Not a valid Pupil recording.")
- Error during parsing `info.csv` (message: "Could not read info.csv file: Not a valid Pupil recording.")

romanroibu:
Fixes  #1428

### DIFF
--- a/pupil_src/shared_modules/player_methods.py
+++ b/pupil_src/shared_modules/player_methods.py
@@ -8,6 +8,7 @@ Lesser General Public License (LGPL v3.0).
 See COPYING and COPYING.LESSER for license details.
 ---------------------------------------------------------------------------~(*)
 """
+import glob
 import logging
 import os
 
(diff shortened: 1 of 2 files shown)


pupil_src/shared_modules/player_methods.py
```diff
@@ -194,23 +195,124 @@ def correlate_data(data, timestamps):
...
+        return True
+    except InvalidRecordingException as e:
+        logger.error(str(e))
         return False
-    return True
+
+
+class InvalidRecordingException(Exception):
+    def __init__(self, reason: str, recovery: str = ""):
+        message = reason + "\n" + recovery if recovery else reason
```
papr: This prints reason twice in some cases. Is that intended?

pupil_src/shared_modules/player_methods.py
```diff
@@ -194,23 +195,124 @@ def correlate_data(data, timestamps):
...
+    def reload(self):
+        self.load(rec_dir=self.rec_dir)
+
+    def load(self, rec_dir):
+
+        def is_video_file(file_path):
+            if not os.path.isfile(file_path):
+                return False
+            # TODO: Are there any other valid extensions?
+            valid_video_extensions = [".mp4"]
```
papr: Yes, there are multiple valid extensions. please check video_capture.utils.VideoSet for a set of valid extensions IIRC /cc Patrick

pupil_src/shared_modules/player_methods.py
```diff
@@ -194,23 +195,124 @@ def correlate_data(data, timestamps):
...
+            _, ext = os.path.splitext(file_path)
+            if ext not in valid_video_extensions:
+                return False
+            return True
+
+        if not os.path.exists(rec_dir):
+            raise InvalidRecordingException(
+                reason=f"Target at path does not exist: {rec_dir}",
```
papr: What is a target? Maybe replace with "dragged files/folder"?

pupil_src/shared_modules/player_methods.py
```diff
@@ -194,23 +195,124 @@ def correlate_data(data, timestamps):
...
+            except KeyError:
+                raise InvalidRecordingException(
+                    reason=f"Target info.csv does not have \"{key}\"",
+                    recovery=""
+                )
+
+        all_file_paths = glob.iglob(os.path.join(rec_dir, "*"))
+
+        # TODO: Should this validation be "are there any video files" or are there specific video files?
+        if not any(is_video_file(path) for path in all_file_paths):
```
papr: I think the current behavior requires at least one video but I would like to change that to not require anything but an info.csv file with start time and duration keys. See Patrick's implementation for generating wordless lookup tables.

pupil_src/shared_modules/player_methods.py
romanroibu: it's not intended. could you please tell me in which cases you're seeing `reason` twice?

papr: I think I misread the priority of the operators. Could you please add round brackets for more explicit grouping?

pupil_src/shared_modules/player_methods.py
romanroibu: done

pupil_src/shared_modules/player_methods.py
romanroibu: I considered this, but since this class has no knowledge of who and how is passing it the `rec_dir` path, `"dragged"` is too specific and can be misleading too.

`"target"` here pretty much means `"filesystem node"`, whatever it may be (directory, file, link, etc.), since this is what we're testing with `os.path.exists`.

papr: Agreed.

pupil_src/shared_modules/player_methods.py
romanroibu: done

pupil_src/shared_modules/player_methods.py
romanroibu: so what is the correct check here? only the presence of `"Start Time"` and `"Duration Time"` in `info.csv`?

I'm looking at `check_for_worldless_recording` in `update_methods.py` but there's no reference to either key. also it seems that this method is generating a `world.fake` file if a real video file is missing. `is_video_file(path)` will return `True` for any video file, including `world.fake`, so the implementation seems to still be valid.

papr: Let's talk this through on Thursday.

Issue #1590, opened 2019-08-20
Surface tracker: fix edit mode vertices stuck to mouse pointer


romanroibu:
When moving surface vertices in edit mode the release click would sometimes be ignored and the vertex was stuck on the mouse pointer for ever. This was not the case before and needs to be fixed.

(reported by @mkassner)

romanroibu:
One way to reproduce:
- enable surface editing
- left-click and hold a corner
- move the mouse fast while holding the click
- while moving - release the click

Very often this results in the corner being stuck to the mouse. If there is an offset between the corner position and the mouse pointer position, it is impossible to pin the corner to a location.

papr:
Fixed via #1591

Issue #1589, opened 2019-08-20
Surface tracker: restore ability to separate overlapping surfaces


romanroibu:
When using the menus that are attached to the top right corner of each surface, it used to be that one would take precedent and consume the input. This is no longer the case.

The problem is that two surfaces on top of each other can never be separated anymore. This needs to be changed back to the old behavior.

(reported by @mkassner)

papr:
Fixed via #1591

PR #1591, opened 2019-08-20
Surface tracker - fix GUI bugs

romanroibu:
- Fixes issue (#1589) where it was impossible to separate overlapping surfaces, because enabling edit mode for one surface would enable it for both, and moving the corner of one surface would move both corners of the overlapping surfaces.
- Fixes issue (#1590) where attempting to move a surface's corner would result in that corner being stuck to the mouse pointer.
- Fixes issue where moving the corner of surface A under the "edit" button of surface B would then make it impossible to move that corner elsewhere.

### DIFF
--- a/pupil_src/shared_modules/surface_tracker/gui.py
+++ b/pupil_src/shared_modules/surface_tracker/gui.py
@@ -391,7 +391,7 @@ class GUI:
 
     def _on_click_menu_buttons(self, action, pos):
         if action == glfw.GLFW_PRESS:
-            for surface in self.tracker.surfaces:
+            for surface in reversed(self.tracker.surfaces):
 
                 if not surface.detected:
                     continue
--- a/pupil_src/shared_modules/surface_tracker/gui.py
+++ b/pupil_src/shared_modules/surface_tracker/gui.py
@@ -385,7 +385,9 @@ class GUI:
             self._on_click_menu_buttons,
             self._on_click_marker_toggles,
         ]
-        _ = any(handler(action, pos) for handler in click_handlers_sorted_by_precedence)
+        for handler in click_handlers_sorted_by_precedence:
+            if handler(action, pos):
+                return
 
     def _on_click_menu_buttons(self, action, pos):
         if action == glfw.GLFW_PRESS:
--- a/pupil_src/shared_modules/surface_tracker/gui.py
+++ b/pupil_src/shared_modules/surface_tracker/gui.py
@@ -380,12 +380,12 @@ class GUI:
 
     def on_click(self, pos, button, action):
         pos = np.array(pos, dtype=np.float32)
-        click_handlers = [
-            self._on_click_menu_buttons,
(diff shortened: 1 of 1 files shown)


pupil_src/shared_modules/surface_tracker/gui.py
```diff
@@ -380,9 +380,12 @@ def _draw_heatmap(self, surface):
...
         pos = np.array(pos, dtype=np.float32)
-        self._on_click_menu_buttons(action, pos)
-        self._on_click_corner_handles(action, pos)
-        self._on_click_marker_toggles(action, pos)
+        click_handlers_sorted_by_precedence = [
+            self._on_click_corner_handles,
+            self._on_click_menu_buttons,
+            self._on_click_marker_toggles,
+        ]
+        _ = any(handler(action, pos) for handler in click_handlers_sorted_by_precedence)
```
pfaion: I think using the early-exit implementation of `any()` is misleading here. This is much clearer in its intent I feel:
```python
for handler in click_handlers_sorted_by_precedence:
    if handler(action, pos):
        return
```

romanroibu: I agree with the point about not being very clear in the intent, but I was following precedent set in `world.py` ([1][1], [2][2], [3][3], [4][4]), `player.py` ([5][5], [6][6], [7][7], [8][8]) and `eye.py` ([9][9]). I think we should settle on one way to describe this concept (early exist on event consumption) and use it consistently, whichever it is.

@papr thoughts?

[1]: https://github.com/pupil-labs/pupil/blob/master/pupil_src/launchables/world.py#L353
[2]: https://github.com/pupil-labs/pupil/blob/master/pupil_src/launchables/world.py#L671
[3]: https://github.com/pupil-labs/pupil/blob/master/pupil_src/launchables/world.py#L676
[4]: https://github.com/pupil-labs/pupil/blob/master/pupil_src/launchables/world.py#L681
[5]: https://github.com/pupil-labs/pupil/blob/master/pupil_src/launchables/player.py#L212
[6]: https://github.com/pupil-labs/pupil/blob/master/pupil_src/launchables/player.py#L630
[7]: https://github.com/pupil-labs/pupil/blob/master/pupil_src/launchables/player.py#L635
[8]: https://github.com/pupil-labs/pupil/blob/master/pupil_src/launchables/player.py#L640
[9]: https://github.com/pupil-labs/pupil/blob/master/pupil_src/launchables/eye.py#L244

pfaion: The fact that *ALL* those references have a two-line comment preceding, which explains what is happening afterwards with the `any()` confirms my suspicion 😃 

> Explicit is better than implicit.
-- https://www.python.org/dev/peps/pep-0020/

papr: Agreed, please change these occurrences to the more explicit implementation in a new PR.

Issue #1588, opened 2019-08-20
Surface tracker crashes on windows


pfaion:
The surface tracker seems to consistently crash on windows (but not on Linux) when
a) using a lot of surfaces (>10) at the same time and
b) when performing an action like moving the corner of a surface before the results of a previous action have finished computing.

This was feedback from ETRA. Not yet reproduced.

papr:
We were not able to reproduce the crash itself, but noticed that Surface Tracker does not perform well on Windows when one has defined a lot of surfaces. This is due to each surface starting its own background processes for different tasks. This generates a lot of overhead on systems that use `spawn`, i.e. Windows.

papr:
@mkassner has proposed to use `threading` instead of `multiprocessing`. @romanroibu will look into the implementation of this proposal.

PR #1604, opened 2019-08-26
[DON'T MERGE] Surface tracker fix background processing

romanroibu:
Addresses #1588


pupil_src/shared_modules/background_helper/base.py
```diff
@@ -0,0 +1,48 @@
...
+Copyright (C) 2012-2019 Pupil Labs
+
+Distributed under the terms of the GNU
+Lesser General Public License (LGPL v3.0).
+See COPYING and COPYING.LESSER for license details.
+---------------------------------------------------------------------------~(*)
+"""
+
+import abc
+import typing
```
pfaion: Let's use `import typing as T` for consistency.
Also in the other files.

pupil_src/shared_modules/background_helper/base.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class EarlyCancellationError(Exception):
+    pass
+
+
+class Task_Proxy(abc.ABC):
```
pfaion: 1. Remove underscore from class names

2. Why is this not called `BaseTaskProxy`? Same with the 2 different implementations (threaded vs multiprocessing). I see that you are exposing them with different names through the module's `__init__.py`, why not name them differently in the first place? This makes it a lot easier to search for them as well.

pupil_src/shared_modules/background_helper/base.py
```diff
@@ -0,0 +1,48 @@
...
+    def __init__(self, name: str, generator, args=(), kwargs={}, **_):
+        self._name = name
+
+    @abc.abstractmethod
+    def fetch(self) -> typing.Iterator[typing.Any]:
+        """Fetches progress and available results from background"""
+        pass
+
+    @abc.abstractmethod
+    def cancel(self, timeout=1):
```
pfaion: Add a comment for behavior of timeout.

papr:
Similar to #1637 Surface Tracker is in a manageable state right now. Closing this PR until the situation changes.

PR #1609, opened 2019-08-28
Fill gaps on video overlays

papr:


pfaion:
@papr There are still visualizations for the last/next pupil location in player (jumping once in the middle of the gap). Same in the exported video. Maybe we should fix this as well? Looks even weirder with just the pupil positions being displayed there to be honest.
image

papr:
Agreed, this needs to be fixed.

pfaion:
@papr Have a look at my changes!
(Unfortunately I can't assign you as reviewer to your own PR 😄 )

pupil_src/shared_modules/video_overlay/workers/overlay_renderer.py
```diff
@@ -39,8 +39,21 @@ def draw_on_frame(self, target_frame):
             return
         overlay_frame = self.video.closest_frame_to_ts(target_frame.timestamp)
         overlay_image = overlay_frame.img
+        try:
+            is_fake_frame = overlay_frame.is_fake
+        except AttributeError:
+            # TODO: this is only fore extra safety until we have a unified Frame class
+            # and can be sure that everything ending up here has the 'is_fake'
+            # attribute!
+            logger.debug(
```
papr:
```suggestion
            logger.warning(
```

pupil_src/shared_modules/video_overlay/utils/image_manipulation.py
```diff
@@ -6,26 +6,32 @@
...
 class ScaleTransform(ImageManipulator):
-    def apply_to(self, image, parameter):
+    def apply_to(self, image, parameter, **kwargs):
         """parameter: scale factor as float"""
         return cv2.resize(image, (0, 0), fx=parameter, fy=parameter)
 
 
 class HorizontalFlip(ImageManipulator):
-    def apply_to(self, image, parameter):
+    def apply_to(self, image, parameter, *, is_fake_frame, **kwargs):
```
romanroibu: @papr why does `ImageManipulator` take an `image` as an argument and not a `frame`? if it did, the `is_fake_frame` argument wouldn't be needed, since all the relevant info to perform the manipulation would be provided in `frame`.

pupil_src/shared_modules/video_overlay/workers/overlay_renderer.py
```diff
@@ -39,8 +39,21 @@ def draw_on_frame(self, target_frame):
             return
         overlay_frame = self.video.closest_frame_to_ts(target_frame.timestamp)
         overlay_image = overlay_frame.img
+        try:
+            is_fake_frame = overlay_frame.is_fake
+        except AttributeError:
```
romanroibu: another way would be to check if `overlay_frame` is an instance of `Frame.FakeFrame`, since only an instance of this class should be considered a "fake frame". this would remove any possibility of getting an instance which does not have the `is_fake` attribute.

pupil_src/shared_modules/video_overlay/utils/image_manipulation.py
pfaion: @papr I like the idea of passing the whole frame object to the image manipulators.
Anything speaking against this? Should I change it that way?

papr: @romanroibu @pfaion The initial idea was to pass the minimum required data object (the image buffer). Since the requirements changed, one could argue the next smallest data object is the frame.

@pfaion feel free to make that change.

pupil_src/shared_modules/video_overlay/workers/overlay_renderer.py
papr: @romanroibu We thought about this, too, but came to the conclusion that it the attribute implementation would be more robust if we changed the frame class hierarchy at some point.

pupil_src/shared_modules/video_overlay/utils/image_manipulation.py
pfaion: @papr @romanroibu 
Turns out this is not so easy currently, as the image property in our frame class is read-only, as it is inferred from the underlying av_frame.

More specifically: In order for this to work, we'd have to change the ImageManipulators to receive and return frames. In `apply_to()` we'd have to modify the image of the frame. But the file_backend frames don't really have an image, but only an internal av_frame, that get's converted into a numpy-array when accessing the `Frame.img` property. Thus we can't consistently modify the image of a frame.

Maybe we'd be able to actually remove the dependency on the av_frame internally, but I'd argue to not do this now. This might become clearer, once we implement a clear pupil frame class that we use across all backends. We will have to touch this code again then anyways.

If we leave it like this, I'd just add a NOTE to change this when we refactor the frame class.
Thoughts?

papr: Leave as it is for now.

pfaion:
@papr 🚀

PR #1620, opened 2019-09-05
UVC Backend: Only call update_from_yuv_buffer if yuv subsampling is YUV422 

papr:
Previously, the uvc backend did not check the provided YUV subsampling if a YUV buffer was available. pyglui only supports YUV 422 subsampling. If `update_from_yuv_buffer` was called with a different subsampling than YUV 422 it resulted in instabilities and incorrect rendering.

The internal conversion from YUV to BGR takes the subsampling into account correctly. Therefore, `update_from_ndarray(frame.bgr)` works correctly.

This change was not implemented in `Base_Source` since not all backends provide the `yuv_subsampling` attribute.

### DIFF
--- a/pupil_src/shared_modules/video_capture/base_backend.py
+++ b/pupil_src/shared_modules/video_capture/base_backend.py
@@ -9,15 +9,14 @@ See COPYING and COPYING.LESSER for license details.
 ---------------------------------------------------------------------------~(*)
 """
 
+import logging
 from time import monotonic, sleep
-from plugin import Plugin
+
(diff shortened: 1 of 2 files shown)


pfaion:
@papr is there any way of reproducing the issues, such that I can do a before/after validation?

papr:
@pfaion yes, use a PI headset and connect to Capture.

pfaion:
@papr I guess the new expected behavior of PI on Capture is still ghost mode, but without errors?

PR #1621, opened 2019-09-05
Disable windows hardware timestamps

### LINKED ISSUE #1614: Disable hardware timestamps on Windows
We have seen too many recordings on Windows whose timestamps are out of sync. Currently, we use the hardware timestamps given by the camera. In the cases of the out-of-sync recordings, these timestamps either had a constant offset (multiple minutes) or drifted during the recording. We will be disabling the hardware timestamp feature and reverting to software timestamps on Windows.

Software frame timestamps are measured by the host computer when the frame has been received from the camera instead of when the frame started exposing (hardware timestamp). This means that software timestamps carry an inherent delay of a few milliseconds.

In the current situation, we are not able to continue supporting hardware timestamps since they do not provide the required robustness on Windows. If you require hardware timestamps you will have to use Pupil Capture on Linux or macOS.

pfaion:
Resolves #1614

### DIFF
--- a/pupil_src/shared_modules/video_capture/uvc_backend.py
+++ b/pupil_src/shared_modules/video_capture/uvc_backend.py
@@ -375,7 +375,7 @@ class UVC_Source(Base_Source):
     def recent_events(self, events):
         try:
             frame = self.uvc_capture.get_frame(0.05)
-            
+
             if np.isclose(frame.timestamp, 0):
                 # sometimes (probably only on windows) after disconnections, the first frame has 0 ts
                 logger.warning(
--- a/pupil_src/shared_modules/video_capture/uvc_backend.py
+++ b/pupil_src/shared_modules/video_capture/uvc_backend.py
@@ -389,7 +389,7 @@ class UVC_Source(Base_Source):
             time.sleep(0.02)
             self._restart_logic()
         else:
-            if self.ts_offset:
+            if self.ts_offset != 0:
                 # c930 timestamps need to be set here. The camera does not provide valid pts from device
                 frame.timestamp = uvc.get_time_monotonic() + self.ts_offset
             frame.timestamp -= self.g_pool.timebase.value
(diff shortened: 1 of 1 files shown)


pupil_src/shared_modules/video_capture/uvc_backend.py
```diff
@@ -180,7 +182,21 @@ def verify_drivers(self):
...
-            self.ts_offset = 0.0
+            if platform.system() == "Windows":
+                # NOTE: Hardware timestamps seem to be broken on windows. Needs further
+                # investigation! Disabling for now.
+                # TODO: Find accurate offsets for different resolutions!
+                offsets = {"ID0": -0.015, "ID1": -0.015, "ID2": -0.07}
+                match = re.match(r"Pupil Cam\d (?P<cam_id>ID[0-2])", self.name)
+                if not match:
+                    logger.debug(f"Could not parse camera name: {self.name}")
+                    self.ts_offset = 0.0
```
papr: Maybe implement a default offset of  `-0.01`? Also apply below for unknown cams
```suggestion
                    self.ts_offset = -0.01
```

pupil_src/shared_modules/video_capture/uvc_backend.py
```diff
@@ -180,7 +182,21 @@ def verify_drivers(self):
...
+                else:
+                    self.ts_offset = offsets[match.group("cam_id")]
+
+            else:
+                # use software timestamps
```
papr:
```suggestion
                # use hardware timestamps
```

pupil_src/shared_modules/video_capture/uvc_backend.py
```diff
@@ -180,7 +182,21 @@ def verify_drivers(self):
...
+                self.ts_offset = 0.0
```
papr:
```suggestion
                self.ts_offset = None
```

pupil_src/shared_modules/video_capture/uvc_backend.py
pfaion: @papr Hm, shouldn't we maybe use software timestamps for unknown cameras? (This is what this was supposed to be)

pfaion: In this case I'd do:
```suggestion
                    self.ts_offset = None
```

papr: `self.ts_offset = None` triggers hardware timestamps. In this specific case, we are on Windows. We do not want to use hardware timestamps in any case on Windows.

pfaion:
@papr :shipit:

PR #1631, opened 2019-09-12
Generate all lookup tables directly after updating a recording

pfaion:
This is to improve the user experience as the UI will then still display "Recording is being updated" while possibly computing a large lookup table.
Otherwise this happened after the pre-drop windows closes and then there is no feedback when generating the lookup tables takes a long time.

The import changes in player.py and pupil_recording/__init__.py were necessary to avoid circular imports.


pfaion:
@papr I tested it with an 8min recording and it improves the user experience not much, but noticeable

PR #1637, opened 2019-09-16
Surface tracker - surface location update optimization

romanroibu:


papr:
Currently the Surface Tracker is in a working state. Closing this PR until a fundamental change is required.

PR #1642, opened 2019-09-17
Extract pupil_detectors into a separate repo

romanroibu:
Requires [pupil-detectors](https://github.com/pupil-labs/pupil-detectors):
```bash
pip3 install git+https://github.com/pupil-labs/pupil-detectors.git --user --force -vvv
```


pupil_src/launchables/eye.py
```diff
@@ -784,26 +678,51 @@ def window_should_update():
...
-                    make_coord_system_pixel_based((f_height, f_width, 3), g_pool.flip)
+                    # make_coord_system_pixel_based((f_height, f_width, 3), g_pool.flip)
                     # render the ROI
                     g_pool.u_r.draw(g_pool.gui.scale)
                     if g_pool.display_mode == "roi":
                         g_pool.u_r.draw_points(g_pool.gui.scale)
 
                     glViewport(0, 0, *window_size)
-                    make_coord_system_pixel_based((*window_size[::-1], 3), g_pool.flip)
+                    # make_coord_system_pixel_based((*window_size[::-1], 3), g_pool.flip)
```
pfaion: @papr 
Why are these commented out?
Can this be removed?

pupil_src/shared_modules/pupil_detector_plugins/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+    """Load and list available plugins, including default
+    
+    Returns:
+        T.Tuple[PupilDetectorPlugin, T.List[PupilDetectorPlugin]]
+        --  Return tuple of default plugin, and list available plugins.
+            Default is required to be in the list of available plugins.
+    """
+    from .detector_dummy_plugin import DetectorDummyPlugin
+    from .detector_2d_plugin import Detector2DPlugin
+    from .detector_3d_plugin import Detector3DPlugin
```
pfaion: Can we import this at the top?
Or would this cause problems with import order?
Would also be good to expose those detectors via the module I think.

pupil_src/shared_modules/pupil_detector_plugins/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+    detector_plugins = [DetectorDummyPlugin, Detector2DPlugin, Detector3DPlugin]
+
+    try:
+        from py3d import Detector3DRefractionPlugin
+
+        detector_plugins.append(Detector3DRefractionPlugin)
+    except ImportError:
+        logging.info("Refraction corrected 3D pupil detector not available")
+
+    return Detector3DPlugin, detector_plugins
```
pfaion: To make it clearer, there should be a variable at the top
```python
default_plugin = Detector3DPlugin
```
And then later
```python
return default_plugin, detector_plugins
```

pupil_src/shared_modules/pupil_detector_plugins/manager.py
```diff
@@ -0,0 +1,75 @@
...
+from . import available_detector_plugins
+
+
+class PupilDetectorManager(Plugin):
+    order = 0.2
+
+    def __init__(self, g_pool):
+        super().__init__(g_pool)
+
+        self._available_pupil_detector_classes = available_detector_plugins()[1]
```
romanroibu: I think this improves readability. The previous version made me think that the property was set to the plugin at index `1`, which seemed arbitrary. This version makes it clear that the property is set to the second value returned by the function.

```suggestion
        _, self._available_pupil_detector_classes = available_detector_plugins()
```

pfaion: Done!

PR #1667, opened 2019-10-02
V1.16 fixes

papr:


pupil_src/shared_modules/storage.py
```diff
@@ -40,18 +47,15 @@ def create_new_unique_id():
...
     def create_unique_id_from_string(string):
         """
         Returns: A unique id as generated by create_new_unique_id(), but for
             the same input string, it will always return the same id.
         """
-        old_state_of_random = random.getstate()
-        random.seed(string)
-        unique_id = StorageItem.create_new_unique_id()
-        random.setstate(old_state_of_random)
+        unique_id = str(uuid.uuid5(UUID_NAMESPACE_PUPIL_LABS, string))
```
ckbaumann: Why here 5 and above 4 bytes? Otherwise it looks good!

papr: > `uuid.uuid4()`: Generate a random UUID.
> `uuid.uuid5(namespace, name)`: Generate a UUID based on the SHA-1 hash of a namespace identifier (which is a UUID) and a name (which is a string).

In one case we just need a random uuid, in the other case we need to be able to generate a reproducible uuid [based on a string](https://github.com/pupil-labs/pupil/blob/master/pupil_src/shared_modules/gaze_producer/model/calibration_storage.py#L138-L140).

Issue #1645, opened 2019-09-17
Add PI support for uvc backend auto-selection


papr:
Apply same change as in https://github.com/pupil-labs/pupil/pull/1525 to the `uvc_backend`.

papr:
Currently, it is not planned to add further features to the software.

PR #1674, opened 2019-10-08
Use PluginList in eye process

pfaion:
Eye already uses a couple of plugins but has to manage them manually.
This results in a lot of duplicate code that is all over the place but belongs together.
World uses the PluginList to extract this logic.

I added a PluginList to eye and extracted the plugins into it.
Some refactoring was required for resolving order-of-execution issues.

This will also allow us to make plugins out of the detectors and integrate them more cleanly,

### DIFF
--- a/pupil_src/launchables/eye.py
+++ b/pupil_src/launchables/eye.py
@@ -460,23 +460,6 @@ def eye(
         if roi_user_settings and tuple(roi_user_settings[-1]) == g_pool.u_r.get()[-1]:
             g_pool.u_r.set(roi_user_settings)
 
-        def replace_source(source_class_name, source_settings):
-            g_pool.capture.deinit_ui()
-            g_pool.capture.cleanup()
-            g_pool.capture = source_class_by_name[source_class_name](
-                g_pool, **source_settings
-            )
-            g_pool.capture.init_ui()
-            if g_pool.writer:
-                logger.info("Done recording.")
-                try:
-                    g_pool.writer.release()
-                except RuntimeError:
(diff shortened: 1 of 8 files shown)


pupil_src/launchables/eye.py
```diff
@@ -189,6 +192,32 @@ def get_timestamp():
         g_pool.get_timestamp = get_timestamp
         g_pool.get_now = get_time_monotonic
 
+        plugins = manager_classes + source_classes
+        g_pool.plugin_by_name = {p.__name__: p for p in plugins}
+
+        preferred_names = [
+            f"Pupil Cam3 ID{eye_id}",
+            f"Pupil Cam2 ID{eye_id}",
+            f"Pupil Cam1 ID{eye_id}",
```
papr: Add relevant names for PI eye cams. Maybe as an additional PR, together with changes to PI auto camera selection?

pupil_src/launchables/eye.py
```diff
@@ -76,6 +76,7 @@ def eye(
        ``recording.stopped``: Stops recording eye video
        ``frame_publishing.started``: Starts frame publishing
        ``frame_publishing.stopped``: Stops frame publishing
+        ``start_eye_plugin``: Start plugins in eye process
```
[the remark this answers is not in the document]
pfaion: Actually there was a space missing at all the other lines :D
Fixed.

pupil_src/launchables/eye.py
papr: Reference #1645

PR #1692, opened 2019-10-21
Fix surface corner drag inaccuracies

pfaion:
When editing surface corners, often the corners would not match with the mouse position.

Here's the fix. Most important is the 2nd commit.

### DIFF
--- a/pupil_src/shared_modules/surface_tracker/surface.py
+++ b/pupil_src/shared_modules/surface_tracker/surface.py
@@ -333,9 +333,29 @@ class Surface(abc.ABC):
         points_A = points_A.reshape((-1, 1, 2))
         points_B = points_B.reshape((-1, 1, 2))
 
-        B_to_A, mask = cv2.findHomography(
-            points_A, points_B, method=cv2.RANSAC, ransacReprojThreshold=100
-        )
+        B_to_A, mask = cv2.findHomography(points_A, points_B)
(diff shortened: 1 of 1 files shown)


pupil_src/shared_modules/surface_tracker/surface.py
```diff
@@ -333,9 +333,29 @@ def _find_homographies(points_A, points_B):
...
-            points_A, points_B, method=cv2.RANSAC, ransacReprojThreshold=100
-        )
+        B_to_A, mask = cv2.findHomography(points_A, points_B)
+        # NOTE: cv2.findHomography(A, B) will not produce the inverse of
+        # cv2.findHomography(B, A)! The errors can actually be quite large, resulting in
+        # on-screen discrepancies of up to 50 pixels. We try to find the inverse
+        # analytically instead with fallbacks.
+
+        try:
+            A_to_B = np.linalg.inv(B_to_A)
```
papr: I wonder how often we hit the pinv condition, and how fast both options are. Maybe it is sufficient to always calculate pinv?

pfaion: I tried around with weird shapes for the outline and I actually never hit pinv but could always just use inv.
Looking at the code in numpy briefly, pinv seems to do something more complicated, but I have no idea where we are there in terms of performance. My feeling is that inv should be better than pinv though.

PR #1694, opened 2019-10-22
Scale heatmap export based on surface size

papr:
Example outputs

1x1:
heatmap_Surface 1 copy

40x10:
heatmap_Surface 1 copy 2

40x100:
heatmap_Surface 1

### DIFF
--- a/pupil_src/shared_modules/surface_tracker/background_tasks.py
+++ b/pupil_src/shared_modules/surface_tracker/background_tasks.py
@@ -443,10 +443,17 @@ class Exporter:
     def _export_surface_heatmap(self, surface, surface_name):
         if surface.within_surface_heatmap is not None:
             logger.info("Saved Heatmap as .png file.")
-            cv2.imwrite(
-                os.path.join(self.metrics_dir, "heatmap" + surface_name + ".png"),
-                surface.within_surface_heatmap,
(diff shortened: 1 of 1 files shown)


pfaion:
Works. One remark though:

I noticed though that the width and height determine the export size now in pixels. This seems unintuitive given that this is supposed to be the real world size (at least that's what it says in the code as well).

So if my surface is 40cm x 60cm I'd enter 40 x 60 but the resulting export will be extremely small. Basically with a resolution of 1px/cm. I find this confusing.

Maybe we can just state in the UI that these are pixel values?

For what are they used besides the export? I noticed they determine the size of the surface window as well (which is also pixel).

pfaion:
I talked to @marc-tonsen about this and we arrived at the conclusion that this is probably the best way to handle this for now.

The reasoning was that the only issue that users appear to be having is how to specify the output image size for the heatmap. While this was possible before via the smoothness, it was unintuitive. This PR should fix this issue for now.

In addition there's maybe a more general confusion about the parameters in the surface UI. If you look at it in detail, real-world-size should probably be different from output size and the correlations between smoothness, output resolution and overlay image resolution could be a lot clearer. But these things should be easy to understand for users for now. If we have more problems with the UX of the surface UI we should rethink and rework the UI from scratch the next time and come up with a good interaction model.

PR #1696, opened 2019-10-22
Fix unstable surface outlines

pfaion:
When defining a new surface, markers from a few frames before and after are aggregated to ensure more stable marker mappings.

This is a problem when the surface is defined in a section where the markers flicker a lot. The bounding boxes are not mapped appropriately and thus the surface definition contains erroneous marker locations. This error will bleed over into the whole mapping. Most importantly it can result in jumping boundaries when only a subset of the markers is visible of which a majority was incorrectly mapped at definition.

The quick fix is here to just don't scan neighboring markers on surface definition. Especially with the new apriltags there is no need to do temporal smoothing anymore.

@papr In the long run we should refactor the surface tracker to get rid of all the code supporting the marker aggregates. This makes sense because there's a lot of complexity only for this use case and we should get rid of it if we are not intending to use it. Since this is a bit more work, I'd say we do this in a separate refactor.

### DIFF
--- a/pupil_src/shared_modules/surface_tracker/surface.py
+++ b/pupil_src/shared_modules/surface_tracker/surface.py
@@ -94,6 +94,9 @@ class Surface(abc.ABC):
         self.surf_to_dist_img_trans = None
         self.num_detected_markers = 0
 
+        # TODO: The aggregation over neighbors for REQUIRED_OBS_PER_MARKER can lead to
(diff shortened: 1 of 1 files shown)


pfaion:
@papr Also tested in capture: very stable surfaces, even when I'm far away and they are flickering like crazy when adding a surface. It stays exactly where it is.

PR #1719, opened 2019-11-01
Fix corrupted gaze data from latest PI recordings

pfaion:
This is important because there was an additional debug file introduced to PI recordings, called `gaze_right ps*_timestamps.npy` which (currently) only contains zeros.

With the previous pattern we intended to match all gaze files for multipart recordings, but now we would include these erroneous zeros, leading to jumps in the gaze data. Specifically these files were corrupted:
* `gaze.pldata`
* `gaze_timestamps.npy`

**So currently when opening a normal single-part PI recording, every second gaze point will be (0,0)!**

Even worse: the gaze timestamps are only generated once upon upgrading of the PI recording to newstyle. This cache is now incorrect and we will have to create an update mechanism for all PI recordings that have been opened once with Player < v1.18.

## To Do:
- [x] Fix corrupted gaze data cache from conversion.

### DIFF
--- a/pupil_src/shared_modules/pupil_recording/info/__init__.py
+++ b/pupil_src/shared_modules/pupil_recording/info/__init__.py
@@ -12,5 +12,7 @@ See COPYING and COPYING.LESSER for license details.
 from .. import Version
 from .recording_info import RecordingInfoFile
 from .recording_info_2_0 import _RecordingInfoFile_2_0
+from .recording_info_2_1 import _RecordingInfoFile_2_1
 
 RecordingInfoFile.register_child_class(Version("2.0"), _RecordingInfoFile_2_0)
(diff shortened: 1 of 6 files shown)


pfaion:
@papr I pushes a draft for upgrading the recordings and adding a new version.
Most notably we don't have to change the min_player_version, but need a way of differentiating between pre and post v1.18. Therefore I bumped the meta_version to 2.1.

Currently this is not yet working. as I found no (working) way of actually bumping the meta info for a recording. But this can serve as a basis for discussing whether this is the right direction at all.

pfaion:
Can be tested by:
1. Make a PI recording with current production.
1. Open in Player v1.16 or v1.17.
1. There should be something broken with the gaze data.
1. Open in in branch.
1. It should work again.

PR #1728, opened 2019-11-11
Remove upsampling functionality from binocular gaze mapper

papr:
This resulted in some pupil datums being mapped multiple times. One can verify the issue by subscribing to binocular gaze `gaze.3d.01.` and measuring the time differences between sequential gaze datums.


papr:
We have evaluated several gaze matching algorithms and have decided to go for #1731.

You can find the evaluation here: https://nbviewer.jupyter.org/github/pupil-labs/pupil-matching-evaluation/blob/master/gaze_mapper_evaluation.ipynb

PR #1731, opened 2019-11-13
Temporal cutoffs in binocular gaze mapper based on estimated FPS

romanroibu:
This PR changes the binocular gaze mapper to use a temporal cutoff based on the estimated FPS of the pupil data stream.

### DIFF
--- a/pupil_src/shared_modules/calibration_routines/gaze_mappers.py
+++ b/pupil_src/shared_modules/calibration_routines/gaze_mappers.py
@@ -79,13 +79,30 @@ class Binocular_Gaze_Mapper_Base(Gaze_Mapping_Plugin):
 
         self.min_pupil_confidence = 0.6
         self._caches = (deque(), deque())
-        self.default_temportal_cutoff = 1/120
+        self.temporal_cutoff = 1/120
+        self.temporal_cutoff_smoothing_ratio = 1/50
         self.sample_cutoff = 10
 
-    def calculate_temporal_cutoff(cache):
-        if len(cache) < 2:
-            return self.default_temportal_cutoff
+    def is_cache_valid(self, cache):
+        return len(cache) >= 2
+
+    def calculate_raw_temporal_cutoff(self, cache):
         return np.mean(np.diff([d['timestamp'] for d in cache]))
+
+    def calculate_temporal_cutoff(self, eye0_cache, eye1_cache):
+        if self.is_cache_valid(eye0_cache) and self.is_cache_valid(eye1_cache):
+            eye0_raw_cutoff = self.calculate_raw_temporal_cutoff(eye0_cache)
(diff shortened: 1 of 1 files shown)


pupil_src/shared_modules/calibration_routines/gaze_mappers.py
```diff
@@ -79,9 +79,14 @@ def __init__(self, g_pool):
 
         self.min_pupil_confidence = 0.6
         self._caches = (deque(), deque())
-        self.temportal_cutoff = 0.3
+        self.default_temportal_cutoff = 0.3
```
papr: Default is 120 Hz
```suggestion
        self.default_temportal_cutoff = 1/120
```

pupil_src/shared_modules/calibration_routines/gaze_mappers.py
```diff
@@ -79,9 +79,14 @@ def __init__(self, g_pool):
...
         self.sample_cutoff = 10
 
+    def calculate_temporal_cutoff(cache):
+        if len(cache) < 2:
+            return self.default_temportal_cutoff
+        return np.mean(np.diff([d['timestamp'] for d in cache]))
+
```
pfaion: This means we aim at only matching frames that lie as close together as our average fps, right?
Shouldn't we at least leave a bit room, like maybe 2* the avg?
Because otherwise a single low-confidence value means already that we don't map binocular, right?
I feel like we actually want to do some fitting to other close samples if there's some low-conf involved.
But please correct me, I'm not in the loop on that dicussion.

pupil_src/shared_modules/calibration_routines/gaze_mappers.py
```diff
@@ -94,6 +116,7 @@ def map_batch(self, pupil_list):
 
     def on_pupil_datum(self, p):
         self._caches[p["id"]].append(p)
+        temporal_cutoff = 2 * self.calculate_temporal_cutoff(*self._caches)
```
pfaion: @romanroibu 
`calculate_temporal_cutoff` stores the result in `self.temporal_cutoff`, but in the end we are using 2* the result.
Why do we have this difference between `self.temporal_cutoff` and the actual cutoff being used in the end?

papr: I agree, the variable and function should be renamed to `self.recently_estimated_framerate` and `estimate_framerate()` respectively, and then `temporal_cutoff = 2 * estimate_framerate(...)`

pfaion: That makes a lot more sense.

papr: On it

Issue #1691, opened 2019-10-21
Get/Set for 2d Detection Properties does not work 


elsmann:
When trying to set or broadcast the 2d pupil detection properties the eye process crashes. 

2019-10-21 10:22:09,184 - eye1 - [ERROR] launchables.eye: Process Eye1 crashed with trace:
Traceback (most recent call last):
  File "launchables/eye.py", line 650, in eye
  File "detector_2d.pyx", line 294, in pupil_detectors.detector_2d.Detector_2D.get_detector_properties
AttributeError: 'pupil_detectors.detector_2d.Detector_2D' object has no attribute 'detectProperties2D'

pfaion:
Hi @elsmann 
which version are you using? Are you running from bundle or from source?

pfaion:
Also can you share some code on how exactly you tried to set or broadcast the properties?

elsmann:
Hi, thanks for the quick answer. I'm running from bundle, Version 1.16.80


Im simply using 

```
def notify(notification):
    """Sends ``notification`` to Pupil Remote"""
    topic = 'notify.' + notification['subject']
    payload = serializer.dumps(notification, use_bin_type=True)
    socket.send_string(topic, flags=zmq.SNDMORE)
    socket.send(payload)
    return socket.recv_string()


def recv_sub():
    """Recv a message with topic, payload."""
    topic = sub.recv_string()
    payload_serialized = sub.recv()
    payload = serializer.loads(payload_serialized)
    return topic, `payload`

```
and 

```
notify({"subject": "pupil_detector.broadcast_properties.1"})
topic, properties = recv_sub()
print(topic, properties)

notify({
     "subject": "pupil_detector.set_property.2d",
     "name": "pupil_size_min",
     "value": 10,
 })

```

which works if detection and mapping mode is set to "3d". 
If I manually set detection mode to "2d" I get the above mentioned Error.

pfaion:
Ok this appears to be a bug.
How important is this for your current workflow?

We are working on a small refactor of the detector API anyways at the moment. If this is not critical we would probably ignore this as it will be fixed with the new API anyways. Otherwise we might be able to look at it a bit more, as it appears to be only a minor issue. Or you can also investigate it yourself and create a PR if you are very eager 😃

elsmann:
It would be great if you could have a look at it, if you find the time! 
I do need the 2d detection for my next steps, but my coding skills are rather limited😅. 
[...]

PR #1742, opened 2019-11-19
Fix bug #1691

Christopher22:
This pull request fixes bug #1691. It is probably one of the simplest bugfixes in history while the bug itself is sometimes quite annoying...


pfaion:
@Christopher22 Thank you very much for your contribution (even if small!)

PR #1743, opened 2019-11-19
Fixation detector fixes

papr:
Please see commit messages for reference:
- https://github.com/pupil-labs/pupil/commit/515783161abdf0165ab1898ae2ac83b992916251
- https://github.com/pupil-labs/pupil/commit/f9720e10a512625e4b1e161f0f498b79b90257ec
- https://github.com/pupil-labs/pupil/commit/e84101b20f20b97e4a35d3d5c4e639b08a3ccf42

### DIFF
--- a/pupil_src/shared_modules/fixation_detector.py
+++ b/pupil_src/shared_modules/fixation_detector.py
@@ -10,17 +10,20 @@ See COPYING and COPYING.LESSER for license details.
 """
 
 """
-Fixations general knowledge from literature review + Goldberg et al. - fixations rarely
-    < 100ms and range between 200ms and 400ms in duration (Irwin, 1992 - fixations
-    dependent on task between 150ms - 600ms) + Very short fixations are considered not
-    meaningful for studying behavior - eye+brain require time for info to be registered
-    (see Munn et al. APGV, 2008) + Fixations are rarely longer than 800ms in duration +
-    Smooth Pursuit is exception and different motif + If we do not set a maximum
-    duration, we will also detect smooth pursuit (which is acceptable since we
-    compensate for VOR) Terms + dispersion (spatial) = how much spatial movement is
-    allowed within one fixation (in visual angular degrees or pixels) + duration
-    (temporal) = what is the minimum time required for gaze data to be within dispersion
-    threshold?
+Fixations general knowledge from literature review
+    + Goldberg et al. - fixations rarely < 100ms and range between 200ms and 400ms in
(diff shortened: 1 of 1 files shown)


pupil_src/shared_modules/fixation_detector.py
```diff
@@ -158,41 +142,47 @@ def gaze_dispersion(capture, gaze_subset, use_pupil=True):
...
 def detect_fixations(
     capture, gaze_data, max_dispersion, min_duration, max_duration, min_data_confidence
 ):
     yield "Detecting fixations...", ()
-    gaze_data = [
+    gaze_data = (
         fm.Serialized_Dict(msgpack_bytes=serialized) for serialized in gaze_data
+    )
+    gaze_data = [
+        datum for datum in gaze_data if datum["confidence"] > min_data_confidence
```
pfaion: This will result potentially in fixations that have low-confidence spots inbetween, if the rest of the gaze data has appropriate dispersion, correct?

Is this wanted? Can the rest of the application deal with this? As in: what does the UI do if we have a fixation with low-confidence gaze-points inbetween?

pupil_src/shared_modules/fixation_detector.py
```diff
@@ -158,41 +142,47 @@ def gaze_dispersion(capture, gaze_subset, use_pupil=True):
...
+    method = (
+        FixationDetectionMethod.GAZE_3D
+        if can_use_3d_gaze_mapping(gaze_data)
+        else FixationDetectionMethod.GAZE_2D
     )
+    logger.info(f"Starting fixation detection using {method.value} data...")
     fixation_result = Fixation_Result_Factory()
 
     Q = deque()
     enum = deque(gaze_data)
```
pfaion: Actually since we are making this nice currently, can we rename the deques (especially the `enum`)?
`Q` is more like `working_queue`
and `enum` should be something along the lines of `remaining_gaze_data`.

pfaion:
@papr If you approve of the changes I could work them in tomorrow if you want me to.

pupil_src/shared_modules/fixation_detector.py
papr: To specify: @pfaion wanted to know if low confidence data should be a termination criterion for fixation detection, i.e. is a fixation valid if there was low confidence data during the duration of the fixation?

Response: Low confidence data does not provide reliable spatial data, therefore it should not be considered at all. A fixation will only be detected if all high-confidence gaze points during the fixation fulfill the minimum-duration-maximum-dispersion criterion.

PR #1750, opened 2019-12-03
Add explicit link for downloading app bundles

willpatera:
There have been a few users who are not familiar with GitHub having difficulty finding out where to download the latest application bundles. 

I propose to add an image here with a "button" within the image that links to releases/latest. For sake of redundancy I also propose to leave the text link "download the latest apps" - but I am open to suggestions of removing this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,9 +31,9 @@ target="_blank">
 </a>
 
 
-[Download the latest apps](https://github.com/pupil-labs/pupil/releases/latest "Download Pupil Capture, Pupil Player, and Pupil Service application bundles")! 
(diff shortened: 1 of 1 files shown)


pfaion:
I think another big problem is that all these link to the release notes. I had multiple people asking on discord where they can find the actual bundles. For the latest release I already changed the intro sentence for the release notes from
> Download the latest bundle and let us know what you think

to
> Download the latest bundle (at the end of the release notes under **Assets**) and let us know what you think

Maybe we can put an HTML-anchor at the end of the release notes and include a button at the top to jump down? Not sure how well that would work on GitHub markdown though.

PR #1751, opened 2019-12-05
Correctly handle empty video files

papr:
Fixes `av_frame is referenced before assignment` crash in file backend if video is actually empty but was not recognized as such.


pfaion:
@papr is this intended to go into **master**?

papr:
I was planning on including it as a bug fix update to v1.19. But given that v1.20 is about to be released, we can merge it into `develop`, too.

PR #1754, opened 2019-12-09
Add link anchors to download section in release notes

pfaion:
So users are directed directly to the bottom of the release notes.


pfaion:
Note that I PR'ed into **master**.

PR #1755, opened 2019-12-09
Support PI remote recording control

papr:
Logs first error message line as error, rest as debug message.

TODO:
- [ ] PI version check


papr:
This feature won't be supported officially in Pupil Capture and therefore this PR will be closed without merging.

PR #1763, opened 2019-12-12
Update timestamps from seconds to nanoseconds

romanroibu:


papr:
This project has been deemed infeasible for now.

PR #1767, opened 2019-12-16
Fix fullscreen calibration

pfaion:
This PR fixes a couple of bugs with the screen marker calibration window:

- On Windows the window won't be black on open anymore
- Capture won't crash anymore (or throw errors) when tabbing out of the calibration
- The calibration will stop when the window is closed/minimized/tabbed out
- The cursor will be captured in fullscreen mode. This reduces the risk of accidentally tabbing out when having more than 1 monitor.

@papr please test all kinds of shenanigans with the calibration window on macOS

### DIFF
--- a/pupil_src/shared_modules/calibration_routines/screen_marker_calibration.py
+++ b/pupil_src/shared_modules/calibration_routines/screen_marker_calibration.py
@@ -427,10 +427,11 @@ class Screen_Marker_Calibration(Calibration_Plugin):
 
         p_window_size = glfwGetFramebufferSize(self._window)
         if p_window_size == (0, 0):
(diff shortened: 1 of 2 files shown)


pfaion:
I still need to test this on Linux.

pfaion:
For some reason the eye windows crash beyond recovery when tabbing out of fullscreen calibration on Linux.

pfaion:
@romanroibu I tried crashing the calibration Window again on my Ubuntu but could not break it. Can you try again if tabbing out while running the fullscreen calibration crashes on your machine (with this branch)?

pfaion:
I tested on Linux and macOS and could not find any problems.
@romanroibu can you have a last look over the code if you see anything suspicious?

PR #1769, opened 2019-12-17
add an option to limit FPS for online pupil detection for performance

bpinsard:
We have a video camera that capture frame at up to 1KHz FPS, and the most important for us is to dump the video at the highest FPS to the disk to allow offline (re)analysis.
In order to improve the performance and avoid losing frames, I quickly implemented an option to subsample the frames sent to the online pupil detection.
As such, we still have an online feedback on the ability of the algorithm to properly detect the pupil with the current camera setup, but at a lower CPU cost.
This is only for the 2D detector, but it could easily be done for the 3D one and factor it in the base plugin class. 
Pushing it as it might be of interest for other users.
Tell me what you think.


pfaion:
Hi @bpinsard 
thank you for your contribution to Pupil!
This is indeed a good solution for your use case. However, we have decided not to include this into Pupil as it would mean exposing more UI that is not relevant to probably 99% of our users. I'll close this PR but we will use it as a reference for the remaining 1%. 🙂

PR #1775, opened 2019-12-19
Update Python Dependencies for Ubuntu 18

moiri:
In order to run pupil capture from source some additional dependencies had to be installed.

### DIFF
--- a/docs/dependencies-ubuntu18.md
+++ b/docs/dependencies-ubuntu18.md
@@ -52,6 +52,9 @@ sudo udevadm trigger
 We recommend using a virtual environment with a valid installation of Python 3.6 or higher.
 
 ```sh
+pip install wheel
+pip install scikit-build
+
 pip install cysignals
 pip install cython
 pip install msgpack==0.5.6
@@ -71,6 +74,9 @@ pip install git+https://github.com/pupil-labs/pyndsi
 pip install git+https://github.com/pupil-labs/pyglui
 pip install git+https://github.com/pupil-labs/nslr
 pip install git+https://github.com/pupil-labs/nslr-hmm
+
+pip install opencv-python
+pip install pupil_detectors
 ```
 
 ## Next Steps


pfaion:
Hi @moiri 
thank you for contributing to Pupil!

There is already an open PR for adapting the developer docs for the new pupil-detectors package: #1775 
Sorry for being late on this.

Regarding the other packages:

**wheel and scikit-build**
`wheel` and `scikit-build` should not have to be added, as they are no dependencies of Pupil. Rather these are dependencies used by `pupil-apriltags` and `pupil-detectors`, but these packages should know which dependencies to install themselves. If you had problems with these dependencies, my guess is that you are using an old version of pip. Please try running `python -m pip install --upgrade pip` to update your pip to the newest version. It would be great if you could check if that resolves your problem with `wheel` and `scikit-build`.

**opencv-python**
There is no need to install `opencv-python` if your opencv is correctly setup. Since running `pip install opencv-python` will install another full (potentially different) version of opencv to your machine, we are not recommending this setup. When you install opencv from `sudo apt install -y python3-opencv libopencv-dev` as we advise in the setup docs, you should have the `cv2` package available for import in Python as this will also install python bindings that are compatible with the apt version of opencv. There actually is a note [in the docs for Ubuntu <=17](https://github.com/pupil-labs/pupil/blob/master/docs/dependencies-ubuntu17.md#opencv-troubleshooting) which I guess we should include in the Ubuntu 18 docs as well!
If you are unable to import `cv2` despite this setup, it could be that you are using a Python virtual environment, which can cause problems with apt-installed Python bindings.
See e.g. [step 4 of this stackoverflow post](https://stackoverflow.com/a/37190408).

I guess takeaways for us are:
- Add a note to upgrade pip before installing python dependencies
- Add a note about opencv-python in the Ubuntu 18 docs
- Add a note on opencv Python bindings in virtual environments to the Ubuntu 17 and 18 docs (potentially also macOS... @papr does this apply there?)

Thank you for bringing these issues to our attention! I'll leave this PR open until you can confirm that updating pip solves the issue with `wheel` and `scikit-build`.

pfaion:
In order to make this more foolproof when installing with old pip versions, we should put a pip upgrade command into the install instructions on all platforms.

PR #1784, opened 2020-01-07
Virtualenv docs

pfaion:
I added a unified note that virtualenvs are recommended. Reasons:
- e.g. PIMonitor has conflicting dependencies with Pupil
- when using a virtualenv you can just run `pip` everywhere instead of having to decide between `pip` and `pip3`
- you don't need `sudo pip` when installing into a virtualenv

Since virtualenvs have trouble with `apt` installed Python packages on Ubuntu, I extended the notes on troubleshooting the OpenCV Python bindings. Also those were missing in the Ubuntu 18 docs. See this PR: #1775 for an example of how this can confuse users.

@papr does the trouble with system-installed opencv-python-bindings in virtualenvs also apply to macOS? Then we should add a note there as well. I found the [`--system-site-packages`](https://virtualenv.pypa.io/en/latest/userguide/#the-system-site-packages-option) option a generally good thing to mention in that regard.

### DIFF
--- a/docs/dependencies-macos.md
+++ b/docs/dependencies-macos.md
@@ -103,7 +103,7 @@ When you see this error, Python cannot find the bindings from your OpenCV instal
 Installing `opencv-python` will install another full (potentially different) version of opencv to your machine, so we are not recommending this setup.
 When you install opencv with `brew install opencv` as we advise above, you should have the `cv2` package available for import in Python as this will install compatible Python bindings already.
 
-However, you might run into these problems when using a virtual environment, as your virtual environment cannot by default access Python packages that were installed from `apt`.
+However, you might run into these problems when using a virtual environment, as your virtual environment cannot by default access Python packages that were installed from `brew`.
 In that case there are 2 options:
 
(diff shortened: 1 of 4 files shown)


pfaion:
I created a draft until we clarify the macOS situation (@papr)

pfaion:
I'll add additional notes to the macOS docs on 'No module named cv2' as well!

docs/dependencies-macos.md
```diff
@@ -93,6 +93,22 @@ pip install git+https://github.com/pupil-labs/nslr-hmm
...
+`ImportError: No module named 'cv2'`
+  
+When you see this error, Python cannot find the bindings from your OpenCV installation.
+
+**We do NOT (!) recommend to install `opencv-python` via pip in that case!** 
+
+Installing `opencv-python` will install another full (potentially different) version of opencv to your machine, so we are not recommending this setup.
+When you install opencv with `brew install opencv` as we advise above, you should have the `cv2` package available for import in Python as this will install compatible Python bindings already.
+
+However, you might run into these problems when using a virtual environment, as your virtual environment cannot by default access Python packages that were installed from `apt`.
```
romanroibu: `brew`, not `apt`

PR #1785, opened 2020-01-08
Scan Path plugin

romanroibu:
##### TODO

- [x] fix loading the cache from disk
    - there seems to be an issue between loading the cache and the cache being immediately invalidated because the controller is getting a `gaze_changed` notification soon after.
- [x] fix issue related to `ipc_pub` that crashes the app on macOS
- [x] move UI from `ScanPathPlugin` into `Vis_Polyline`
- [x] correctly handle multi-part videos
    - when receiving a "gap frame", the algorithm should reset.


pupil_src/shared_modules/scan_path/scan_path_background_task.py
```diff
@@ -0,0 +1,113 @@
...
+        yield progress, frame.index, gaze_datums, corrected_gaze_datums
+
+
+def generate_frames_with_gaze(g_pool):
+
+    video_path = os.path.join(g_pool.rec_dir, "world.mp4") #TODO: Use PupilRecording
+
+    fs = File_Source(g_pool, source_path=video_path)
+    
+    gp = GazeFromRecording(g_pool)
```
papr: Gaze from recording is not the only source for gaze. At the end we might need to pass gaze directly to the background process, similar to the fixation detector.

papr:
Depends on https://github.com/pupil-labs/pupil/pull/1795

papr:
Decision on time frame setting persistence? By recording or by Player session?

PR #1788, opened 2020-01-09
Roi plugin

pfaion:
This PR cleans up the detector ROI code and moves it to a plugin.

While moving the stuff over, I found a ton (!) of bugs, unhandled cornercases and UI inconsistencies that were not handled correctly. The new ROI should be a lot more stable and provide better UX.

**Rationale**
1. Get rid of duplicate (plugin) logic for roi in eye.py. I also found bugs in this.
2. Allow correct ordering of the ROI processing (after backend, before detectors). This is important because we had some frame drops and hack because the old ROI was processed after the detectors (which were plugins).
3. Simplify confusing roi properties. The roi grew over time and just had added properties. In the end there were (for x direction): `lx, ux, nx, minx, maxx, frame_width` where you only need `minx, maxx, frame_width`.

**API Changes**
I changed the network format for setting the ROI from
`(minx, maxx, miny, maxy)` to
`(minx, MINY, MAXX, maxy)`
This is the same format that the old roi used internally, the new roi uses internally and the detectors use internally. This will make the code more consistent, but we need to inform users that the API changed! I added improved error logging for setting the roi over the network to provide better feedback when this causes problems.


pupil_src/shared_modules/roi.py
```diff
@@ -0,0 +1,405 @@
...
+    Some notes on behavior:
+    - Modifying bounds will always confine to the frame size and keep and area of >= 1.
+    - Changing the frame size will scale the bounds to the same relative area.
+    - If any frame dimension is <= 0, the ROI becomes invalid.
+    - Setting the frame size of an invalid ROI to a valid size re-initializes the ROI.
+    """
+
+    def __init__(self, frame_size: Vec2) -> None:
+        """Create a new RoiModel with bounds set to the full frame."""
+        self._change_callbacks: T.List[ChangeCallback] = []
```
papr: Please make `RoiModel` an `Observable`.

pupil_src/shared_modules/roi.py
```diff
@@ -0,0 +1,405 @@
...
+    def _set_to_full_frame(self, frame_size: Vec2) -> None:
+        """Initialize to full frame for given frame_size."""
+        width, height = (int(v) for v in frame_size)
+        self._frame_width = width
+        self._frame_height = height
+        self._minx = 0
+        self._miny = 0
+        self._maxx = width - 1
+        self._maxy = height - 1
+        self._changed()
```
papr: Rename to `on_changed()`

pfaion:
@papr wasn't sure if that's how it is intended. Works like a charm though :smile:
https://github.com/pupil-labs/pupil/pull/1788/commits/22260fa5dd37f74887c4ff6968808c5455c6a8a5#diff-73ad68db2be4a4e52820b99b4471c76eR156-R161

PR #1792, opened 2020-01-21
Video capture slimming

### LINKED ISSUE #1672: Start with default devices: Source plugin not changed
Bug encountered on Windows, capture version v1.16-80.

Consistent way to reproduce (might not be minimal):
- Connect headset to mobile phone and run Pupil Mobile
- Open Capture
- Select Pupil Mobile backend and **Start with default devices**
- Everything should be working as expected, all cameras show feed from Pupil Mobile
- Select Local USB as backend and **Start with default devices**

Now the manager is set to UVC, but the source stays at NDSI.
When now connecting the headset via USB, the cameras are not found.
Hitting **Start with default devices** again also does not change the source and the cams stay off!
The connection can only be recovered by e.g. selecting Pupil Mobile backend again, then switching back to Local USB again and hitten **Start with default devices** once more.

See NDSI source with Local USB backend:
image

pfaion:
# Video Capture Slimming

This PR simplifies the capture selection in Pupil Capture significantly.
Most important user-facing changes:
- Realsense and Fake backends have been deprecated
- File backend is not selectable anymore, but works by drag-and-dropping
- There is now only a single plugin for managing the capture source, showing:
  - Capture selection dropdown
  - Info/Settings for currently running capture
- There are now two activation modes:
  - Auto: activate an input device across all processes, best matching cameras will be selected automatically
  - Manual: activate specific cameras separately for every process
- The 'Ghost Mode' was renamed to 'Camera disconnected' or similar
- The capture plugin UI has been removed from Pupil Player

Closes #1672 

## To-do checklist:
- [x] Deprecate FakeSource
- [x] Deprecate RealSense
- [x] Plugin name should stay constant "Video Source" or similar, independent of source instance
- [x] UI switch between auto and manual mode
- [x] Manual/auto toggle is broadcasted
- [x] Mode should be stored in settings
- [x] Drop down for activating devices/ cameras
- [x] Drop shows devices in auto mode
- [x] Drop shows cameras in manual mode
- [x] All Managers will be loaded
- [x] Managers register themselves in g_pool
- [x] Managers offer list of devices / list of cameras
- [x] Device selection in auto mode is broadcasted
- [x] Device selection in auto mode resolves appropriate cams
- [x] Fix pyre log floods with permanent ndsi backend
- [x] File source is useable by drag-dropping
- [x] Group source instance settings
- [x] All source settings are grouped under a single plugin menu
- [x] Dropdown menu UI without label?
- [x] Managers offer activation API
- [x] Managers do not have UI
- [x] Cleanup old auto selection and backend selection
- [x] Rename ghost-mode to disconnected
- [x] Think about time_sync warning? Maybe in NDSI source?
- [x] Add no devices found to dropdown.
- [x] Fix ui repaint with UVC disconnect
- [x] ~Cleanup recover logic in NDSI backend~
- [x] Cleanup remaining references for realsense (ctrl-shift-f)
- [x] Cleanup old comments/docstrings (already partially outdated)
- [x] Cleanup imports
- [x] Cleanup TODOs in code
- [x] Test player
- [x] Decide whether to hide source UI completely in player
- [x] Catch ndsi bug "does not support data subscription"
- [x] Fix broken mode selector in file source
- [x] Fix broken session persistence
- [x] HMD source works as before
- [x] Add in final changes from UI/UX feedback


pfaion:
Do not merge yet as we have to do a final pass over the UI/UX.

pupil_src/launchables/eye.py
```diff
@@ -213,19 +213,21 @@ def get_timestamp():
...
-        default_capture_settings = (
-            "UVC_Source",
-            {
-                "preferred_names": preferred_names,
-                "frame_size": (320, 240),
-                "frame_rate": 120,
-            },
-        )
+
+        default_capture = "UVC_Source"
```
romanroibu: `<nitpick>`
I'd suggest `default_capture_name` to be more explicit, and symmetry. Feel free to ignore this suggestion :)
`</nitpick>`

pupil_src/launchables/world.py
```diff
@@ -288,6 +288,7 @@ def get_timestamp():
         ]
         g_pool.plugin_by_name = {p.__name__: p for p in plugins}
 
+        default_capture = "UVC_Source"
```
romanroibu: ditto

pupil_src/shared_modules/video_capture/base_backend.py
```diff
@@ -63,22 +71,150 @@ class Base_Source(Plugin):
...
+    def init_ui(self):
+        self.add_menu()
+        self.menu.label = "Video Source"
+        self.update_menu()
+
+    def deinit_ui(self):
+        self.remove_menu()
+
+    def source_list(self):
+        source_type = "Camera" if self.manual_mode else "Device"
```
romanroibu: What's the difference between `"Camera"` and `"Device"`? :thinking:
Also, is it just for UI text?

pfaion: A device offers a collections of cameras.
In manual mode you select cameras, in auto mode you select devices.

That line is just for UI, correct.

pupil_src/launchables/eye.py
pfaion: Makes total sense!

pupil_src/shared_modules/video_capture/base_backend.py
romanroibu: thanks for the clarification!

pfaion:
@romanroibu @papr please test the following features:
- with capture: connect a local USB headset and a PM phone
  - activate one and then the other in device-mode
  - activate cameras individually
  - close and reopen the eye windows
  - close capture and check the opened cameras will be restored correctly
- just for sanity: run offline pupil detection in player

romanroibu:
@pfaion I was trying out this branch and I noticed something that is a bit confusing:

- when I open `Capture`, in the `Video Source` menu, the selected option in the dropdown is `Activate Device`.
- when I open the dropdown, I see 2 options (`Activate Device` and `Local USB`)
- when I try to select `Local USB`, I see error logs saying `"The selected camera is already in use or blocked."`

I think it's better to only show `Activate Device` if nothing is currently selected. If there is an active camera, just show that in the dropdown without the `Activate Device` option. Otherwise, I think it's confusing to use.

Let me know what you think.

papr:
This is actually intended behavior.
1) Given the pyglui implementation, you need an entry that does nothing such that the user can close the menu without overwriting the video source that is running already.
2) Managing stateful selector is tricky if the devie disconnects.
3) Attempting to open an already active UVC device will always result in the "in use or blocked" message due to how one source replaces the other.

I agree that this could be improved from an UX perspective but unfortunately, we have to work around technical issues here.

papr:
It works like a charm on macOS. 👍

pfaion:
@romanroibu I completely agree. That's also what I suggested in the beginning. Not so easy apparently.

PR #1795, opened 2020-01-27
Use Announcer and Listener for gaze data change

romanroibu:
##### TODO

- [x] Provide an optional `token` to `Announcer.announce_new`; this would allow to announce "new" data, that _might_ be the same as the previously announced data (e.g. running the gaze mapping on the same pupil data with the same configurations).


papr:
Depends on https://github.com/pupil-labs/pupil/pull/1795

PR #1810, opened 2020-02-17
Eyeball Outline for Player

pfaion:
This PR adds the green eyeball outline visualization to the following **Player** plugins:
- eye overlay
- eye video export

The existing switch for toggling pupil ellipse visualization has been adjusted to include the eyeball outline as well.

When 3D data is not available, we only show the pupil ellipse just as before.


papr:
Tested successfully on macOS

jonmatthis:
Could you please adjust this to make it possible to show the red pupil outline WITHOUT the green 3d overlay?

PR #1811, opened 2020-02-18
Drop non-monotonic frames instead of crashing

pfaion:
Fixes that Pupil crashes when writing a video and encountering non-monotonic timestamps.

Instead of raising, we log this case.


pfaion:
@papr I was just thinking. Should we log a warning instead of debug?
Since this is rather something that we would want users to report to us?

papr:
@pfaion The frame drop should actually happen in `UVC Source`. Each video source is responsible for providing frames with monotonic timestamps. In the originally reported issue, the issue was that the libuvc time correction generated negative timestamps. Therefore, it is the uvc source's responsibility to drop these.

PR #1812, opened 2020-02-18
Drop frames with non-monotonic timestamps in uvc

pfaion:
Sometimes UVC might provide us with non-monotonic timestamps.
If we don't drop these frames, we might crash somewhere else.


pupil_src/shared_modules/video_capture/uvc_backend.py
```diff
@@ -423,6 +423,17 @@ def recent_events(self, events):
                 # c930 timestamps need to be set here. The camera does not provide valid pts from device
                 frame.timestamp = uvc.get_time_monotonic() + self.ts_offset
             frame.timestamp -= self.g_pool.timebase.value
+
+            if (
+                self._recent_frame is not None
+                and frame.timestamp < self._recent_frame.timestamp
```
papr:
```suggestion
                and frame.timestamp <= self._recent_frame.timestamp
```

PR #1828, opened 2020-03-17
Remove eye_movement detector and Windows C++ docs

pfaion:
I noticed that the only thing we need visual studio for on Windows was `nslr`, which did not have prebuilt wheels. Since we deprecated the eye_movement detector, we could actually get rid of nslr.

- removed all **eye_movement** code and replaced cross-module used code
- removed **nslr** as dependency for all platforms
- cleaned up Windows docs:
  - we don't need a hard-coded **C:\work** anymore!
  - we don't need **Visual Studio**
  - we don't need the **developer command prompt**, simple `cmd.exe` is enough
  - we don't need **OpenCV** binaries anymore in pupil_external
  - I removed the notes on Windows paths, since users don't need to adjust paths anymore
  - I also removed old docs for installing **optimization_calibration** and the therefore needed **ceres** install instructions


docs/dependencies-windows.md
```diff
@@ -129,34 +87,23 @@ Open your command prompt and `Run as administrator` in the directory where the w
...
+Then run:
 ```sh
 git clone https://github.com/pupil-labs/pupil.git
 ```
 
 ### Include pupil_external in PATH Variable
 
 - Follow the instructions under the System Environment Variables section above to add a new environment variable to PATH
-- Add the following folder: `C:\work\pupil\pupil_external`
+- Add the full path the `pupil_external` folder of the repository that you just cloned, e.g. `C:\work\pupil\pupil_external`
```
papr:
```suggestion
- Add the full path to the `pupil_external` folder of the repository that you just cloned, e.g. `C:\work\pupil\pupil_external`
```

PR #1829, opened 2020-03-19
Stop nonmonotonic recording

pfaion:
When timestamps are handled from external sources, the sources need to take care of syncing time. This can lead to jumps in timestamps when syncing while recording.

Since the video writer classes work under the assumption of monotonic data, we cannot handle negative time jumps caused by syncing. Previously Capture would just crash when receiving a negative time jump.

This PR handles negative time jumps more gracefully:
- A running recording will be stopped, so the partial recording is useable
- An error message will be shown, more detailed information will be logged as debug info
- Capture does not crash anymore

### DIFF
--- a/pupil_src/shared_modules/av_writer.py
+++ b/pupil_src/shared_modules/av_writer.py
@@ -89,7 +89,7 @@ def write_timestamps(file_loc, timestamps, output_format="npy"):
         )
 
 
-class NonMonotonicError(ValueError):
+class NonMonotonicTimestampError(ValueError):
     """Indicates that a Writer received non-monotonic data to write."""
 
     pass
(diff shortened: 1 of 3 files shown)


pfaion:
@papr please give this a try. I tested this once by pausing the debugger while recording and manually messing with the timestamps in memory, then resumed the debugger. Everything worked as excepted: recording stopped and was useable in Player. I could continue to start another recording.

pfaion:
@papr As a side note: we should also extract the recording info from `eye.py` into a recorder plugin at some time, currently this is all still patched together in the eye main loop.

pupil_src/shared_modules/av_writer.py
```diff
@@ -89,6 +89,12 @@ def write_timestamps(file_loc, timestamps, output_format="npy"):
         )
 
 
+class NonMonotonicError(ValueError):
```
papr:
```suggestion
class NonMonotonicTimestampError(ValueError):
```

pfaion: Done

PR #1831, opened 2020-03-21
fixed #1830

### LINKED ISSUE #1830: ValueError: int is not allowed for map key when using msgpack 1.0.0
**Bug:**
"ValueError: int is not allowed for map key" will be raised when using msgpack 1.0.0 and you want to load data recorded with two eye cameras.

**Explanation:**
This is to the fact that they changed their default setting for msgpack.unpackb see: https://github.com/msgpack/msgpack-python/issues/411 which they also discuss as major braking points in https://github.com/msgpack/msgpack-python#major-breaking-changes-in-msgpack-10

**Solution:** 
adding `strict_map_key=False` as argument whenever `msgpack.unpackb` or `msgpack.loads` is called

sven-mayer:
this will address the problem described in #1830  by adding strict_map_key=False to the function call.

As msgpack prior to 1.0.0 used this as default value this will keep backward compatibility but now also ensures that msgpack >= 1.0.0  works as they put the default parameter to True.

### DIFF
--- a/pupil_src/shared_modules/file_methods.py
+++ b/pupil_src/shared_modules/file_methods.py
@@ -237,6 +237,7 @@ class Serialized_Dict(object):
                 self._ser_data,
                 raw=False,
                 use_list=False,
+                strict_map_key=False,
                 object_hook=self.unpacking_object_hook,
                 ext_hook=self.unpacking_ext_hook,
             )


papr:
Thank you for providing a solution to #1830!

Currently, this project requires [`msgpack==0.5.6`](https://github.com/pupil-labs/pupil/blob/bcfbf7d2b03bfbae14788d431fd47fe334e9fe4b/docs/dependencies-macos.md#install-python-libraries). To my knowledge, the `strict_map_key` was introduced in `0.6.0` and is therefore not compatible with our required version.

I will close this PR for now. But as soon as we upgrade our msgpack requirement, we can merge your contribution.

PR #1832, opened 2020-03-23
Use custom error if video does not have frames

papr:
The old implementation raises a TypeError as `av.AVError` expects an additional error code. Instead of passing any error code, the new implementation raises (and catches) a custom error in the case that a video is valid but does not contain any frames.

### DIFF
--- a/pupil_src/shared_modules/video_capture/utils.py
+++ b/pupil_src/shared_modules/video_capture/utils.py
@@ -25,10 +25,6 @@ VIDEO_EXTS = ("mp4", "mjpeg", "h264", "mkv", "avi", "fake")
 VIDEO_TIME_EXTS = VIDEO_EXTS + ("time",)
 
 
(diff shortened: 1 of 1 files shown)


pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -189,8 +193,8 @@ def check_valid(self):
             # 3. decode() yields None
             first_frame = next(cont.decode(video=0), None)
             if first_frame is None:
-                raise av.AVError("Video does not contain any frames")
-        except av.AVError:
+                raise VideoDoesNotContainFramesError()
+        except (av.AVError, VideoDoesNotContainFramesError):
             return False
```
pfaion: Do we use the error somewhere else? Because we immediately catch it, I thought we could just:
```python
if first_frame is None:
    # No frames
    return False
````

PR #1833, opened 2020-03-24
Fix windows path issues

pfaion:
This PR fixes multiple issues on Windows with Unicode characters and spaces in the path to Pupil.

Most notably `libwdi` cannot handle non-ASCII characters in the path to unpack and install drivers to/from. This was bypassed by checking the working directory for ASCII compatibility and using `C:\Windows\Temp` (specifically, to avoid non-ASCII characters in potential default TEMP directory) as fallback location.

By using a new temporary folder for every driver, potential permission errors that occurred when cleaning up folders created by elevated commands have also been prevented. A frequently occurring error due to unescaped spaces in the old cleanup routine was thereby also fixed.

### DIFF
--- a/pupil_src/shared_modules/video_capture/uvc_backend.py
+++ b/pupil_src/shared_modules/video_capture/uvc_backend.py
@@ -190,16 +190,16 @@ class UVC_Source(Base_Source):
                 temp_path = Path("C:\\Windows\\Temp")
                 if not temp_path.exists():
                     raise RuntimeError(
-                        "Username contains Unicode characters and default Temp folder "
-                        "location could not be found. Please use a different username "
-                        "which only uses english characters!"
+                        "Your path to Pupil contains Unicode characters and the "
+                        "default Temp folder location could not be found. Please place "
+                        "Pupil on a path which only contains english characters!"
                     )
                 logger.debug(
                     "Detected Unicode characters in working directory! "
                     "Switching temporary driver install location to C:\\Windows\\Temp"
                 )
             else:
-                # if pwd has only ascii characters: use default temp location
+                # if cwd has only ascii characters: use default temp location
                 temp_path = None
 
             for id in ids_to_install:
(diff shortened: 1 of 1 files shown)


pupil_src/shared_modules/video_capture/uvc_backend.py
```diff
@@ -203,29 +203,43 @@ def verify_drivers(self):
...
+                except PermissionError as e:
+                    # This can be raised when cleaning up the TemporaryDirectory, if the
+                    # process was started from a non-admin shell for a non-admin user
+                    # and has only been elevated for the powershell commands. The files
+                    # then belong to a different user and cannot be deleted. We can
+                    # ignore this, as temp dirs will be cleaned up on shutdown anyways.
+                    logger.warning(
+                        "Received a PermissionError while trying to install drivers. "
+                        "If the driver do not work, please try running from an "
+                        "administrator shell again!"
```
pfaion: I just realized this can be made more user-facing... I'll adjust this soon.

pupil_src/shared_modules/video_capture/uvc_backend.py
```diff
@@ -175,35 +177,56 @@ def verify_drivers(self):
...
-                cmd_str += (
-                    f'''Start-Process PupilDrvInst.exe -Wait -Verb runas'''
-                    f''' -WorkingDirectory "{work_dir}"'''
-                    f''' -ArgumentList '--vid {id[0]} --pid {id[1]} --desc "{id[2]}" --vendor "Pupil Labs" --inst'; '''
+                str(Path.cwd().resolve()).encode("ascii")
+            except UnicodeEncodeError:
+                temp_path = Path("C:\\Windows\\Temp")
+                if not temp_path.exists():
+                    raise RuntimeError(
+                        "Username contains Unicode characters and default Temp folder "
```
pfaion: This is also technically not correct. It could also be the path to the bundle, not only the username. I'll adjust this as well.

PR #1835, opened 2020-03-25
Improve multi-part recording handling in Player

papr:
- c386e11 - Fixes a crash in offline pupil detection caused by an invalid video part
- 6480e7f - Loads video parts on-demand instead of pre-loading them. This decreases the time that Pupil Player needs to open a recording and avoids a crash due to too many open files in case of recordings with a lot of video parts (long Pupil Mobile recordings).


pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -174,34 +174,33 @@ def __init__(self, path: str) -> None:
...
-            self.cont = cont
-            return True
+            return cont
 
-    def load_valid_container(self):
-        if self.is_valid:
-            return self.cont
-        else:
-            return None
+    def attempt_open_container(self):
```
pfaion: Since this can raise an `AVError`, the naming is a bit weird. You have to "try attempt open" to use this.
Why not also raise an `InvalidContainerError` (see comment above) and rename this to `open_container()`. Then you can:
```python
try:
  vid.open_container()
except InvalidContainerError:
  # ...
```
Also, since you already hide away the `av.open()` in a separate function (good call btw), this would also hide the `AVError` with a more appropriately-named error.

pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -174,34 +174,33 @@ def __init__(self, path: str) -> None:
...
         self.ts = None
         self._pts = None
-        self._is_valid = self.check_valid()
+        self._is_valid = None  # calculated on demand
 
     @property
     def is_valid(self):
+        if self._is_valid is None:
+            self._is_valid = self.load_valid_container() is not None
         return self._is_valid
```
pfaion: Read the other comments first!
...
This becomes then e.g.
```python
def is_valid(self):
    # check cached state
    if self._is_valid is None:
        try:
            self.load_container()
        except InvalidContainerError:
            self._is_valid = False
        else:
            self.is_valid = True
    return self._is_valid
```
which might be a bit more verbose, but reads so much clearer in what's going on here.

pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -211,10 +210,13 @@ def load_ts(self):
             return
         self.ts = self._fix_negative_time_jumps(self.ts)
 
-    def load_pts(self):
-        packets = self.cont.demux(video=0)
+    def load_pts(self, container=None):
+        if container is None:
+            container = self.attempt_open_container()
```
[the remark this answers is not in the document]
papr: This function is actually only called, if the container is valid: https://github.com/pupil-labs/pupil/pull/1835/files#diff-84845973e9fa00c663d23ca5d1efa86cR358

Nonetheless, we should not be leaking the `AVError`. Instead I would raise `InvalidContainerError` as suggested below.

pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -174,34 +174,33 @@ def __init__(self, path: str) -> None:
...
+        self._is_valid = None  # calculated on demand
 
     @property
     def is_valid(self):
+        if self._is_valid is None:
+            self._is_valid = self.load_valid_container() is not None
         return self._is_valid
 
-    def check_valid(self):
+    def load_valid_container(self):
```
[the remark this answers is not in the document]
papr: I agree. In this case, we can simply merge `load_container()` and `attempt_open_container()`

pupil_src/shared_modules/video_capture/utils.py
papr: Agreed

pupil_src/shared_modules/video_capture/utils.py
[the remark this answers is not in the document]
pfaion: I think you could mitigate this by always requiring a container.
Then the container has to be valid when passing this in.
Since this function seems to be called only once in a context where we pass a valid container, correct?

pupil_src/shared_modules/video_capture/utils.py
papr: I agree, although I would move the `self._is_valid = True` into the try-block

pfaion: Totally fine with that.

pupil_src/shared_modules/video_capture/utils.py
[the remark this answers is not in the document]
papr: agreed

PR #1845, opened 2020-04-09
Remove manual libuvc copy for macOS bundles

papr:
libuvc and its dependencies should automatically be detected when importing pyuvc during bundling.

Reverts bd3c72a81b9bde464c5ff45e54d636ecf3ffdcb5


pfaion:
This commit is not the exact negative of the commit you said it would revert (https://github.com/pupil-labs/pupil/commit/bd3c72a81b9bde464c5ff45e54d636ecf3ffdcb5)

Is this intended?

papr:
Yes. The line in `deployment/deploy_capture/bundle.spec` has been in there but does not belong there.

PR #1850, opened 2020-04-17
Decouple dual detectors

pfaion:
This PR modifies the detector plugins in order to allow running 2D and 3D detectors in parallel, without having the 2D detector twice.

This branch only work together with https://github.com/pupil-labs/pupil-detectors/pull/9

### DIFF
--- a/pupil_src/shared_modules/pupil_detector_plugins/detector_2d_plugin.py
+++ b/pupil_src/shared_modules/pupil_detector_plugins/detector_2d_plugin.py
@@ -38,6 +38,7 @@ class Detector2DPlugin(PupilDetectorPlugin):
 
     label = "C++ 2d detector"
     identifier = "2d"
+    order = 0.100
 
     def __init__(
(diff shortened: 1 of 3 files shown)


pfaion:
@papr I'd bump the version on master after merging, so we can just run bump2version (which also creates a tag, that I'd rather have on master)

pfaion:
@papr on it.

pfaion:
@papr is this what you had in mind? I added the version check to the 3D plugin only, since the changes only concern this one.

I did not use `assert` since I wanted to also log an error for the user. Since this is only in the eye process, the world window will continue to run and you would only see an eye process crash otherwise. Now you get an error telling you that pupil_detectors is outdated.

pfaion:
@papr should we merge this? This would make the `restructured-gaze-calibration-and-mapping` branch incompatible with the current pip version of pupil_detectors. Maybe we should delay this until you had the time to deploy v1.0.5?

papr:
I will make a release now

PR #1855, opened 2020-04-23
Gazer 2d v2

pfaion:
Add the new version of the 2D Gazer


pupil_src/shared_modules/gaze_mapping/gazer_2d_v2.py
```diff
@@ -0,0 +1,246 @@
...
+    @staticmethod
+    def _validate_feature_dimensionality(X):
+        assert X.ndim == 2, "Required shape: (n_samples, n_features)"
+        assert X.shape[1] == _MONOCULAR_FEATURE_COUNT, (
+            f"Received shape: {X.shape}. "
+            f"Expected shape (n_samples, {_MONOCULAR_FEATURE_COUNT})"
+        )
+
+
+class Gazer2D(GazerBase):
```
romanroibu, line 162: can we rename this to `Gazer2D_v2x` to be consistent with `Gazer2D_v1x`? also, the module should follow the `v1` pattern and be called `gazer_2d_v2x`.

To be clear, I don't prefer the convention of `v1` over any other convention; just want things to be named consistently.

pfaion:
I also realized we now have a new dependency: scikit-learn!
This is why the travis fails.
I'll update the travis dependency setup and the install docs as well.

pupil_src/shared_modules/gaze_mapping/gazer_2d_v2.py
pfaion: @papr what are your thoughts on this? I understood that we will get rid of the v1 anyways? Do we want to keep this versioning?

papr: We will remove the `_vYx` suffix before the release. Specifically, we will remove `Gazer2D_v1x` completely, and change `Gazer3D_v1x` to `Gazer3D`. I just need to make another pass through the code before committing to the final version.

pfaion: Ok so I see this as resolved

pfaion: No wait. Currently the filename is `gazer_2d_v2.py`. Do we also want to change that to `gazer_2d.py` @papr ?

pfaion: I just did this. If that's not ok, pls comment.

papr: Correct call

PR #1858, opened 2020-04-28
Remove Gazer2D_v1x

romanroibu:


pupil_src/shared_modules/calibration_choreography/natural_feature_plugin.py
```diff
@@ -27,7 +27,8 @@
     ChoreographyMode,
     ChoreographyAction,
 )
-from gaze_mapping import Gazer2D_v1x, Gazer3D_v1x
+from gaze_mapping import Gazer2D
+from gaze_mapping import Gazer3D_v1x
```
pfaion, line 31: Why are these even in here? They're not being used, are they?
Same below in `screen_marker_plugin.py`

pupil_src/shared_modules/gaze_producer/model/legacy/calibration_v1.py
```diff
@@ -63,7 +63,13 @@ def updated_name(self) -> str:
     def updated(self) -> Calibration:
 
         gazer_class, gazer_params = _gazer_class_and_params_from_gaze_mapper_result(self.result)
-        is_offline_calibration = gazer_params is not None
+
+        if gazer_params is not None:
+            is_offline_calibration = True
+            status = self.status
+        else:
+            is_offline_calibration = True
```
papr, line 71: The purpose of this flag is to differentiate between recorded calibrations (read-only) and offline calibrations (editable). This change makes all calibrations editable.

Recorded 2d calibrations should potentially be refit based on the "calibrationsetup" data that _should_ be included in the recording. This is not guaranteed if the calibration was created in a separate recording. The correct solution would be to find a way to upgrade the 2d params from the v1x version.

Nonetheless, given the current implementation, I would rather not upgrade/support 2d recorded calibrations than show them as editable calibrations.

papr: In other words, we would drop the support for using recorded 2d calibrations in offline calibrations. This affects people who started the recording _after_ the performing a 2d calibration procedure and want to apply a manual offset correction to their recorded data.

Issue #1284, opened 2018-08-30
Pupil Capture crash


schtiefel:
Lately *Capture* sometimes crashes a few seconds after starting — it has not yet happened at a later time.

Console output including trace looks like this:
```
MainProcess - [INFO] os_utils: Disabling idle sleep not supported on this OS version.
world - [INFO] launchables.world: Application Version: 1.8.26
world - [INFO] launchables.world: System Info: User: ###, Platform: Windows, Machine: ###, Release: 10, Version: 10.0.17134
world - [INFO] plugin: Added: <class 'plugin.Analysis_Plugin_Base'>
world - [INFO] plugin: Added: <class 'blink_detection_fixed.Blink_Detection_Fixed'>
world - [WARNING] launchables.world: Process started.
Running  PupilDrvInst.exe --vid 1443 --pid 37424
OPT: VID number 1443
OPT: PID number 37424
...
eye0 - [INFO] camera_models: No user calibration found for camera Pupil Cam2 ID0 at resolution (400, 400)
eye0 - [INFO] camera_models: No pre-recorded calibration available
eye0 - [WARNING] camera_models: Loading dummy calibration
Running  PupilDrvInst.exe --vid 1443 --pid 37424
OPT: VID number 1443
OPT: PID number 37424
...
eye1 - [INFO] camera_models: No user calibration found for camera Pupil Cam2 ID1 at resolution (400, 400)
eye1 - [INFO] camera_models: No pre-recorded calibration available
eye1 - [WARNING] camera_models: Loading dummy calibration
eye1 - [WARNING] uvc: Could not set Value. 'Backlight Compensation'.
eye0 - [WARNING] launchables.eye: Process started.
Estimated / selected altsetting bandwith : 617 / 800.
!!!!Packets per transfer = 32 frameInterval = 83263
eye1 - [WARNING] launchables.eye: Process started.
Estimated / selected altsetting bandwith : 617 / 800.
!!!!Packets per transfer = 32 frameInterval = 83263
world - [ERROR] launchables.world: Process Capture crashed with trace:
Traceback (most recent call last):
  File "launchables\world.py", line 482, in world
  File "OpenGL\error.py", line 232, in glCheckError
OpenGL.error.GLError: GLError(
        err = 1281,
        description = b'ung\xfcltiger Wert',
        baseOperation = glViewport,
        cArguments = (0, 0, 1330, 720)
)
```

Windsooon:
[...]

PR #1870, opened 2020-05-12
Default detectors

papr:
Please give this PR a test, once with a connected headset and once without.


papr:
@pfaion In the context of this PR could you please also have a look at https://github.com/pupil-labs/pupil/issues/1284

The issue is caused if there is no scene image but pupil data that is recognized to be a blink.

pfaion:
@papr Seems to work fine. I noticed though that Pupil Service seems to drop all logs in the service process. I noticed since the blink detector logs blink onset/offset for Capture, but not for Player.

PR #1871, opened 2020-05-12
Allow replacing `by_class` unique plugins

papr:
Changes previous behavior where running `by_class` unique plugins could not be replaced.

This behaviour is especially problematic if `by_class` plugins are started by default. In this case, there would not be a possibility to change their configuration.

With the new behaviour, a new configuration can be loaded by restarting the plugin with the new config. The old plugin will be removed.

### DIFF
--- a/pupil_src/shared_modules/plugin.py
+++ b/pupil_src/shared_modules/plugin.py
@@ -350,13 +350,13 @@ class Plugin_List(object):
     def __str__(self):
         return "Plugin List: {}".format(self._plugins)
 
-    def add(self, new_plugin, args={}):
+    def add(self, new_plugin_cls, args={}):
         """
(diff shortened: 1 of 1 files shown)


pupil_src/shared_modules/plugin.py
```diff
@@ -385,6 +367,42 @@ def add(self, new_plugin, args={}):
         if self.g_pool.app in ("capture", "player"):
             plugin_instance.init_ui()
 
+    def _avoid_duplication(self, new_plugin_cls):
```
pfaion, line 370: I'd call this `_remove_duplicates(...)`.
`avoid` does not really tell what's happening.

pfaion:
Otherwise works fine!

PR #1873, opened 2020-05-14
Draw 2D pupil ellipse in blue and cleanup code

pfaion:
This will color the 2D ellipse visualization in blue, which helps now that we have both 2D and 3D detectors running in parallel. This change affects both Capture and Player.

I also increased the polygon count of the ellipse outline to match that of the 3D eyeball visualization for a smoother pupil. After some testing, I did not touch the thickness, so pupil ellipses are still drawn with 1px, while the eyeball outline has 2px as before.

Example:
Screenshot from 2020-05-14 13-49-08


pfaion:
@papr Player uses the same function to draw the ellipses to both eye overlay and eye video export.
So this already works.

PR #1877, opened 2020-05-15
Hide realsense cameras in camera selection list

pfaion:
Since v1.22 Pupil Capture does not offer built-in support for Intel RealSense cameras anymore. The cameras would still show up in the camera selection list, although activating them would have no effect. This would confuse users, as they could use the cameras previously.

This PR filters out all cameras with **RealSense** in the name from the camera activation list.

TODO:
- [x] Check if this really works, I don't have a RealSense available to test


papr:
Reintroduces https://github.com/pupil-labs/pupil/pull/782/commits/719b4671318efef0aef002cd37f8e9b7d4f7ff69

pfaion:
@papr I tested on the office machine:
- RealSense does not show up in UVC backend anymore!
- I also checked and the custom plugin still works!

PR #1881, opened 2020-05-20
Merge Pupil Remote and Frame Publisher plugins

romanroibu:
### Changes
- Merged the `Frame Publisher` and `Pupil Remote` plugins into a single `Network API` plugin


pupil_src/shared_modules/data_api/data_api_plugin.py
```diff
@@ -0,0 +1,95 @@
...
+        super().__init__(g_pool)
+        self.__frame_publisher = FramePublisherController(**kwargs)
+        self.__frame_publisher.add_observer("on_frame_publisher_did_stop", self.on_frame_publisher_did_stop)
+
+
+        self.__pupil_remote = PupilRemoteController(g_pool, **kwargs)
+
+        # UI
+        self.__frame_publisher_ui_menu = FramePublisherMenu(self.__frame_publisher)
+        self.__pupil_remote_ui_menu = PupilRemoteMenu()
```
papr, line 42: UI should not be initialized in `__init__()` if possible.

romanroibu: the components themselves don't call any UI function and don't create any `pyglui` instances in their `__init__`.

do you still think that initializing these objects should move to `init_ui`?

PR #1882, opened 2020-05-20
Add glGetError() check to every Plugin.gl_display()

pfaion:
We often receive OpenGL errors as results of buggy pyglui code that gets called in `gl_display()` of plugins. Since pyglui does not check `glGetError()`, we run into these errors at other places at the code when e.g. using pyopengl. By checking after every plugin's `gl_display()` call we can at least partially localize the error!

Since it's easy to skip calling the base `Plugin.gl_display()` from subclasses, I monkeypatched a wrapper into all subclasses.


pfaion:
@papr I added the check to only run this with `--debug`

PR #1888, opened 2020-05-26
Fix gaze corruption when deactivating gaze mappers

pfaion:
I had 2 offline Gaze Mappers and deactivated one. As a result there was a lot of weirdly duplicated and wrong data in the gaze export.

I did not figure out the whole chain of what was going wrong but stumbled across this piece of code. Never use mutable types as default parameter values! After fixing this the problem disappeared for me.

Since I didn't fully validate that this was indeed the cause of the problem, maybe see if you can reproduce the issue before (and not afterward).

### DIFF
--- a/pupil_src/shared_modules/gaze_producer/model/gaze_mapper.py
+++ b/pupil_src/shared_modules/gaze_producer/model/gaze_mapper.py
@@ -29,8 +29,8 @@ class GazeMapper(StorageItem):
         status="Not calculated yet",
         accuracy_result="",
         precision_result="",
-        gaze=[],
-        gaze_ts=[],
+        gaze=None,
+        gaze_ts=None,
     ):
         self.unique_id = unique_id
         self.name = name
@@ -44,8 +44,8 @@ class GazeMapper(StorageItem):
         self.status = status
         self.accuracy_result = accuracy_result
         self.precision_result = precision_result
-        self.gaze = gaze
-        self.gaze_ts = gaze_ts
+        self.gaze = gaze if gaze is not None else []
+        self.gaze_ts = gaze_ts if gaze_ts is not None else []
 
     @property
     def calculate_complete(self):


romanroibu:
wow, forgot that Python is weird in this regard (at least from what my expectations are) 🤦

romanroibu:
To avoid such issues in the future, I propose we consistently use `...` as default values for non-optional parameters (parameters that shouldn't be `None`), and replace `...` with a newly created value in the body.

pfaion:
@romanroibu I would argue against using `...`.

I think using `None` as default value is actually a very very common workflow. On the other side I haven't seen `...` being used outside of the Pupil codebase (except extended numpy slicing) and I'm always confused when I run across some random `...`. We already use it for multiple different purposes, just yesterday I saw it being used to repesent a wildcard key in some custom `__getitem__` function.

And regarding:
> I propose we consistently use `...` as default values for non-optional parameters (parameters that shouldn't be `None`),

I would argue that _non-optional_ parameters should not get a default value at all.

romanroibu:
@patrickFuerst I think I didn't explain very well what I meant, sorry for the confusion. Hopefully I can illustrate it better with an example. Consider a function with the following semantics:

```python
foo(x=100)  # trivial case
foo(x=None)  # None is a valid value for x
foo()  # x should be set to the default value, which is not None
```

pfaion:
@romanroibu ok I understand. You mean cases where `None` is a valid value and the default should be something else than `None`. I agree these are annoying cases. But generally, I feel they are pretty rare (or at least should be rare, since this means the parameter has some very very weird type).

In the case of this PR there's no issue with using `None` as default value. I still feel strongly that we shouldn't use `...` everywhere for default values. Unfortunately I couldn't find the original purpose of the `...` (it was introduced in 3.0). But the docs also state that they are normally used for extended slicing: https://docs.python.org/3/library/constants.html#Ellipsis

PR #1893, opened 2020-05-28
Integrate Gazer changes into Pupil Service

romanroibu:


pupil_src/launchables/service.py
```diff
@@ -275,17 +273,12 @@ def handle_notifications(n):
...
-                for gaze_datum in new_gaze_data:
-                    gaze_pub.send(gaze_datum)
-
                 events = {}
-                events["gaze"] = new_gaze_data
                 events["pupil"] = [pupil_datum]
                 for plugin in g_pool.plugins:
                     plugin.recent_events(events=events)
+                for gaze_datum in events.get("gaze", []):
+                    gaze_pub.send(gaze_datum)
```
papr, line 266: Please revert this hunk. I decided to keep the existing gaze mapping + publishing logic in #1891 as we would be trading a higher gaze publishing delay for a bit more plugin-like behavior. This trade-off is not worth it in my opinion.

PR #1895, opened 2020-05-29
Allow to set ZMQ HWM for Eye Streams

moiri:
Allow to change the socket option HWM for the eye data streams

- during initialization by extending the notification `eye_process.should_start` with the field `hwm`
- at runtime with a GUI text input (this will cause a ZMQ reconnection)
- at runtime with the notification `eye_stream.set_zmq_option.hwm` (this will cause a ZMQ reconnection)


papr:
@moiri Hi, could you elaborate on the use case of these features? In which cases do you think is it necessary to change the zmq hwm from its default value to a custom one? And is it necessary to be able to do so dynamically? Or do you think it would be possible to hardcode a specific hwm that fits better than the default one?

moiri:
@papr I have a scenario where I always need the latest data sample from the device. I prioritize latency over throughput, hence I don't want to fill up the buffer with "old" data which I first have to process before getting to the latest data sample. Hence, I would like to set the HWM to 1 (minimum). This may result in the loss of some data samples which is ok as long as I always get the latest sample.

Being able to set this dynamically is not necessary. For my purpose it would be sufficient to set it during initialization (i.e. with the notification `eye_process.should_start` and the field `hwm`).

papr:
I am not sure if setting the HWM to 1 does what you intend. Looking at https://rfc.zeromq.org/spec/29/#the-pub-socket-type I see 
> For processing outgoing messages:
> [...]
> - SHALL silently drop the message if the queue for a subscriber is full.

The queue in your case would have a length of 1 and would be full most of the time. If I understand the document correctly, you would be dropping newer data instead of replacing old data in the sending queue. This is not what you intended by setting the HWM to 1, correct?

moiri:
You are correct, but I have not found any way to overwrite the existing data in the queue. I can deal with on old data sample way better than with 1000 old data samples (which is the default). So HWM = 1 is good enough.

pupil_src/launchables/eye.py
```diff
@@ -63,6 +63,7 @@ def eye(
     eye_id,
     overwrite_cap_settings=None,
     hide_ui=False,
+    hwm=None,
```
papr, line 66:
```suggestion
    pub_socket_hwm=None,
```

pupil_src/launchables/eye.py
```diff
@@ -95,7 +96,7 @@ def eye(
 
     zmq_ctx = zmq.Context()
     ipc_socket = zmq_tools.Msg_Dispatcher(zmq_ctx, ipc_push_url)
-    pupil_socket = zmq_tools.Msg_Streamer(zmq_ctx, ipc_pub_url)
+    pupil_socket = zmq_tools.Msg_Streamer(zmq_ctx, ipc_pub_url, hwm)
```
papr, line 99:
```suggestion
    pupil_socket = zmq_tools.Msg_Streamer(zmq_ctx, ipc_pub_url, pub_socket_hwm)
```

pupil_src/launchables/eye.py
```diff
@@ -376,6 +377,13 @@ def set_window_size():
             glfw.glfwSetWindowSize(main_window, f_width, f_height)
 
         general_settings.append(ui.Button("Reset window size", set_window_size))
+
+        g_pool.hwm = pupil_socket.get_hwm()
+        def update_hwm(new_hwm):
+            g_pool.hwm = new_hwm
+            pupil_socket.set_hwm(new_hwm)
+
+        general_settings.append(ui.Text_Input("hwm", g_pool, setter=update_hwm, label="ZMQ High Water Mark"))
```
papr, line 386:
```suggestion
```

pupil_src/launchables/eye.py
```diff
@@ -533,6 +541,9 @@ def window_should_update():
                         )
                     except KeyError as err:
                         logger.error(f"Attempt to load unknown plugin: {err}")
+                elif subject.startswith("eye_stream.set_zmq_option.hwm"):
+                    if notification["eye_id"] == eye_id:
+                        update_hwm(notification['hwm'])
```
papr, line 546:
```suggestion
```

pupil_src/launchables/eye.py
```diff
@@ -704,6 +715,7 @@ def eye_profiled(
     eye_id,
     overwrite_cap_settings=None,
     hide_ui=False,
+    hwm=None,
```
papr, line 718:
```suggestion
    pub_socket_hwm=None,
```

pupil_src/launchables/eye.py
```diff
@@ -723,6 +735,7 @@ def eye_profiled(
             "eye_id": eye_id,
             "overwrite_cap_settings": overwrite_cap_settings,
             "hide_ui": hide_ui,
+            "hwm": hwm,
```
papr, line 738:
```suggestion
            "pub_socket_hwm": pub_socket_hwm,
```

pupil_src/main.py
```diff
@@ -282,6 +282,7 @@ def log_loop(ipc_sub_url, log_level_debug):
                 topic, n = cmd_sub.recv()
                 if "notify.eye_process.should_start" in topic:
                     eye_id = n["eye_id"]
+                    hwm = n["hwm"] if "hwm" in n else None
```
papr, line 285:
```suggestion
                    pub_socket_hwm = n.get("pub_socket_hwm")
```

pupil_src/main.py
```diff
@@ -296,6 +297,7 @@ def log_loop(ipc_sub_url, log_level_debug):
                             eye_id,
                             n.get("overwrite_cap_settings"),
                             parsed_args.hide_ui,
+                            hwm,
```
papr, line 300:
```suggestion
                            pub_socket_hwm,
```

pupil_src/shared_modules/zmq_tools.py
```diff
@@ -169,6 +173,21 @@ def send(self, payload, deprecated=()):
...
+
+    def set_hwm(self, hwm):
+        """
+        Disconnect the socket, set a new High Water Mark (HWM) value and
+        reconnect the socket.
+        """
+        self.socket.disconnect(self.url)
+        self.socket.set_hwm(hwm)
+        self.socket.connect(self.url)
+
```
papr, line 190:
```suggestion
```

pupil_src/shared_modules/zmq_tools.py
```diff
@@ -134,8 +134,12 @@ class Msg_Streamer(ZMQ_Socket):
     Not threadsave. Make a new one for each thread
     """
 
-    def __init__(self, ctx, url):
+    def __init__(self, ctx, url, hwm=None):
+        self.url = url
```
papr, line 138: There is one more suggestion.
```suggestion
```

moiri:
Thanks for the suggestions and the fast replies!
I tested the changes and it works fine.

papr:
@moiri Gratulations to your first Pupil project source code contribution! Thank you very much!

PR #1899, opened 2020-06-02
Improve Offline Pupil Detection data storage during detection

papr:
Instead of using `Mutable_Bisector`s (which use `nump.insert`), use a timestamp->datum mapping. This has the advantage of being faster and avoids duplicated entries when restarting the detection while there is an existing detection running.

This PR also correctly announces cached pupil data as "existing" instead of "new". This avoids recalculating e.g. offline gaze mappings on restoring a session.

### DIFF
--- a/pupil_src/shared_modules/player_methods.py
+++ b/pupil_src/shared_modules/player_methods.py
@@ -323,6 +323,35 @@ class PupilDataBisector:
         return data_by_topic
 
 
+class PupilDataCollector:
+    def __init__(self):
+        self._collection = collections.defaultdict(dict)
+
+    def append(self, topic, datum, timestamp):
+        pupil_topic = PupilTopic.create(topic, datum)
+        self._collection[pupil_topic][timestamp] = datum
+
+    def clear(self):
+        self._collection.clear()
+
+    def as_pupil_data_bisector(self) -> PupilDataBisector:
+        bisectors = {}
+        for topic, timestamps_data in self._collection.items():
+            timestamps = list(timestamps_data.keys())
+            data = list(timestamps_data.values())
+            bisector = Bisector(data, timestamps)
+            bisectors[topic] = bisector
+        pupil_data_bisector = PupilDataBisector(bisectors=bisectors)
+        return pupil_data_bisector
+
+    def count_collected(self, eye_id=None, detector_tag=None):
+        num_collected = 0
+        for topic, values in self._collection.items():
+            if PupilTopic.match(topic, eye_id, detector_tag):
+                num_collected += len(values)
+        return num_collected
+
+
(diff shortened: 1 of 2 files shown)


pupil_src/shared_modules/player_methods.py
```diff
@@ -48,7 +48,7 @@ def __init__(self, data=(), data_ts=()):
                 )
             )
         elif not len(data):
-            self.data = []
+            self.data = np.asarray([])
             self.data_ts = np.asarray([])
```
pfaion, line 52: Is there a specific reason for why we use `asarray` here?
Looks like we could also use `np.array([])` or even `np.empty(0)`

papr: It is mostly used to mirror the other `asarray` calls below. I guess using `np.empty` would be easier to read. What do you think? I am open to changing these two lines.

pfaion: I think I have seen `np.array([])` in multiple places throughout the codebase. Maybe that's a bit more readable than `np.empty(0)`. Too bad there's no `np.empty()` :smile:

pfaion:
@papr Did you find a reliable way of reproducing the crash with `insert`?

papr:
The issue was caused when the Bisector was initialized with data (which was transformed into a ndarray) and then data was inserted during the pupil redetection. I think this was the case when restoring cached offline pupil data and then hitting recalculate.

pupil_src/shared_modules/pupil_producers.py
```diff
@@ -311,29 +313,24 @@ def __init__(self, g_pool):
...
-
         self.eye_frame_num = [0, 0]
-        self.eye_frame_num[0] = len(self._pupil_data_store[0, "3d"]) #TODO: Figure out the number of frames independent of 3d detection
-        self.eye_frame_num[1] = len(self._pupil_data_store[1, "3d"]) #TODO: Figure out the number of frames independent of 3d detection
-
-        self.pause_switch = None
-        self.detection_paused = False
+        # TODO: Figure out the number of frames independent of cached 3d detection
+        # self.eye_frame_num[0] = len(pupil_data_from_cache[0, "3d"])
+        # self.eye_frame_num[1] = len(pupil_data_from_cache[1, "3d"])
```
pfaion, line 327: dead code

papr:
```suggestion
```

pupil_src/shared_modules/player_methods.py
papr, line 52:
```suggestion
            self.data = np.array([])
            self.data_ts = np.array([])
```

PR #1902, opened 2020-06-03
Small improvements for Pupil Core v2 (2)

romanroibu:
### Changes
- Removed `"Choreography"` from calibration plugins' labels
- Added custom labels for calibration plugins in selection dropdown
- Added custom order for calibration plugins in selection dropdown
- Renamed the selection dropdown label from `"Gazer"` to `"Method"`
- Added custom order for single marker modes in selection dropdown
- Added automatic hiding of marker scale for single marker calibration in `"Physical Marker"` mode

### DIFF
--- a/pupil_src/shared_modules/calibration_choreography/base_plugin.py
+++ b/pupil_src/shared_modules/calibration_choreography/base_plugin.py
@@ -360,6 +360,7 @@ class CalibrationChoreographyPlugin(Plugin):
         self.menu.append(self.__ui_selector_choreography)
         self.menu.append(self.__ui_selector_gazer)
         self.menu_icon.order = self.order
+        self.menu_icon.tooltip = "Calibration"
 
         if self.shows_action_buttons:
 
--- a/pupil_src/shared_modules/calibration_choreography/base_plugin.py
+++ b/pupil_src/shared_modules/calibration_choreography/base_plugin.py
@@ -337,7 +337,17 @@ class CalibrationChoreographyPlugin(Plugin):
         # even choreographies that subclass concrete choreography implementations.
         return CalibrationChoreographyPlugin
 
+    def _gazer_description_text(self) -> str:
+        return ""
+
+    def _init_custom_menu_ui_elements(self) -> list:
+        return []
+
     def init_ui(self):
+
+        desc_text = ui.Info_Text(
+            self._gazer_description_text()
+        )
 
         self.__ui_selector_choreography = ui.Selector(
             "selected_choreography_class",
@@ -354,13 +364,25 @@ class CalibrationChoreographyPlugin(Plugin):
             selection=self.user_selectable_gazer_classes(),
         )
 
+        custom_ui_elements = self._init_custom_menu_ui_elements()
+
         super().init_ui()
         self.add_menu()
         self.menu.label = self.label
-        self.menu.append(self.__ui_selector_choreography)
-        self.menu.append(self.__ui_selector_gazer)
         self.menu_icon.order = self.order
         self.menu_icon.tooltip = "Calibration"
+
+        # Construct menu UI
+        self.menu.append(self.__ui_selector_choreography)
+        self.menu.append(desc_text)
+        if len(custom_ui_elements) > 0:
+            self.menu.append(ui.Separator())
+            for ui_elem in custom_ui_elements:
+                self.menu.append(ui_elem)
+            self.menu.append(ui.Separator())
+        else:
+            self.menu.append(ui.Separator())
+        self.menu.append(self.__ui_selector_gazer)
 
         if self.shows_action_buttons:
 
--- a/pupil_src/shared_modules/calibration_choreography/base_plugin.py
+++ b/pupil_src/shared_modules/calibration_choreography/base_plugin.py
@@ -264,6 +264,7 @@ class CalibrationChoreographyPlugin(Plugin):
                 )
             cls = default_cls
         self.__selected_gazer_class = cls
+        self._update_gazer_description_ui_text()
 
     @property
     def status_text(self) -> str:
@@ -344,6 +345,13 @@ class CalibrationChoreographyPlugin(Plugin):
     def _init_custom_menu_ui_elements(self) -> list:
         return []
 
+    def _update_gazer_description_ui_text(self):
+        try:
+            ui_text = self.__ui_gazer_description_text
+        except AttributeError:
+            return
+        ui_text.text = self.selected_gazer_class._gazer_description_text()
+
     def init_ui(self):
 
         desc_text = ui.Info_Text(
@@ -365,6 +373,11 @@ class CalibrationChoreographyPlugin(Plugin):
             selection=self.user_selectable_gazer_classes(),
         )
 
+        self.__ui_gazer_description_text = ui.Info_Text("")
+        self._update_gazer_description_ui_text()
+
+        best_practices_text = ui.Info_Text("Read more about best practices at docs.pupil-labs.com")
+
         custom_ui_elements = self._init_custom_menu_ui_elements()
 
         super().init_ui()
@@ -384,6 +397,8 @@ class CalibrationChoreographyPlugin(Plugin):
         else:
             self.menu.append(ui.Separator())
         self.menu.append(self.__ui_selector_gazer)
+        self.menu.append(self.__ui_gazer_description_text)
+        self.menu.append(best_practices_text)
 
         if self.shows_action_buttons:
 
--- a/pupil_src/shared_modules/calibration_choreography/base_plugin.py
+++ b/pupil_src/shared_modules/calibration_choreography/base_plugin.py
@@ -337,7 +337,7 @@ class CalibrationChoreographyPlugin(Plugin):
         # even choreographies that subclass concrete choreography implementations.
         return CalibrationChoreographyPlugin
 
-    def _gazer_description_text(self) -> str:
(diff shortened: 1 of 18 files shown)


pupil_src/shared_modules/calibration_choreography/base_plugin.py
```diff
@@ -123,6 +123,14 @@ class CalibrationChoreographyPlugin(Plugin):
...
     is_session_persistent = True
 
+    @classmethod
+    def selection_label(cls) -> str:
+        return self.label
+
+    @classmethod
+    def selection_order(cls) -> float:
+        return float("inf")
+
```
pfaion, line 133: Why are these methods? They don't _do_ anything. Couldn't it just be a class variable?

pupil_src/shared_modules/calibration_choreography/single_marker_plugin.py
```diff
@@ -56,7 +56,9 @@ class SingleMarkerMode(enum.Enum):
 
     @staticmethod
     def all_modes() -> T.List["SingleMarkerMode"]:
-        return list(SingleMarkerMode)
+        modes = list(SingleMarkerMode)
+        modes = sorted(modes, key=lambda m: m.order)
+        return list(modes)
```
pfaion, line 61: `sorted()` already returns a list.

Maybe we could even
```python
return sorted(SingleMarkerMode, key=lambda m: m.order)
```
And when I think about it more: why not just define them in the right order?
Then you can also remove the `order()` property below.
Then you can also remove this method here and just iterate the enum in the places where needed.

pupil_src/shared_modules/calibration_choreography/single_marker_plugin.py
```diff
@@ -161,9 +191,28 @@ def init_ui(self):
...
+        self.menu.append(self.__ui_selector_monitor_name)
         self.menu.append(self.__ui_slider_marker_scale)
+        self._ui_update_visibility_digital_marker_config()
+
+    def _ui_update_visibility_digital_marker_config(self):
+        try:
+            if self.__ui_slider_marker_scale is None or self.menu is None:
+                return
+        except AttributeError:
+            return
```
pfaion, line 204: When can any of this happen? Being `None` or causing AttributeError?
Especially `__ui_slider_marker_scale` seems to be never set to `None` anywhere.

pupil_src/shared_modules/calibration_choreography/single_marker_plugin.py
papr: I like that this is implemented explicitly. I would not want the UI order to be dependent on the enum definition order.

pfaion: Ok I guess that's a question of taste :slightly_smiling_face: 
Still, my first argument about `sorted` holds.

pupil_src/shared_modules/calibration_choreography/single_marker_plugin.py
papr: The `marker_mode` property calls `_ui_update_visibility_digital_marker_config()`. It is possible that for-what-ever-reason `init_ui()` was not called (valid if there was no ui). In this case you would not want `_ui_update_visibility_digital_marker_config()` to raise an exception

pupil_src/shared_modules/calibration_choreography/single_marker_plugin.py
papr, line 61: So the both suggestions are:

1. Explicit, doing one thing at a time
```suggestion
        modes = list(SingleMarkerMode)
        modes = sorted(modes, key=lambda m: m.order)
        return modes
```

2. one-liner
```suggestion
        return sorted(SingleMarkerMode, key=lambda m: m.order)
```

pupil_src/shared_modules/calibration_choreography/base_plugin.py
papr: So I guess this change is technically possible
```suggestion
    selection_label: str = label
    selection_order: float = float("inf")
```
but the major difference would be the order of execution. If you want `selection_label` to fall back to the subclasses' `label` class attribute by default, then you need this to be a class method.

pfaion: Agreed, but we don't have a subclass without a `selection_label`. This looks like functionality we don't actually use.

pfaion: Also now that I look at it, the `label` in the base class is set to `None`, which isn't even a string. So it would actually break if you don't set a `label` in the subclass.

papr: Yes, you have to set a `label` in the subclass. But using the `selection_label` classmethod implementation above, you would not _have to_ overwrite the `selection_label`.

pupil_src/shared_modules/calibration_choreography/single_marker_plugin.py
pfaion: Then I would make this more obvious by using a `self.__has_ui` which gets set to `True` in `init_ui()` and to `False` in `__init__()` and `deinit_ui()`. And then just check for this and return in `_ui_update_visibility_digital_marker_config`. Makes it obvious why wouldn't continue here.

pfaion: Or even better check for `self.__has_ui` in the `marker_mode` property before calling `_ui_update_visibility_digital_marker_config`.

pupil_src/shared_modules/calibration_choreography/single_marker_plugin.py
pfaion: I'd be fine with both.

papr: @romanroibu Your call

pupil_src/shared_modules/calibration_choreography/single_marker_plugin.py
papr: @romanroibu What do you think about this?

pupil_src/shared_modules/calibration_choreography/base_plugin.py
pfaion: Which is probably what we would be doing anyway if we are ever going to add another choreography. I just mean that this is extra code for a use case that we don't have and that will probably never happen.

romanroibu: @pfaion from experience I found that using `@classmethod`s for things like this yields more flexible, understandable and predictable behaviour. You *could* do arbitrary transformations to produce the `selection_label` (including getting the label of the superclass, modifying it in some way, etc.). And since this is not performance critical code, I don't see any benefit switching to a class attribute. Ideally, this would be a `@classproperty` if existed, but since it doesn't, I think the flexibility of `@classmethod` is worth it.

Now that I'm looking at it, I think `label` was supposed to be a `@classmethod` as well, and instead of setting it to `None`, we would mark it as abstract for concrete subclasses to implement. If I remember correctly, that approach didn't work because we access `label` in `__init__subclass__`, but conceptually it is an abstract member for subclasses to implement.

pupil_src/shared_modules/calibration_choreography/single_marker_plugin.py
pfaion: Ok now I get it.
Which again makes me dislike the whole try-except because I didn't get it from reading the code  :smile: 

Yeah I don't know. Feels like the `Plugin` base class should also set `self.menu = []` or something. Maybe I've written too much C++ in the past weeks to be disturbed by Python's dynamically appearing member variables.

pupil_src/shared_modules/calibration_choreography/base_plugin.py
pfaion: Ok.

pupil_src/shared_modules/calibration_choreography/single_marker_plugin.py
pfaion: :snake:

romanroibu: Yeah, I don't like Python's dynamically appearing member variables, and the whole `try`/`except` construct here, but it seems cleaner than adding another `self.__has_ui` variable to keep track of the state of `self.__ui_slider_marker_scale`. Tradeoffs 🤷

pupil_src/shared_modules/pupil_producers.py
```diff
@@ -266,6 +277,11 @@ def _legend_font(self, scale):
 
 
 class Pupil_From_Recording(Pupil_Producer_Base):
+
+    @classmethod
+    def plugin_menu_label(cls) -> str:
+        return "Pupil From Recording"
```
papr, line 289:
```suggestion
        return "Pupil Data From Recording"
```

pupil_src/shared_modules/calibration_choreography/base_plugin.py
```diff
@@ -327,8 +337,18 @@ def base_class(cls):
         # even choreographies that subclass concrete choreography implementations.
         return CalibrationChoreographyPlugin
 
+    def _gazer_description_text(self) -> str:
```
papr, line 340:
```suggestion
    def _choreography_description_text(self) -> str:
```

PR #1916, opened 2020-06-16
Fix calibration validation for Pupil Capture v2

romanroibu:
##### Changes
- Calibration validation button is only enabled is the selected gazer class is the same as the active gazer class

### DIFF
--- a/pupil_src/shared_modules/calibration_choreography/base_plugin.py
+++ b/pupil_src/shared_modules/calibration_choreography/base_plugin.py
@@ -518,6 +518,10 @@ class CalibrationChoreographyPlugin(Plugin):
         )
 
     def _perform_start(self):
+        if self.__is_active:
+            logger.debug("[PROGRAMMING ERROR] Called _perform_start on an already active calibration choreography.")
+            return
+
         current_mode = self.__current_mode
 
         logger.info(f"Starting  {current_mode.label}")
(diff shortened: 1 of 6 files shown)


pupil_src/shared_modules/calibration_choreography/base_plugin.py
```diff
@@ -315,12 +315,17 @@ def on_choreography_successfull(
             calib_data = {"ref_list": ref_list, "pupil_list": pupil_list}
             self._start_plugin(self.selected_gazer_class, calib_data=calib_data)
         elif mode == ChoreographyMode.VALIDATION:
+            gazer_class = self.g_pool.active_gaze_mapping_plugin.__class__
+            assert gazer_class == self.selected_gazer_class
```
pfaion, line 319: Can this assert crash with the HMD setup?

When restarting from session settings, the HMD choreo will get reset to the default choreo, but the HMD gazer will still be active until a new gazer is calibrated. Is it possible to start the validation in this case? If not, this is probably fine, but if you can start the validation at that point, this will crash.

romanroibu: > Discussed this in person, but leaving response for history.

The validation button is enabled only if the condition of the `assert` holds, so it shouldn't ever happen. On the other hand, without this `assert`, if the active gazer is different from the selected gazer, then the parameters of the active gazer will not be compatible with the selected gazer class, which will cause a crash down the line, but will be harder to debug.

PR #1917, opened 2020-06-17
Support only gaze heatmap within surface in Pupil Capture

romanroibu:


pupil_src/shared_modules/surface_tracker/surface_tracker_online.py
```diff
@@ -95,6 +95,8 @@ def set_gaze_hist_len(val):
         )
 
     def recent_events(self, events):
+        if self._ui_heatmap_mode_selector:
```
pfaion, line 98:
```suggestion
        if self._ui_heatmap_mode_selector is not None:
```

PR #1921, opened 2020-06-23
Fix calibration validation for Pupil Capture v2 (2)

romanroibu:
Follow-up fixes for #1916


pfaion:
Does this now start the validation on the running gazer, even when a different gazer is selected in the UI (but not yet fit)?

papr:
@pfaion Yes, [it does](https://github.com/pupil-labs/pupil/pull/1921/files#diff-c6003491af63a149302f067f6ff03e2cR319-R320).

PR #1933, opened 2020-06-29
Prevent duplication of recorded calibration

pfaion:
This PR fixes that recorded calibrations can be duplicated.
The changes will only make sense once #1932 has been merged. 

Since they are not modifiable, duplicating them does not make sense as the copy would also be non-modifiable.


papr:
I will create a CU Task for that

pfaion:
@papr I agree, that would be ideal but would require some additional work in the calibration architecture, as I don't think there's currently a mechanism to access the underlying buttons. Let's postpone.

PR #1939, opened 2020-06-30
Introduce UniqueTaskManager to avoid duplicated post-hoc calibration background processes

papr:
Previously, a user could start multiple background tasks by accident by clicking the `Recalculate` button in the post-hoc calibration UI multiple times. This PR introduces a simple `PluginTaskManager` (previously used) subclass that replaces existing tasks that use a common identifier as the new task. By using the calibration/gaze mapper unique ids as identifiers, there can only be one active task for each calibration/gaze mapper.

### DIFF
--- a/pupil_src/shared_modules/gaze_producer/controller/calibration_controller.py
+++ b/pupil_src/shared_modules/gaze_producer/controller/calibration_controller.py
@@ -51,7 +51,7 @@ class CalibrationController(Observable):
         )
         task.add_observer("on_completed", on_calibration_completed)
         task.add_observer("on_exception", tasklib.raise_exception)
-        self._task_manager.add_task(task)
+        self._task_manager.add_task(task, identifier=calibration.unique_id)
         return task
 
     def on_calibration_computed(self, calibration):
--- a/pupil_src/shared_modules/gaze_producer/controller/calibration_controller.py
+++ b/pupil_src/shared_modules/gaze_producer/controller/calibration_controller.py
@@ -51,7 +51,9 @@ class CalibrationController(Observable):
         )
         task.add_observer("on_completed", on_calibration_completed)
         task.add_observer("on_exception", tasklib.raise_exception)
-        self._task_manager.add_task(task, identifier=calibration.unique_id)
+        self._task_manager.add_task(
+            task, identifier=f"{calibration.unique_id}-calibration"
+        )
         return task
 
     def on_calibration_computed(self, calibration):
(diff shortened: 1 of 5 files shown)


pfaion:
@papr Seems to work fine. What do you think about logging when we overwrite a task?

pupil_src/shared_modules/tasklib/manager.py
```diff
@@ -122,6 +125,10 @@ def add_task(self, task_new, identifier: str):
         UniqueTaskManager._patch_task(task_new, identifier)
         task_duplicated = self._get_duplicated_task(identifier)
         if task_duplicated is not None:
+            logger.debug(
+                f"{self} replacing queued task '{task_duplicated.process.name}' "
+                f"({identifier})"
+            )
```
pfaion, line 131: The `TaskInterface` does not define `.process`, which is only valid for background tasks.
Also I would add info if the task was already running vs still queued.
I wouldn't add `{self}` as we do that nowhere, but maybe we can actually add the line-number to the general logging format in the future.

```suggestion
            state = "running" if task_duplicated.running else "queued"
            logger.debug(f"Replacing {state} task with ID '{identifier}'")
```
Example output when calculating a calibration twice:
```
player - [DEBUG] tasklib.manager: Replacing running task with ID 'bd1f4749-5f55-484e-8131-6f8c87b8a5dd'
```

pupil_src/shared_modules/gaze_producer/controller/calibration_controller.py
```diff
@@ -51,7 +51,7 @@ def on_calibration_completed(status_and_result):
         )
         task.add_observer("on_completed", on_calibration_completed)
         task.add_observer("on_exception", tasklib.raise_exception)
-        self._task_manager.add_task(task)
+        self._task_manager.add_task(task, identifier=calibration.unique_id)
```
papr, line 54:
```suggestion
        self._task_manager.add_task(task, identifier=f"{calibration.unique_id}-calibration")
```

pfaion: @papr is this blacked? Or written here on GitHub?

papr: Github

papr: Updated for black format

pfaion:
I resolved the merge conflicts caused by #1940

PR #1943, opened 2020-07-01
Re-use existing recorded calib names

pfaion:
Calculating the number of a `Recorded Calibration #` would also take into account the identical calibration if it was extracted before. Since the calibration will be replaced, it should not be taken into account when generating the number suffix.

This bug was revealed by #1938


pfaion:
I was just rethinking and actually re-using the name does not make much sense when overwriting. I'll make a new PR with the changes.

PR #1944, opened 2020-07-01
Use sequential numbering for recorded calibrations

pfaion:
Calculating the number of a `Recorded Calibration #`  would also take into account the identical calibration if it was extracted before. This would result in **Recorded Calibration 2** being renamed to **Recorded Calibration 3** upon opening the recording again. This bug was revealed by #1938.

After a naive approach of even more filtering for generating the new number (#1943) I decided to rewrite the suffix generation logic. The key is that we only create the recorded calibrations from the loop where we extract them from the recorded notifications. Since these are sequential, we know exactly which number every recorded calibration has and can just pass this. Since they will be overwritten anyway, this is the easiest and cleanest solution imo.

### DIFF
--- a/pupil_src/shared_modules/gaze_producer/model/calibration_storage.py
+++ b/pupil_src/shared_modules/gaze_producer/model/calibration_storage.py
@@ -58,7 +58,7 @@ class CalibrationStorage(Storage, Observable):
         )
 
     def __create_prerecorded_calibration(
-        self, result_notification: CalibrationResultNotification
+        self, number: int, result_notification: CalibrationResultNotification
     ):
         timestamp = result_notification.timestamp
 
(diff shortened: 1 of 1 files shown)


pupil_src/shared_modules/gaze_producer/model/calibration_storage.py
```diff
@@ -179,6 +170,7 @@ def _load_calibration_from_file(self, file_name):
 
     def _load_recorded_calibrations(self):
         notifications = fm.load_pldata_file(self._rec_dir, "notify")
+        counter = 1
         for topic, data in zip(notifications.topics, notifications.data):
```
papr, line 174:
```suggestion
        topics_and_data = zip(notifications.topics, notifications.data)
        for counter, (topic, data) in enumerate(topics_and_data, start=1):
```

pupil_src/shared_modules/gaze_producer/model/calibration_storage.py
```diff
@@ -201,9 +193,10 @@ def _load_recorded_calibrations(self):
                 logger.debug(str(err))
                 continue
             calibration = self.__create_prerecorded_calibration(
-                result_notification=note
+                number=counter, result_notification=note
             )
             self.add(calibration, overwrite=True)
+            counter += 1
```
papr, line 199:
```suggestion
```

pupil_src/shared_modules/gaze_producer/model/calibration_storage.py
pfaion: There's two `continue`s in the loop. `enumerate()` would give us notification number, not the calibration number, if I see the correctly?

papr: Correct. Github had part of the loop collapsed.

PR #1947, opened 2020-07-01
Prevent crash when rendering eyeball outline with negative values

### LINKED ISSUE #1912: Eye overlay 3d pupil rendering issue in Pupil Player v2.0
This is actually 2 separate problems, one is that I have some sessions that have been causing the player to freeze when doing surface detection or just playing the video. When the application freezes I have to force quit the program.  I am still trying to sort out what is causing the crashes and will report on that later but for now I'll focus on what happens after the force quit.

When I restart v2.0 Pupil Player it opens a blank black terminal window and a blank white app window and does nothing else. Uninstalling and reinstalling did nothing and I've seen the behavior also in v1.23 so it seems to be how the settings are stored in a location different from the actual install location.

I found that if I run a copy of v1.22 it mentions the session file is for different version and resets things and v1.22 starts normally. If I close v1.22 and then start v2.0, it also starts normally. 

I've been able to reproduce this several times in v2.0 (and also in v1.23). I have sessions saved with v1.12 and it seems like some of these are are sometimes causing the errors where the program freezes. I'd prefer to use v2.0 for improved fixation detection and surface tracking. 

Where are the session settings files are stored on Windows 10? I'd like to make a copy of the settings so I can reinstate them if they get corrupted. With my current solution I have to reset all my plugin settings for eye overlay, surface detection and fixation detection which is annoying to keep redoing as I have ~150 tracking sessions I need to analyze and the error has happened several times today.

pfaion:
Fixes #1912 

I added better logging for potential errors here.
Example output for the recording from #1912:

```
player - [DEBUG] video_overlay.utils.image_manipulation: Error rendering 3D eye-ball outline! Skipping...
eye_ball: {'center': (246.79170715146995, 209.36043689754123), 'axes': (-639.8371093372608, -639.8371093372608), 'angle': 90.0}
<class 'cv2.error'>: OpenCV(4.2.0) /io/opencv/modules/imgproc/src/drawing.cpp:1927: error: (-215:Assertion failed) axes.width >= 0 && axes.height >= 0 && thickness <= MAX_THICKNESS && 0 <= shift && shift <= XY_SHIFT in function 'ellipse'
```

### DIFF
--- a/pupil_src/shared_modules/video_overlay/utils/image_manipulation.py
+++ b/pupil_src/shared_modules/video_overlay/utils/image_manipulation.py
@@ -100,7 +100,6 @@ class PupilRenderer(ImageManipulator):
                     f"eye_ball: {eye_ball}\n"
                     f"{type(e)}: {e}"
                 )
-                pass
 
     def render_ellipse(self, image, ellipse, color):
         outline = self.get_ellipse_points(
--- a/pupil_src/shared_modules/video_overlay/utils/image_manipulation.py
+++ b/pupil_src/shared_modules/video_overlay/utils/image_manipulation.py
@@ -10,9 +10,12 @@ See COPYING and COPYING.LESSER for license details.
 """
 
 import abc
+import logging
 
 import cv2
 import numpy as np
+
+logger = logging.getLogger(__name__)
 
 
 class ImageManipulator(metaclass=abc.ABCMeta):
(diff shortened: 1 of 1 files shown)


pupil_src/shared_modules/video_overlay/utils/image_manipulation.py
```diff
@@ -86,9 +89,17 @@ def render_pupil_3d(self, image, pupil_position):
...
+                #   - Sometimes raises ValueError when some components of the eye_ball
+                #     are 'nan'. TODO: Investigate where the 'nan' comes from.
+                #   - There are reports of negative eye_ball axes, raising cv2.error.
+                #     TODO: Investigate cause in detectors.
+                logger.debug(
+                    "Error rendering 3D eye-ball outline! Skipping...\n"
+                    f"eye_ball: {eye_ball}\n"
+                    f"{type(e)}: {e}"
+                )
                 pass
```
papr, line 103:
```suggestion
```

PR #1948, opened 2020-07-02
Ignore macOS-specific meta files when using glob

papr:
On some occasions, macOS stores file attributes in meta files that use `._` as prefix. These are matched in cases were we use glob with a pattern starting with `*`.

This PR excludes files starting with `.` to avoid matching the meta files and causing issues down the line.

All other use cases of glob in the project either exclude `.`-prefixed files already or use a concrete prefix which avoids matching meta files.

### DIFF
--- a/pupil_src/shared_modules/gaze_producer/model/legacy/calibration_storage_updater.py
+++ b/pupil_src/shared_modules/gaze_producer/model/legacy/calibration_storage_updater.py
@@ -33,7 +33,7 @@ class CalibrationStorageUpdater(CalibrationStorage):
         if not calib_dir.is_dir():
             return  # TODO: Raise exception - "calibrations" must be a directory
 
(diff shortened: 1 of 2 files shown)


papr:
This PR generally ignores all hidden files (`.`-prefixed). Technically, we could prevent new calibration names to be prefixed with `.` but I would like to avoid adding additional logic and creating potential backward compatibilities.

pfaion:
Hm ok, fair point. `CalibrationStorage` also already filters for `[!.]`
Maybe we should add a low-priority CU task for not allowing `.`-prefixed calibrations.
_Or_ adjust the filename for calibration storage to not start with the calibration name.

papr:
I think it is too late to adjust the naming convention. We should keep this type of issue in mind for the future.

PR #1949, opened 2020-07-02
Ensure frame publisher restarts when restarting NetworkAPIPlugin

pfaion:
A restart of the `NetworkAPIPlugin` would cause the frame publishing to stop.

The bug occurred in HMD-Eyes, where the eye preview was not updating when connecting.
This was caused by HMD-Eyes sending a start-plugin notification to ensure that `NetworkAPIPlugin` is running.
Since we allow replacing plugins _by class_ this would result in a restart.

I removed the start/stop logic from the `FramePublisherController`, since it is actually always started.
The `NetworkAPIPlugin` now sends a start notification in `__init__()` which fixes the behavior.

### DIFF
--- a/pupil_src/shared_modules/network_api/controller/frame_publisher_controller.py
+++ b/pupil_src/shared_modules/network_api/controller/frame_publisher_controller.py
@@ -20,9 +20,8 @@ logger = logging.getLogger(__name__)
 
 
 class FramePublisherController(Observable):
-    def on_frame_publisher_did_start(self, format: FrameFormat):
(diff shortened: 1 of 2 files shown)


papr:
Restructured commits to be more clear about what the changes are supposed to accomplish. Credit to @pfaion for finding and fixing the issue in c0fe638

pfaion:
@papr even better than what I had. Unfortunately I cannot approve of my own PR :smile: 

@romanroibu If you don't have any more suggestions, feel free to merge after you had a look at this.

PR #1951, opened 2020-07-02
Detect reference locations only on button press

pfaion:
`calculate_all()` is called on button press and also when pupil data changes.
We don't want to run reference location detection in any automatic workflow since it can take quite a while and consumes a lot of resources.

### DIFF
--- a/pupil_src/shared_modules/gaze_producer/controller/calculate_all_controller.py
+++ b/pupil_src/shared_modules/gaze_producer/controller/calculate_all_controller.py
@@ -29,14 +29,21 @@ class CalculateAllController:
 
     def calculate_all(self):
         """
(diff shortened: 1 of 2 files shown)


papr:
Does this PR fix the execution order issue?

pfaion:
I'm not sure which issue you mean.

Currently, we are running `calculate_all()` including the reference location detection every time the pupil data changes, which includes the case when we initially switch to post-hoc gaze.

With this PR, changing pupil data will not trigger a `calculate_all()` in the old sense if we do not yet have reference locations in storage. This should be the only change introduced here.

papr:
I was referring to `#5yt3jj` which is fixed by this PR.

PR #1952, opened 2020-07-02
Ensure calculate_all creates defaults

pfaion:
When pressing `Calculate all` when no calibration or no gaze mapper exists, nothing will happen.
We want a default calibration and gaze mapper to appear.

### DIFF
--- a/pupil_src/shared_modules/gaze_producer/controller/calculate_all_controller.py
+++ b/pupil_src/shared_modules/gaze_producer/controller/calculate_all_controller.py
@@ -43,6 +43,9 @@ class CalculateAllController:
         self._calculate_all_calibrations()
 
     def _calculate_all_calibrations(self):
+        if self._calibration_storage.is_empty:
+            self._calibration_storage.add_default_calibration()
+
         for calibration in self._calibration_storage:
             calculation_possible = (
                 self._calibration_controller.is_from_same_recording(calibration)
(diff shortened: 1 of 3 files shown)


pfaion:
**NOTE**
Unfortunately, the gaze_producer UI framework does not react to changes in the storage classes.
Currently, adding a new calibration/mapper works, but the UI will still show as if there's none and only display the "Add new" button.

Unfortunately, most of the UI does not react to changes in the underlying models (storages and items), but holds the logic for when to update itself. This obviously only works when changes were introduced through the UI. I think this is generally a sub-optimal design pattern for UI/Data separation. Even worse, there already seem to be a few parts where this dependency was reversed, e.g. the `GazeMapperMenu` reacts to changes in the `CalibrationStorage` to repaint itself when e.g. Calibrations are deleted/added etc. In order to make this PR work, we would need to add something similar, but this will make the flow of control even more confusing in the UI.

pfaion:
@papr not entirely sure what to do here. Options:
1) Leave it as it was: calculate_all does not create default calib/mapper if none exist
2) Let the UI for both hook into their corresponding storage and re-render always on `storage.add()`. This would mean a double re-render on clicking `new calib/mapper` since the base UI class already re-renders on this button click. The unnecessary overhead should be negligible, but we are then once more mixing different dependency directions/paradigms.

pfaion:
Option 1. might be the most rational here.

pfaion:
Going for option 1. and discarding this draft.

PR #1958, opened 2020-07-07
Coord_System: Avoid zero-width dimensions

papr:
`glOrtho` does not allow zero-width dimensions. Users of Coord_System should not have to care about this implementation detail. Therefore, Coord_system replaces the zero-width dimensions appropriately to avoid an error down the line.

This fixes an issue with pupil producer timelines who do not span a sufficiently large coordinate system due to an insufficient amount of data.

### DIFF
--- a/pupil_src/shared_modules/gl_utils/utils.py
+++ b/pupil_src/shared_modules/gl_utils/utils.py
@@ -188,6 +188,12 @@ class Coord_System(object):
 
     def __init__(self, left, right, bottom, top):
         super(Coord_System, self).__init__()
+        if left == right:
+            left -= 1
+            right += 1
+        if top == bottom:
+            top -= 1
+            bottom += 1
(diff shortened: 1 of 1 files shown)


pfaion:
@papr How does not having enough pupil data result in a zero-width `Coord_System`? Or is this specifically for **no** pupil data?

papr:
Pupil producers setup their timeline coordinate systems based on the available data in order to avoid the necessity of transforming the data. The `x` range uses min. and max. world timestamps as limits, the `y` range uses the min. and max. value of the corresponding parameter, e.g. pupil diameter. If there is no pupil data, the `y` range is set to `[0, 1]`. If the min. and max. values of the available parameter are the same, a zero-width dimension would be created. In this case, Coord_System extends the corresponding dimensions that the values would be drawn in the center of the timeline.

PR #1961, opened 2020-07-07
Allow GazerHMD3D to be initialised without eye translations

papr:
Eye translations are only required for fitting/calibrating. If the gazer is used purely for mapping gaze, the eye translations are not necessary. This PR allows the usage of recorded hmd calibrations in Pupil Player.


papr:
@pfaion Please test this branch with Capture and Service.

PR #1965, opened 2020-07-07
Post-hoc gaze mapping: Avoid non-monotonic progress updates

papr:
`gazer.map_pupil_to_gaze` does not yield gaze with monotonic timestamps. Binocular pupil matches are delayed internally. To avoid non-monotonic progress updates, we use the largest timestamp that has been returned up to the current point in time.


pfaion:
@papr develop or master?

papr:
Develop. This is nothing critical and I would rather wait for more potential issues before creating a second RC.

PR #1972, opened 2020-07-15
Disable read only buttons in post-hoc calibration menu

romanroibu:
This PR ensures that the `Duplicate` and `Delete` buttons in Post-hoc Calibration menu are disabled, if those actions are not available and would generate an error log. This happens when the calibration is from a different recording, or the calibration is pre-recorded.

EDIT: Requires https://github.com/pupil-labs/pyglui/releases/tag/v1.28 or later

### DIFF
--- a/pupil_src/launchables/player.py
+++ b/pupil_src/launchables/player.py
@@ -132,7 +132,7 @@ def player(
         )
 
         assert VersionFormat(pyglui_version) >= VersionFormat(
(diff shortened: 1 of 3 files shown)


papr:
Disabled buttons look awful at the moment. We will need to make a color adjustment in pyglui before merging this PR. I will set it to draft until then.

PR #1975, opened 2020-07-21
Improve Observable

ckbaumann:
@papr please add more reviewers as you see fit.

- Currently, the wrapper stores a reference to the wrapped function causing a reference cycle. This prevents garbage collection by reference counting. This is fixed by using a weakref to the object and referencing only the unbound method.
- I added test cases that test if wrapped methods still work before being wrapped (this is not related to the changed wrapper, these should have already been tested in the previous version)
- Use f-strings instead of `format` because we require Python >=3.6
- Add a descriptor for wrapped methods that prevents them from being replaced

I tested the code with python 3.6.10 and 3.8.2

### DIFF
--- a/pupil_src/shared_modules/observable.py
+++ b/pupil_src/shared_modules/observable.py
@@ -13,6 +13,8 @@ import functools
 import inspect
 import weakref
 
+from attr import has
+
 
 class ObserverError(Exception):
     pass
@@ -147,7 +149,12 @@ def _get_wrapper_and_create_if_not_exists(obj, method_name):
             "Attribute {} of object {} is a class method and, thus, "
             "cannot be observed!".format(method_name, obj)
         )
-    # TODO: Sanity check for observed method
+    elif _method_was_modified(obj, method_name):
+        raise TypeError(
+            f"Attribute '{method_name}' of object {obj} cannot be observed because it "
+            "is not part of the original class definition or has been assigned to a "
+            "method of a different object!"
+        )
     else:
         return _ObservableMethodWrapper(obj, method_name)
 
@@ -158,6 +165,20 @@ def _is_classmethod(obj, method_name):
         return isinstance(method_def, classmethod)
     except (AttributeError, KeyError):
         return False
+
+
+def _method_was_modified(obj, method_name):
+    method_in_original_class = hasattr(obj.__class__, method_name)
+    if not method_in_original_class:
+        return True
+    expected_func = getattr(obj.__class__, method_name)
+    expected_self = obj
+
+    method = getattr(obj, method_name)
+    bound_func = method.__func__
+    bound_self = method.__self__
+
+    return expected_self is not bound_self or expected_func is not bound_func
 
 
 def remove_observer(obj, method_name, observer):
--- a/pupil_src/shared_modules/observable.py
+++ b/pupil_src/shared_modules/observable.py
@@ -141,13 +141,13 @@ def _get_wrapper_and_create_if_not_exists(obj, method_name):
         return observed_method
     elif not inspect.ismethod(observed_method):
         raise TypeError(
-            "Attribute {} of object {} is not a method but {} and, thus, "
-            "cannot be observed!".format(method_name, obj, observed_method)
+            f"Attribute '{method_name}' of object {obj} is not a method but "
+            f"{observed_method} and, thus, cannot be observed!"
         )
     elif _is_classmethod(obj, method_name):
         raise TypeError(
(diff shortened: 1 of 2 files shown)


ckbaumann:
@pfaion Done, thanks for clarifying!

pfaion:
@ckbaumann what's the reason for adding the additional constraint of observed methods being "unmodified"? Is this required for the switch to weak references?

I know that we monkeypatch some methods in a couple of places throughout the codebase. I agree that this is not necessarily good practice, but sometimes it's the only reasonable way to get stuff done 😄 I'd be quite confident that so far we don't observe any patched method, so it shouldn't be a problem.

Having thought about this now, I would say it's probably good to prevent these until we need it because it might cause some headaches if used incorrectly. I justed wanted to know if you thought this would be a good change anyways or whether it was required for the changes you listed in the description above ☝️ 

If we're talking about observing patched methods we could also think about preventing patching observed methods (basically the other way round), which would still be possible with the proposed changes and would also be a big cause for trouble if done at some point. I think you should be able to prevent this by using the descriptor interface. This would make it even more consistent imho.

pupil_src/shared_modules/observable.py
```diff
@@ -187,29 +207,43 @@ def _get_wrapper_or_raise_if_not_exists(obj, method_name):
...
-        setattr(self._obj, self._method_name, self)
+        functools.update_wrapper(self, self._wrapped_func)
+        # functools adds a reference to the wrapped method that we need to delete
+        # to avoid cyclic references
+        self.__wrapped__ = None
+        setattr(self._obj_ref(), self._method_name, self)
 
     def remove_wrapper(self):
-        setattr(self._obj, self._method_name, self._original_method)
+        setattr(obj, self._method_name, self._get_wrapped_bound_method())
```
pfaion, line 238: This shouldn't work, as `obj` is not defined here.
But anyways I think this method is not used anywhere? Can we just remove it? What was it used for? Is there anything missing?

ckbaumann: Thanks a lot for finding this, I thought this was covered by tests, so I didn't test it explicitly.

I fixed the method and added a test for it. And I ran a code coverage analysis and we cover now almost everything with the tests.

Still, we can discuss if we want to remove it. It's currently not used, but when I originally wrote `Observable` I thought it would be cleaner to make it possible to remove the wrapper and it's not a big chunk of code, so I don't see much harm in keeping it. Maybe @papr or @romanroibu have thoughts on this as well.

romanroibu: Agree with @ckbaumann, if it's straight forward to implement and might be useful in the future, I don't think we should remove it. Especially since it's already there, and covered by tests. IMHO

ckbaumann:
@pfaion 

It depends on what you patch the method with. We only support observing methods, so if you replace a method with a mock function or so it won't work. If it is still a method after the patch, support is quite easy. I actually implemented the more general version first but then thought that this might be too confusing to read with no real benefit. You basically just need to store two weak refs, one to the object and one to `method.__self__` and then use the right object in the right context.

Yes, patching wrappers will lead to big problems. That was something that I shortly thought about when I wrote it and then just said that reasonable usage should not include this ;)

Let me look into decorators, I don't know enough about them to see if they might be a better option.

pupil_src/shared_modules/observable.py
papr: I agree, too. Resolving thread.

papr:
@ckbaumann Please be aware that the descriptor interface is something different than decorators.

ckbaumann:
@papr This was just a typo, decorators wouldn't be very useful indeed :)

pfaion:
I guess my idea was to overwrite `__set__` on wrapped observables such that it will error when trying to patch a wrapped method. You could still patch it with `setattr` I think (?) as that should circumvent the descriptor interface, but then you are deliberately shooting yourself in the foot.

pfaion:
Sorry for closing, did not intend to.

ckbaumann:
@pfaion I implemented your idea of using a descriptor to protect wrappers from being replaced. It should only add a moderate additional overhead and should make it at least more difficult to shoot yourself in the foot. I'm not 100% sure if this does not introduce subtle bugs, so please have a careful look.

pfaion (approved):
@ckbaumann sorry for the long delay. So far this looks good I think.
I played around with the observables a bit and think this should be fine now.

papr:
@ckbaumann Thank you very much!

PR #1976, opened 2020-07-21
Cross-platform display scaling

### LINKED ISSUE #959: Support Ubuntu display scaling
The Pupil applications do not adapt to display scaling available in the Ubuntu settings.

The right window is scaled appropriately by 2x while Capture is not:
screenshot from 2017-11-27 09-42-16

### LINKED ISSUE #1981: UI distortion on Windows when resetting window size or changing interface scale
I think those two effects might be related to the same underlying problem. They only happen on Windows.

1) When hitting "Reset window size" the layout is distorted until interacting with the UI or resizing the window manually:
screen

2) When changing the interface size, the layout is distorted until interacting with the UI or resizing the window manually.
Switching from 1.0 to 1.4:
image

Switching from 1.0 to 0.6:
image

romanroibu:
This issue addresses inconsistent behaviour regarding rendering and changing the window size on displays with different scale factors. Now, when clicking `"Reset window size"`, the application will reset the camera frame to its original size. Additionally, the `"Interface size"` selector has a more consistent behaviour, keeping the camera frame size the same, while scaling the UI elements to the selected size factor. Also, this PR makes sure that the size of the window is kept consistent across sessions, even when the display scale is different.

Fixes #959 
Fixes #1981 

---

##### Test-cases

- UI input should work correctly on a scaled display.
  - Ubuntu 18.4
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)
  - Windows
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)
  - macOS
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)
- Changing the user GUI scale (`"General" menu` > `"Interface size" dropdown`) from `1` to `0.6` and `1.4` keeps the camera frame size the same, while scaling the UI elements by `0.6` and `1.4` respectively. The window width changes only to account for the new scaled UI width.
  - Ubuntu 18.4
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)
  - Windows
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)
  - macOS
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)
- When resizing the window, the UI and the camera frame are redrawn according to the new size.
  - Ubuntu 18.4
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)
  - Windows
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)
  - macOS
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)
- Resetting the window size (`"General" menu` > `"Reset window size" button`) will resize the window to have the camera frame drawn at it's original size.
  - Ubuntu 18.4
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)
  - Windows
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)
  - macOS
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)
- Changing the scale factor of the display will affect the apparent window size according to the new display scale factor.
  - Ubuntu 18.4 (on Ubuntu, the display scale factor gets updated only on reopening the window)
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)
  - Windows
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)
  - macOS
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)
- Closing the window on a display, and reopening it on a display with a different scale factor will result in the window being scaled correctly according to the new factor (one way to test this is to click `"Reset window size"` before closing the window; when it's reopened on a display with a different scale factor, clicking `"Reset window size"` should have no effect)
  - Ubuntu 18.4
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)
  - Windows
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)
  - macOS
    - ✅ `world.py` (`@rom`)
    - ✅ `eye.py` (`@rom`)
    - ✅ `player.py` (`@rom`)
    - ✅ `service_ui.py` (`@rom`)


romanroibu:
@pfaion Could you please test this PR on Windows and Linux? Especially since the original issue referenced Ubuntu. Feel free to update the original description comment with your test-cases.

pfaion:
@papr @romanroibu what's actually the target behavior here? If you read through my notes in the hackMD, it might (I know they are a bit messy) have become clear that glfw does different things on the different operating systems and we might not be able to reproduce native system behavior entirely.

papr:
@romanroibu @pfaion Please see this test cases/requirements

Generally, when switching to a monitor with a different content scale, we need the following to work:
1. Window size adjustment (we get this for free using `GLFW_SCALE_TO_MONITOR=TRUE`)
2. Application content scale (`g_pool.gui.scale`) set to `glfwGetWindowContentScale()[0] * user_scale`
3. UI input mapping should work correctly using https://github.com/pupil-labs/pupil-invisible-monitor/blob/9324d412f52f82b10fdbd16fca40dbb9615b958c/src/pupil_invisible_monitor/window.py#L242-L248
4. `Reset window size` on a scaled display should result in same size as in 1. after dragging a default-sized window from an unscaled to a scaled display. (We might need to incoporate `glfwGetWindowContentScale()[0]` in the calculation of the default window size)

Regarding the limitations on Linux: Given that `glfwGetWindowContentScale` does not update until app-restart, we will have to live with that. 1.-4. should work on Windows and MacOS. On Linux as least 3. should work without restart. 1.-4. should work after restarting the application.

romanroibu:
@papr @pfaion I was looking at the suggestion by @papr to add `screen_to_pixel` from `Pupil Invisible Monitor`, but I'm not sure this is really necessary.

The implementation for `screen_to_pixel` applies the `hdpi_factor` if `window_size` is different from `framebuffer_size`. Currently, the `hdpi_factor` in `Pupil Core` is always applied without any condition. This PR also transitions from calculating the `hdpi_factor` based on the ration between `window_size` and `framebuffer_size`, to a `glfw` call.

Do you think there might be an edge-case where the `window_size` and `framebuffer_size` is different, but the `hdpi_factor` is `1`? If so, how should the UI input coordinates be scaled then? Simply applying `hdpi_factor` wouldn't result in any changes, so should the ration between the `window_size` and `framebuffer_size` (aka the old way of calculating the `hdpi_factor`)? Seems like this situation should happen, but you have more experience with OpenGL-related APIs.

pfaion:
I feel like the scale caching should not be done through a parameter, but rather through two variables for window_size and framebuffer_size (global unfortunately as this stuff is not wrapped in a class). Since I think you know exactly when the windows_size or the framebuffer_size change, you can just adjust the cache variables in these cases and then just compute the framebuffer_scale every time you need it. It's just a division, so we don't need to cache the result of the division I guess?

pfaion:
I'll give this a try after lunch.

papr:
I would prefer it if we cached `scale_content` and `scale_framebuffer` in `g_pool` as this would give plugins access to this value instead of having to calculate it by themselves. Also, I think it is important to calculate and save these scale factors instead of window and framebuffer sizes as they represent specific concepts that we mixed up before. My hope is that we can separate these concepts better in the future.

pfaion:
@papr which plugins do need access to that? I was hoping we could abstract this away.

pfaion:
Current Windows issues:
- The button "Reset Window Size" does not work correctly. The window stays at the "old size" (when you would hold a real-world ruler onto the display). As @papr said, we probably need to take the content scale into account when calculating the default window size.
- There are some text rendering issues on scaled monitors, e.g. here in the NetworkAPI:
    image
- Probably connected to the previous, the scrolling does not work correctly for menus with a lot of text. As you see, here is no scrollbar:
    image

While we're at "Reset Window Size", can we fix the long-standing issue of the layout being broken after hitting the button? (fixes itself upon the first repaint triggered by resizing or interacting with the UI):
image

papr:
> There are some text rendering issues on scaled monitors, e.g. here in the NetworkAPI:

@euryalus reported this a few days ago on macOS, too:
image

> While we're at "Reset Window Size", can we fix the long-standing issue of the layout being broken after hitting the button? (fixes itself upon the first repaint triggered by resizing or interacting with the UI):

Could you please clarify what you mean by "after hitting the button"? Which button are you referring, too?

papr:
> @papr which plugins do need access to that? I was hoping we could abstract this away.

Content scale is needed by several plugins that draw opengl ui, e.g. system graphs. You are right that the frambuffer scale is not relevant for plugins as we handle it in our glfw callbacks. Nonetheless, I would like to see

pfaion:
On Ubuntu we do not recognize switching the display scaling while Pupil is running, as was expected.
When starting on a scaled display, we have the same behavior as on windows: "Reset windows size" produces a wrong size and the text rendering is messed up.

I don't see the broken layout after hitting the "Reset window size" button, however. Maybe that was only always a Windows issue? It happens if you manually adjust the window size to a different size (dragging corners around) and the hit the button "Reset window size" in the general settings menu. The size resets, but the framebuffer is not updated and contains the stuff from the previous frame, but completely distorted (as you see in the screenshot). This happens as long as I can remember on Windows, I remember talking about this at some point but we didn't do anything about it so far. If you click a button or drag a corner again it jumps back to normal.

pfaion:
I noticed one more thing on Ubuntu: I closed Capture (on the scaled display), reset the display scaling to 100%, and restarted Capture. Now all windows were twice the size as normal. Even the eye-windows, which I did not touch at all while experimenting before. I'll check that as well on windows, can you test this on macOS?

pfaion:
Regarding the text stuff, I remember seeing some text rendering issues on retina displays on macOS with Pupil in the past as well, but iirc this was that text was cut off on the right side (as opposed to the bottom) when adjusting the width of the sidebar. Can you guys confirm this? Either on this branch or on develop?

pfaion:
I found one more issue on Windows only: after adjusting the gui scale/interface size, the menubar icons are messed up:
scaling from 1.0 to 1.4:
image
scaling from 1.0 to 0.6:
image

These issues behave the same way as the "Reset window size" distortion issue on Windows. If you manually adjust the window size again or interact with the UI, everything jumps back to normal.

papr:
>  If you manually adjust the window size again or interact with the UI, everything jumps back to normal.

This sounds like the frame buffer changed callback was not called correctly in thesecases. Could you check that?

pfaion:
Here's a screen recording of the "Reset window size" issue:

screen

pfaion:
Weirdly, calling `on_resize(main_window, *glfw.glfwGetFramebufferSize(main_window))` manually, even at the following frame does not solve the issue.

papr:
If this is an issue independent of scaling, please create a new issue for it. Please also update your glfw and check if glfw.glfwGetFramebufferSize(main_window) is returning expected values

pfaion:
Ok, since this also happens on develop, let's handle it separately.

Here's the summary of the issues I found with this branch, all of them on both Windows and Ubuntu:

- Pressing "Reset window size" does not reset to the correct size when on scaled displays.
- When closing Pupil on a scaled display and reopening it on an unscaled one, the window size is too large.
- Text behaves like the calculation of vertical text space is wrong on scaled displays.

papr:
> When closing Pupil on a scaled display and reopening it on an unscaled one, the window size is too large.

This means, we will have to store the content scale in the session settings, and if it is different on load than during storing, we have to divide the restored window size by the old/new factor.

> Text behaves like the calculation of vertical text space is wrong on scaled displays.

This is a pyglui issue and needs fixing outside of pupil.

papr:
> Text behaves like the calculation of vertical text space is wrong on scaled displays.

Fixed in https://github.com/pupil-labs/pyglui/pull/114

pupil_src/launchables/eye.py
```diff
@@ -370,11 +370,28 @@ def set_scale(new_scale):
...
             f_width, f_height = g_pool.capture.frame_size
-            f_width *= 2
-            f_height *= 2
-            f_width += int(icon_bar_width * g_pool.gui.scale)
-            glfw.glfwSetWindowSize(main_window, f_width, f_height)
+            frame_scale_factor = 2
+
+            # Scale the capture frame by a constant scale factor
+            f_width *= frame_scale_factor
+            f_height *= frame_scale_factor
```
pfaion, line 460: The added comment still doesn't tell me why we do this :stuck_out_tongue: 
Is this just so that the eye windows are more clearly visible?

papr:
@pfaion Please use this to set a minimum window size: https://www.glfw.org/docs/latest/window_guide.html#window_sizelimits

pupil_src/launchables/world.py
```diff
@@ -325,15 +325,63 @@ def get_timestamp():
...
             if w == 0 or h == 0:
                 return
-            hdpi_factor = glfw.getHDPIFactor(window)
-            g_pool.gui.scale = g_pool.gui_user_scale * hdpi_factor
+
+            # Always clear buffers on resize to make sure that the black stripes left/right
+            # are black and not polluted from previous frames. Make sure this is applied on
+            # the whole window and not within glViewport!
+            gl_utils.glClear(gl_utils.GL_COLOR_BUFFER_BIT)
+            gl_utils.glClearColor(0, 0, 0, 1)
```
pfaion, line 381: This comment is copy-pasted from PIMonitor and does not make sense here since we don't keep the viewport in a fixed aspect ratio (and thus don't have black stripes). I'm not even sure the clearing is required here? Is this just a copy-paste or is this required?

papr: Clearing the buffer is probably necessary as `consume_events_and_render_buffer()` is called afterward. If we do not clear the buffer we will probably see content being drawn on top of each other.

pfaion:
After resizing the window below minimal height, the placement of the hotkey-buttons is wrong. Need to check if this is still reproducible after enforcing the minimum window size.
Before:
image
After getting the height resize error:
image

pupil_src/launchables/world.py
pfaion: Shouldn't this snippet then be in all UIs? I see the glClear only in world.py currently.

pfaion: Also we don't have `consume_events_and_render_buffer()` in service_ui, but in all other launchables. Maybe we should make this consistent?

pfaion: Without this snippet, you get some overlap glitching while resizing, but this disappears once you stop dragging the borders around. This also what happens currently in eye, service, and player anyways, as the clear is missing there.

With it, the overlap glitches are just black.

pfaion: @papr should we get rid of this or add it to all launchables?

papr: @pfaion Add it to all launchables

pfaion:
@papr the latest commit should fix the opengl errors caused by too small sizes and thereby prevent the hotkey button positioning glitches. I used 2 * icon_size as a reasonable minimal size:
image

pfaion:
@papr please give this another last look on macOS. Linux and Windows look fine now.

PR #1978, opened 2020-07-22
Fix comment rot, pointing to renamed function

pfaion:
The function was renamed, the comment wasn't.


pupil_src/launchables/world.py
```diff
@@ -315,7 +315,7 @@ def get_timestamp():
             ("Dummy_Gaze_Mapper", {}),
             ("Display_Recent_Gaze", {}),
             # Calibration choreography plugin is added bellow by calling
```
papr, line 317:
```suggestion
            # Calibration choreography plugin is added below by calling
```

PR #1980, opened 2020-07-23
Fix serialization issue in zmq log handler

papr:
Previously, the zmq log handler's exception handling expected the "msg" field to be a string. If an object is passed that is not msgpack-serializable, the exeception handling will fail, e.g logger.debug([uuid.uuid4()]). This PR solves this issue by converting the "msg" field to a string in case of an initial serialization failure.


pfaion:
When would this happen?

PR #1984, opened 2020-07-27
Support PI headset on/off classifier data

romanroibu:
This PR adds support for integrating the output of the headset on/off classifier data into gaze data confidence for Pupil Invisible recordings. Previously, the confidence of gaze data for such recordings was always set to `1.0`. The new implementation sets the confidence to `0.0` if the classifier output for the given gaze datum is `0` (which means `off`), or `1.0` if the output is `255` (`on`). If the classifier data file is missing, the implementation falls back to the old implementation of setting the confidence to `1.0`.

### DIFF
--- a/pupil_src/shared_modules/video_capture/utils.py
+++ b/pupil_src/shared_modules/video_capture/utils.py
@@ -515,14 +515,14 @@ def pi_gaze_items(root_dir):
             timestamps = timestamps[:size]
 
         conf_data = load_worn_data(find_worn_path(timestamps_path))
-        if conf_data and len(conf_data) != len(timestamps):
+        if conf_data is not None and len(conf_data) != len(timestamps):
             logger.warning(
                 f"There is a mismatch between the number of confidence data ({len(conf_data)}) "
                 f"and the number of timestamps ({len(timestamps)})! Not using confidence data."
             )
             conf_data = None
 
-        if not conf_data:
+        if conf_data is not None:
             conf_data = (1.0 for _ in range(len(timestamps)))
 
         yield from zip(raw_data, timestamps, conf_data)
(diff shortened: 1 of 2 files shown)


romanroibu:
This still needs to be tested on PI recordings with and without the classifier data file (also for multi-part recordings).

pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -478,6 +488,18 @@ def load_raw_data(path):
...
+        if not (path and path.exists()):
+            return None
+
+        def confidence_from_uint8(uint8: int) -> float:
+            conf = uint8 / 255.0
+            conf = max(0.0, min(conf, 1.0))
+            return float(conf)
+
+        with open(path, "rb") as fh:
+            return [confidence_from_uint8(uint8) for uint8 in fh.read()]
```
papr, line 501: Please use `np.fromfile(str(path), "<u1")` to read the file and [`np.clip()`](https://numpy.org/doc/stable/reference/generated/numpy.clip.html) to clip the values. A conversion to float should not be required anymore after dividing by 255 as the result is converted automatically to float64.

pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -496,4 +518,16 @@ def load_raw_data(path):
...
+        conf_data = load_worn_data(find_worn_path(timestamps_path))
+        if conf_data and len(conf_data) != len(timestamps):
+            logger.warning(
+                f"There is a mismatch between the number of confidence data ({len(conf_data)}) "
+                f"and the number of timestamps ({len(timestamps)})! Not using confidence data."
+            )
+            conf_data = None
+
+        if not conf_data:
+            conf_data = (1.0 for _ in range(len(timestamps)))
```
papr, line 531: Initially, I was incline to suggest using `np.ones_like(timestamps)` but given that this will be used in a `yield from` statement, I think your solution is prefarable.

papr (approved):

papr:
Test recordings:
- [Single-part](https://drive.google.com/file/d/1iwBFh3mdi1xb9Im5VxIn4BrHpiRauLML/view?usp=sharing)
- [Multi-part](https://drive.google.com/file/d/18uQaTagITFP3x19qAyJZXT_XR0NBRHR-/view?usp=sharing)

PR #1987, opened 2020-08-03
Eye window: Set 500x500 as default content size

papr:
Using #1976 as the target as this PR is branched of it. Can be changed to `develop` once it is reviewed.

Changes:
- **Use 192x192 as the preferred frame size explicitly** - Previously, 192x192 was chosen implicitly for Pupil Cam2 and Cam3 as the UVC backend will fallback to the closest frame size if the preferred frame size is not available.
-  **Set 500x500 as default content area** - Pupil Cam1 uses 4:3 resolutions. Pupil Cam2 and Cam3 use 1:1 resolutions. As all Pupil Core and VR/AR add-ons are shipped with Pupil Cam2 and Cam3 cameras, we adjust the default eye window size to a 1:1 content aspect ratio. The size of 500 was chosen s.t. the menu still fits.

@pfaion Please test with different scaling options on
- [X] Linux _(PFA)_
- [X] Windows _(PFA)_

---
- [x] Tested on macOS with normal + retina display

### DIFF
--- a/pupil_src/launchables/eye.py
+++ b/pupil_src/launchables/eye.py
@@ -398,7 +398,12 @@ def eye(
             glfw.glfwWindowHint(glfw.GLFW_VISIBLE, 0)  # hide window
         title = "Pupil Capture - eye {}".format(eye_id)
 
-        width, height = session_settings.get("window_size", (640 + icon_bar_width, 480))
+        # Pupil Cam1 uses 4:3 resolutions. Pupil Cam2 and Cam3 use 1:1 resolutions.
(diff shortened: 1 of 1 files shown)


pfaion:
This is what the default eye window looks like now:
image


@papr is there a specific reason for the 3D detector menu being opened? I think it has the least useful settings of all menus in the eye window. Previously it also contained the 2D settings, which might be a lot more useful to users. Or maybe we just show the camera image with all menus collapsed?

papr:
It is opened because it is the most recently loaded plugin. I think collapsing the menus by default makes sense.

PR #1992, opened 2020-08-19
Cleanup camera models

pfaion:
This is a refactoring PR, cleaning up the code and logic used in camera_models.py.
It will make it easier to integrate eye camera intrinsics for correct usage of the focal length in the 3D detector later on.

Summary of changes:
1. Moved all duplicated code between the radial and fish-eye camera into the existing base class

2. Used a factory pattern and proper base-class registration for file (de)serialization of camera models.
It seems that there also was some bug caused by the previous logic, but it did not have any visible effect:
Saved dummy intrinsics would be loaded as a radial camera model instead of the dummy camera model.

3. I reworded all uses of "calibration" referring to intrinsics to "intrinsics". Also fixed inconsistent usage of "pre-recorded", sometimes meaning "included in the recording" and sometimes meaning "default values that we recorded previously at the company".

### DIFF
--- a/pupil_src/shared_modules/camera_models.py
+++ b/pupil_src/shared_modules/camera_models.py
@@ -152,89 +152,9 @@ pre_recorded_calibrations = {
 }
 
 
-def load_intrinsics(directory, cam_name, resolution):
-    """
-    Loads a pre-recorded intrinsics calibration for the given camera and resolution. If no pre-recorded calibration is available we fall back on default values.
-    :param directory: The directory in which to look for the intrinsincs file
-    :param cam_name: Name of the camera, e.g. 'Pupil Cam 1 ID2'
-    :param resolution: Camera resolution given as a tuple.
-    :return: Camera Model Object
-    """
-    file_path = os.path.join(
-        directory, "{}.intrinsics".format(cam_name.replace(" ", "_"))
-    )
-    try:
-        calib_dict = load_object(file_path, allow_legacy=False)
-
-        if calib_dict["version"] < __version__:
-            logger.warning("Deprecated camera calibration found.")
-            logger.info(
-                "Please recalibrate using the Camera Intrinsics Estimation calibration."
-            )
-            os.rename(
-                file_path, "{}.deprecated.v{}".format(file_path, calib_dict["version"])
-            )
-
-        intrinsics = calib_dict[str(resolution)]
-        logger.info("Previously recorded calibration found and loaded!")
-    except Exception:
-        logger.info(
-            "No user calibration found for camera {} at resolution {}".format(
-                cam_name, resolution
-            )
-        )
-
-        if (
-            cam_name in pre_recorded_calibrations
-            and str(resolution) in pre_recorded_calibrations[cam_name]
-        ):
-            logger.info("Loading pre-recorded calibration")
-            intrinsics = pre_recorded_calibrations[cam_name][str(resolution)]
-        else:
-            logger.info("No pre-recorded calibration available")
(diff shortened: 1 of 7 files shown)


pfaion:
I recommend looking at the commits one by one

pupil_src/shared_modules/camera_models.py
```diff
@@ -299,13 +219,121 @@ def solvePnP(
...
+        super().__init_subclass__(*args, **kwargs)
+        # register subclasses by cam_type
+        if cls.cam_type == ...:
+            raise NotImplementedError("Subclass needs to define 'cam_type'!")
+        if cls.cam_type in Camera_Model.subclass_by_cam_type:
+            raise ValueError(
+                f"Error trying to register camera model {cls}: Camera model with"
+                f" cam_type '{cls.cam_type}' already exists:"
+                f" {Camera_Model.subclass_by_cam_type[cls.cam_type]}"
+            )
```
papr, line 235: Is this really necessary? This prevents people from implementing their own class for a specific `cam_type`.

pupil_src/shared_modules/camera_models.py
```diff
@@ -249,28 +250,29 @@ def save(self, directory, custom_name=None):
             "resolution": self.resolution,
             "cam_type": self.cam_type,
         }
-        # Try to load previous camera calibrations
+        # Try to load previous camera intrinsics
```
papr, line 253:
```suggestion
        # Try to load previously recorded camera intrinsics
```

pupil_src/shared_modules/camera_models.py
```diff
@@ -249,28 +250,29 @@ def save(self, directory, custom_name=None):
...
+            f"Intrinsics for camera {cam_name} at resolution {self.resolution} saved"
             f" to {save_path}"
         )
 
     @staticmethod
     def from_file(directory, cam_name, resolution):
         """
-        Loads a pre-recorded intrinsics calibration for the given camera and resolution. If no pre-recorded calibration is available we fall back on default values.
+        Loads recorded intrinsics for the given camera and resolution. If no recorded
+        intrinsics are available we fall back on default values.
```
papr, line 275:
```suggestion
        intrinsics are available we fall back to default values.
```

pupil_src/shared_modules/camera_models.py
```diff
@@ -280,36 +282,37 @@ def from_file(directory, cam_name, resolution):
...
             ):
-                logger.info("Loading pre-recorded calibration")
-                intrinsics = pre_recorded_calibrations[cam_name][str(resolution)]
+                logger.info("Loading default intrinsics!")
+                intrinsics = default_world_intrinsics[cam_name][str(resolution)]
             else:
-                logger.info("No pre-recorded calibration available")
-                logger.warning("Loading dummy calibration")
+                logger.info("No default intrinsics available!")
+                logger.warning("Loading dummy intrinsics!")
```
papr, line 315:
```suggestion
                logger.warning("No default intrinsics available! Loading dummy intrinsics!")
```

papr (changes requested):

pupil_src/shared_modules/camera_models.py
pfaion: I put it there since the DummyModel inherits from RadialModel and if you forget to specify cam_type when inheriting, your child class will take over everything from the parent class. This can easily happen and is most certainly not what you want.

Does it make sense to implement my own radial model with which I want to overwrite the behavior of the built-in model? I see my original reason as a more pressing issue.

papr (approved):

PR #1995, opened 2020-08-24
Pass eye camera focal_length to 3D detector

pfaion:
This PR uses the eye camera intrinsics (specifically focal_length) to get a more accurate 3D eyeball position estimate.

Depends on [pupil-detectors](https://github.com/pupil-labs/pupil-detectors) v1.1.1, upgrade with
```
pip install -U pupil-detectors
```

**Summary**
- Added known intrinsics for the eye cameras to camera_models.py
- On recording stop, current eye camera intrinsics are saved to the recording (like with world.intrinsics)
- The 3D detector is initialized with `g_pool.capture.intrinsics.focal_length` (cannot be changed at runtime)
- On focal_length changes (e.g. by changing the resolution), the 3D detector and all dependent UI is re-initialized
- Added clear warnings that loading dummy intrinsics is not good
- Added upgrade mechanism (RecordingInfo v2.3) which creates eye(0/1).intrinsics files when opening an older recording, as these will be tried to load from. To make it easy, we just serialize all intrinsics for eye cameras Cam1 and Cam2, the correct intrinsics will be picked later based on the actual resolution. This is not correct if Cam3 was used to recording, but there is no way of figuring this out.

**Tests**
1) Run Capture and open the 3D detector debug window while wearing a headset. The eyeball-center z coordinate should be relatively stable (+/-5mm). Switch the resolution of the eye camera. The debug window reopens with a reset model. After stabilizing, the eyeball-center z coordinate should settle down to the same value as with the other resolution (+/-5mm).

2) Make a recording, the recording should contain `eye0.intrinsics` and `eye1.intrinsics` files.

3) Open the recording in player and check that the post-hoc pupil detection works fine.

4) Open an older recording in player and check that the post-hoc pupil detection works fine. The recording should contain `eye0.intrinsics` and `eye1.intrinsics` files afterwards.

5) Open a fresh pupil mobile recording in player and check that the post-hoc pupil detection works fine. The recording should contain `eye0.intrinsics` and `eye1.intrinsics` files afterwards.

### DIFF
--- a/pupil_src/launchables/eye.py
+++ b/pupil_src/launchables/eye.py
@@ -503,6 +503,7 @@ def eye(
         toggle_general_settings(True)
 
         g_pool.writer = None
+        g_pool.rec_path = None
 
         # Register callbacks main_window
         glfw.glfwSetFramebufferSizeCallback(main_window, on_resize)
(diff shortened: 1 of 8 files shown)


pfaion:
@papr @romanroibu I added some tests to the description, maybe you can coordinate with regards to testing and having a look at the code changes.

pfaion:
@papr should I add a comment that only the focal lengths of the eye camera intrinsics were actually measured and the rest is currently just default values (for principal point) and identities (for distortion coordinates)?

papr:
Yes, please do so

pfaion:
I added GitHub labels as long as we have not released pupil-detectors v1.1.1 to PyPI.

romanroibu (approved):
performed the tests (except for the last step with mobile recording) and everything seems to work file 👍🏻

pfaion:
I'll fix the merge conflicts...

papr (approved):

PR #1996, opened 2020-08-25
Use regex to match gaze timestamp files

romanroibu:
This PR updates the gaze timestamps file matching from a glob pattern to a regex.

### DIFF
--- a/pupil_src/shared_modules/video_capture/utils.py
+++ b/pupil_src/shared_modules/video_capture/utils.py
@@ -501,7 +501,7 @@ def pi_gaze_items(root_dir):
     # - is followed by one or more digits
     # - ends with "_timestamps.npy"
     gaze_timestamp_paths = match_contents_by_name_pattern(
-        pl.Path(root_dir), "^gaze ps[0-9]+_timestamps.npy$"
+        pl.Path(root_dir), r"^gaze ps[0-9]+_timestamps.npy$"
     )
 
     for timestamps_path in gaze_timestamp_paths:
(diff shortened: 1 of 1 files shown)


pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -526,3 +529,10 @@ def load_worn_data(path):
             conf_data = (1.0 for _ in range(len(timestamps)))
 
         yield from zip(raw_data, timestamps, conf_data)
+
+
+def match_contents_by_name_pattern(parent_dir: Path, name_pattern) -> T.List[Path]:
+    # Get all non-recursive directory contents
+    contents = parent_dir.absolute().glob("*")
+    # Filter content that matches the name by regex pattern
+    return [c for c in contents if re.match(name_pattern, c.name) is not None]
```
pfaion, line 538: I think the function name could be better, `match` sounds like it returns a boolean or some match-object.
Maybe `filter_dir`? This would be similar to `os.listdir()` which also does not go into sub-directories.
Also, the type `str` of name_pattern is missing.

Proposals:
A) Simple, maybe not clear if filtering name or full path?
```python
def filter_dir(dir: Path, name_pattern: str) -> T.List[Path]:
```
B) More verbose about what is filtered
```python
def filter_dir_by_name_pattern(dir: Path, name_pattern: str) -> T.List[Path]:
```

papr: I would suggest `matched_files_from_pattern` and using `contents = filter(Path.is_file, parent_dir.iterdir())`

pfaion (approved):

papr (approved):
Test:
1. Made a recording with 16 parts
2. Duplicated `gaze ps13.time`
3. Renamed to `gaze ps13 dup.time`
4. Opened in v2.3 -> error while opening as expected, no crash.
5. Opened in PR branch -> worked as if I did not perform steps 1-3

PR #1998, opened 2020-08-25
Add pye3d support

pfaion:
This PR adds a plugin for the pye3d pupil detector.

The plugin will only be loaded if pye3d is available. In this case, it will replace the default 3D pupil detector.
Please note that pye3d is distributed under a different license than Pupil.

**Open Questions**
1) Since the base class `PupilDetectorPlugin` abstracts away `handle_events()` and child classes only implement `detect()`, we generally do not have access to detection results of other detectors in `detect()`.
    However, any 3D detector needs the 2D result. To increase confusion, the old 3D detector does not work with the 2D pupil datum, but with the raw binary blob. Pye3D on the other hand needs the 2D pupil datum. We should come up with a better way of passing this, maybe just give access to the `events` dict in `detect()`?

2) Should we merge this even if the integration is not yet fully complete? Theoretically, we can release this code, even if it does not yet have all features (e.g. freezing the eye model), as it won't be used without having pye3d available anyways. I think that's a great chance for working incrementally without having one massive feature branch in the end.


papr:
I am getting this error when running the pye3d detector during post-hoc detection:
```py
  File "/Users/papr/work/pupil/pupil_src/launchables/player.py", line 672, in player
    p.recent_events(events)
  File "/Users/papr/work/pupil/pupil_src/shared_modules/pupil_producers.py", line 422, in recent_events
    assert pm.PupilTopic.match(topic, eye_id=pupil_datum["id"])
AssertionError
```

P.S. it might be necessary to delete the `user_settings_eye*` files from your **player_settings** since the eye process tries to restore the legacy 3d detector if it was loaded in the previous session.

pupil_src/shared_modules/pupil_detector_plugins/pye3d_plugin.py
```diff
@@ -0,0 +1,126 @@
...
+            raise RuntimeError("No 2D detection result! Needed for pye3D!")
+
+        datum_2d["raw_edges"] = []
+        result = self.detector.update_and_detect(
+            datum_2d, frame.gray, debug_toggle=self.is_debug_window_open
+        )
+
+        eye_id = self.g_pool.eye_id
+        result["timestamp"] = frame.timestamp
+        result["topic"] = f"pupil.{eye_id}"
```
papr, line 55:
```suggestion
        result["topic"] = f"pupil.{eye_id}.{self.identifier}"
```

pupil_src/shared_modules/pupil_detector_plugins/pye3d_plugin.py
```diff
@@ -0,0 +1,126 @@
...
+logger = logging.getLogger(__name__)
+
+
+class Pye3DPlugin(PupilDetectorPlugin):
+    uniqueness = "by_class"
+    icon_font = "pupil_icons"
+    icon_chr = chr(0xEC19)
+
+    label = "Pye3D"
+    order = 0.101
```
papr, line 29:
```suggestion
    order = 0.101
    identifier = "3d"
```

papr (changes requested):

pupil_src/shared_modules/pupil_detector_plugins/pye3d_plugin.py
```diff
@@ -0,0 +1,127 @@
...
+        datum_2d["raw_edges"] = []
+        result = self.detector.update_and_detect(
+            datum_2d, frame.gray, debug_toggle=self.is_debug_window_open
+        )
+
+        eye_id = self.g_pool.eye_id
+        result["timestamp"] = frame.timestamp
+        result["topic"] = f"pupil.{eye_id}.{self.identifier}"
+        result["id"] = eye_id
+        result["method"] = "3d c++"
```
papr, line 58: We might want to change that line though.
```suggestion
        result["method"] = "pye3d"
```

papr (approved):
Tested all pupil-related plugins. Works as expected.

pupil_src/shared_modules/pupil_detector_plugins/pye3d_plugin.py
papr: Not sure if this would affect hmd-eyes though!

pfaion: This would be quite a big change actually.
Since Pupil v2.0 we adjusted the docs everywhere that you now get data from both detectors and that you should filter on the `method` key, e.g. for `3d c++`.
So a lot of users might rely on that and we would have to change a lot of documentation.
On the one hand, it makes sense that this is now a new method.
On the other hand, since we replace the old 3D we might just re-use the method key?

papr: I was not aware of the recommendation to filter for the complete string. The gazers only filter for the "2d"/"3d" part. In this case, I would resolve this with a `won't fix`

PR #2000, opened 2020-08-31
Fix post-hoc calibration timeline update

romanroibu:
This PR fixes an issue where changing the calibration in post-hoc gaze mapping would not update the timeline to reflect the calibration's range.

---

##### Steps to reproduce
1. Make a Capture recording including at least one calibration
1. Open recording in Player
1. Select post-hoc gaze calibration
1. At least one calibration should show up as "Recorded Calibration"
1. The "default gaze mapper" has "Recorded Calibration" selected as calibration
1. Create a new "Default Calibration"
1. Change calibration range to something else than the complete recording
1. In the "default gaze mapper" select "default calibration" as calibration
    - [!!!] At this point the calibration range timeline should update to the corresponding range
1. Changing the calibration range again correctly updates the visualisation


papr (approved):
Reproduced testing procedure. Was not able to reproduce issue using the PR branch. Works as intended.

PR #2005, opened 2020-09-04
Allow resetting 3D model via network notification

pfaion:


papr (approved):

papr (changes requested):
black failed

pfaion:
The base branch `pye3d` did not yet have the recent black update.
I merged `develop` into `pye3d` and rebased this PR.
Should be fine now.

pupil_src/shared_modules/pupil_detector_plugins/pye3d_plugin.py
```diff
@@ -58,6 +58,14 @@ def detect(self, frame, **kwargs):
 
         return result
 
+    def on_notify(self, notification):
+        super().on_notify(notification)
+
+        subject = notification["subject"]
+        if subject == "pupil_detector.3d.reset_model":
+            if "id" not in notification or notification["id"] == self.g_pool.eye_id:
```
romanroibu, line 66: Shouldn't this be `if "id" in notification and notification["id"] == self.g_pool.eye_id`?
Or am I misunderstanding the intent of the boolean logic here?

pfaion: The idea was that if you omit the `id`, the notification will just apply to all eye processes.
What I used here is basically a boolean shortcut for:
```python
if "id" not in notification:
  # simply apply to all eye processes
  self.reset_model()
else if notification["id"] == self.g_pool.eye_id:
  # filter for specific eye processes
  self.reset_model()
```
If you want to, I could write it like this to make the two separate use cases more explicit.

papr: I am ok with both variants.

papr (approved):

PR #2007, opened 2020-09-08
Ensure windows are visible when recovering a session

romanroibu:
This PR fixes issues related to windows being outside the display bounds when recovering a session.

### DIFF
--- a/pupil_src/shared_modules/glfw.py
+++ b/pupil_src/shared_modules/glfw.py
@@ -520,6 +520,35 @@ __c_callbacks__ = {}
 __py_callbacks__ = {}
 
 
+_Margins = T.NamedTuple(
+    "_Margins", [("left", int), ("top", int), ("right", int), ("bottom", int)]
+)
+
+
+_Rectangle_Tuple = T.NamedTuple(
+    "_Rectangle", [("x", int), ("y", int), ("width", int), ("height", int)]
+)
+
+
+class _Rectangle(_Rectangle_Tuple):
+    def intersection(self, other: "_Rectangle") -> T.Optional["_Rectangle"]:
+        in_min_x = max(self.x, other.x)
+        in_min_y = max(self.y, other.y)
+
+        in_max_x = min(self.x + self.width, other.x + other.width)
+        in_max_y = min(self.y + self.height, other.y + other.height)
+
+        if in_min_x < in_max_x and in_min_y < in_max_y:
+            return _Rectangle(
+                x=in_min_x,
+                y=in_min_y,
+                width=in_max_x - in_min_x,
+                height=in_max_y - in_min_y,
+            )
+        else:
+            return None
+
+
 def glfwGetError():
     _glfwGetError = _glfw.glfwGetError
     _glfwGetError.argtypes = [POINTER(c_char_p)]
@@ -625,14 +654,6 @@ def glfwGetFramebufferSize(window):
     width, height = c_int(0), c_int(0)
     _glfw.glfwGetFramebufferSize(window, byref(width), byref(height))
     return width.value, height.value
-
-
-_Rectangle = T.NamedTuple(
-    "_Rectangle", [("x", int), ("y", int), ("width", int), ("height", int)]
-)
-_Margins = T.NamedTuple(
-    "_Margins", [("left", int), ("top", int), ("right", int), ("bottom", int)]
-)
 
 
 def glfwGetWindowFrameSize(window) -> _Margins:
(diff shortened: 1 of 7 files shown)


romanroibu:
@papr @pfaion The implementation currently in this PR tests if the origin of the window is within any of the connected monitors. This, however, can lead to situations where even windows with as little as 1 pixel (origin) within the monitor keep their position. I was thinking about an alternative implementation where we would calculate an intersection rectangle of the window and each monitor. The rectangle would be valid if it's at least `10x10` pixels (or any other size). If there are no valid rectangles - the position is reset to default. Wanted to see what you think about current and proposed implementations.

pupil_src/shared_modules/glfw.py
```diff
@@ -424,6 +424,7 @@ class GLFWmonitor(Structure):
 glfwGetPrimaryMonitor = _glfw.glfwGetPrimaryMonitor
 glfwGetPrimaryMonitor.restype = POINTER(GLFWmonitor)
 # glfwGetMonitorPos               = _glfw.glfwGetMonitorPos
+# glfwGetMonitorWorkarea          = _glfw.glfwGetMonitorWorkarea
```
papr, line 427: Is this commented out on purpose?

pupil_src/shared_modules/glfw.py
```diff
@@ -656,6 +657,14 @@ def glfwGetMonitorPos(monitor):
     return xpos.value, ypos.value
 
 
+def glfwGetMonitorWorkarea(monitor):
+    xpos, ypos, width, height = c_int(0), c_int(0), c_int(0), c_int(0)
+    _glfw.glfwGetMonitorWorkarea(
+        monitor, byref(xpos), byref(ypos), byref(width), byref(height)
+    )
+    return xpos.value, ypos.value, width.value, height.value
```
papr, line 665: I think returning a named tuple would make sense here, e.g. `Rect` (similar to `NSRect` in ObjC)

papr (commented):
I like the idea about the intersection area. But since windows are only movable if their "header" is clickable, I suggest applying your logic such that the header has a minimum visible area before resetting the window to its default position.

pupil_src/shared_modules/glfw.py
romanroibu: Yes; added only for consistency with other definitions.

pupil_src/shared_modules/glfw.py
```diff
@@ -624,6 +628,21 @@ def glfwGetFramebufferSize(window):
     return width.value, height.value
 
 
+_Rectangle = collections.namedtuple("_Rectangle", ["x", "y", "width", "height"])
+_Margins = collections.namedtuple("_Margins", ["left", "top", "right", "bottom"])
```
papr, line 632: I feel like we could use `T.NamedTuple` subclasses here instead and also move `rectangle_intersection()` to `_Rectangle.intersection()`

pupil_src/shared_modules/glfw.py
```diff
@@ -761,3 +790,46 @@ def get_framebuffer_scale(window) -> float:
...
+        width=content_rect.width + frame_edges.left + frame_edges.right,
+        height=content_rect.height + frame_edges.top + frame_edges.bottom,
+    )
+
+
+def get_window_title_bar_rect(window) -> _Rectangle:
+    frame_rect = get_window_frame_rect(window)
+    frame_edges = glfwGetWindowFrameSize(window)
+    return _Rectangle(
+        x=frame_rect.x, y=frame_rect.y, width=frame_rect.width, height=frame_edges.top
```
papr, line 816: I don't understand why `height=frame_edges.top`. If I am not mistaken `top` corresponds to
> in screen coordinates, [...] the top edge of the window frame

romanroibu: `top` here represents the distance between the top edge of the window frame and the top edge of the window content, which in case of `top` should be equal or slightly bigger than then height of the window title bar.

papr: Ah, makes sense. Thanks for clarifying!

papr:
@pfaion Please give this recovery logic a try on Windows. Please try it with different multi-screen setups, i.e. disconnected screen to the left/top/bottom/right.

romanroibu:
@papr @pfaion I incorporated all the latest feedback. Please let me know if there's anything else that might need changing.

papr (changes requested):
When I was suggesting to use `typing.NamedTuple`, I was mostly referring to using the [type annotation possibility to define the tuple](https://docs.python.org/3/library/typing.html#typing.NamedTuple), making it more readable. Please adopt the type annotation variant accordingly.

papr (approved):

pfaion (approved):
Works fine in different window layouts!

PR #2009, opened 2020-09-08
Fix GLError when minimizing eye windows on Windows

romanroibu:


pupil_src/launchables/eye.py
```diff
@@ -236,9 +236,10 @@ def consume_events_and_render_buffer():
             glfw.glfwMakeContextCurrent(main_window)
             clear_gl_screen()
 
-            glViewport(0, 0, *g_pool.camera_render_size)
-            for p in g_pool.plugins:
-                p.gl_display()
+            if 0 <= g_pool.camera_render_size[0] and g_pool.camera_render_size[1]:
```
papr, line 239: Why do you only check for the x value to be greater than `0`?

pfaion: In PIMonitor we used something like
```python
if 0 not in self.framebuffer_size:
```
which apparently works, but I agree that checking also for negative numbers makes sense.
Maybe
```python
if all(c > 0 for c in g_pool.camera_render_size):
```
makes sense?

pupil_src/launchables/eye.py
```diff
@@ -290,10 +290,14 @@ def on_resize(window, w, h):
             nonlocal window_size
             nonlocal content_scale
 
-            # Always clear buffers on resize to make sure that there are no overlapping
-            # artifacts from previous frames.
-            gl_utils.glClear(gl_utils.GL_COLOR_BUFFER_BIT)
-            gl_utils.glClearColor(0, 0, 0, 1)
+            framebuffer_size = glfw.glfwGetFramebufferSize(window)
+            is_minimized = framebuffer_size is not None and 0 in framebuffer_size
```
papr, line 294: If the `framebuffer_size is None` we would be calling `glClear()`. I do not know if this is wanted behavior.

romanroibu: The boolean expression for `is_minimized` was taken from [`pupil-invisible-monitor`](https://github.com/pupil-labs/pupil-invisible-monitor/blob/v1.3/src/pupil_invisible_monitor/window.py#L76).

pfaion: PIM uses a very different general architecture for this, you cannot just copy-paste the code literally.
In Pupil, `glfwGetFramebufferSize` always returns a tuple, so it can never be `None`.

As I wrote in [my previous comment](https://github.com/pupil-labs/pupil/pull/2009#discussion_r485403182), which @papr resolved, I liked your previous implementation of checking for positive numbers (as opposed to checking only non-zero) even more. Not sure why we dropped that.

pfaion (changes requested):

pfaion (changes requested):
I'm still getting an error when un-minimizing the eye-window again:
```
eye0 - [ERROR] root: Encountered PyOpenGL error: GLError(
        err = 1281,
        description = b'invalid value',
        baseOperation = glClear,
        cArguments = (GL_COLOR_BUFFER_BIT,)
)
```
This does not necessarily mean that it breaks in `glClear` btw!
Actually it most likely has nothing to do with `glClear`, as the `invalid value` error cannot happen in our cases, as we always only pass pre-defined bitmasks.

@romanroibu the issue is that OpenGL does not raise any issues with commands that you run. Instead you can check the last error code for it's value. That means that if you run a command, which sets error code 1281 and then you run 10 other successful commands, the error code will still be 1281. Good OpenGL libraries check and clear the error code after every command. Unfortunately we mix a "good" OpenGL library (PyOpenGL) with our own pyglui mess, which does not check. So very often PyOpenGL reports issues that were caused by pyglui commands run before. I assume that this is what happens here as well.

romanroibu:
@papr @pfaion Solved the issues by early exiting the body of the `on_resize` callback if the window is minimised (which is checked by calling `glfwGetWindowAttrib(window, GLFW_ICONIFIED)`)

Tested on Windows and macOS, but please give it a run on the platforms you have access to.

papr (approved):

pfaion (approved):

PR #2011, opened 2020-09-14
Remove 3D pupil detector

romanroibu:
This PR removes the default 3D pupil detector plugin from Pupil Core and introduces a simplified pupil detector network api.

---

[ClickUp Task](https://app.clickup.com/t/7wuqjf)


pupil_src/launchables/eye.py
```diff
@@ -221,17 +221,24 @@ def get_timestamp():
...
-            ("HMD_Streaming_Manager", {}),
-            ("File_Manager", {}),
-            ("PupilDetectorManager", {}),
-            ("Roi", {}),
         ]
 
+        # Detectors needs to be loaded first to set `g_pool.pupil_detector`
+        for pupil_detector in [default_2d, default_3d]:
+            if pupil_detector is not None:
+                default_plugins.append((pupil_detector.__name__, {}))
```
pfaion, line 229: Do we actually need `default_2d`/`default_3d` now?
Maybe just a list `default_detectors`, which includes 2d and if available pye3d?
Then we don't have this problem of setting one to `None` and checking for it.
Not sure where else we reference the `default_2d`/`default_3d` from pupil_detector_plugins/__init__.py?

pupil_src/launchables/eye.py
```diff
@@ -248,24 +248,15 @@ def load_runtime_pupil_detection_plugins():
         default_plugins = [
             # TODO: extend with plugins
             (default_capture_name, default_capture_settings),
+            ("UVC_Manager", {}),
+            *[(p.__name__, {}) for p in available_detectors],
+            ("NDSI_Manager", {}),
+            ("HMD_Streaming_Manager", {}),
+            ("File_Manager", {}),
+            ("PupilDetectorManager", {}),
```
papr: I cannot find a reference for that. I think this plugin got removed?

pupil_src/shared_modules/pupil_detector_plugins/detector_base_plugin.py
```diff
@@ -66,130 +104,135 @@ def init_ui(self):
...
-            event[EVENT_KEY] = [detection_result]
+        # Append the new detection result to the previous results
+        event[EVENT_KEY] = previous_detection_results + [detection_result]
 
+        # Save the most recent detection result for visualization
         self._recent_detection_result = detection_result
 
-    @abc.abstractmethod
-    def detect(self, frame, **kwargs):
+    def on_notify(self, notification):
```
papr, line 133: I feel like we should add a thin layer of sanity checking and wrap this into a try/except

pupil_src/shared_modules/pupil_detector_plugins/detector_base_plugin.py
```diff
@@ -66,130 +104,135 @@ def init_ui(self):
...
+        subject_components = notification["subject"].split(".")
+
+        if subject_components[0] == "pupil_detector":
```
papr, line 136: I feel like we should add a thin layer of sanity checking and wrap the code below into a try/except

pupil_src/shared_modules/pupil_detector_plugins/detector_base_plugin.py
```diff
@@ -66,130 +104,135 @@ def init_ui(self):
...
+            # {
+            #   "suject": "pupil_detector.<action>",
+            #   "target": <target_process>, # optinal
+            #   ...
+            # }
+
+            action = subject_components[1]
```
papr, line 143: I feel like we should add sanity checking for this line, i.e. check if there is an actual second component

papr (commented):
Meeting notes
- every detector publishes their own property broadcast
- differentiate between target process and target detector
- use class name to identify detector target
- process and detector targets are optional for requesting the current property values
- setting properties requires at least a detector target
- notifications emitted should include eye_id and detector class name in the subject as well as in the payload

pupil_src/shared_modules/pupil_detector_plugins/detector_base_plugin.py
```diff
@@ -112,7 +115,7 @@ def recent_events(self, event):
             return
 
         # Detect if resolution changed
-        self._update_frame_size_if_changed(new_frame_size=(frame.width, frame.height))
+        self.__update_frame_size_if_changed(new_frame_size=(frame.width, frame.height))
 
         # Get results of previous detectors
         # TODO: This is currently used by Pye3D to get the results of the 2D detection
```
papr, line 121: @romanroibu What do you think? Should we move this into the pye3d detector plugin? Also, instead of the `event` key, it might be sensible to have a detector-specific key, maybe based on the `identifier`?

papr (approved):

PR #2015, opened 2020-09-23
Replace internal GLFW bindings with pyGLFW

romanroibu:
This PR replaces the internal [`glfw.py`][2] bindings with external [`pyGLFW`][1].

Install the external bindings using `pip`:
```sh
pip install glfw
```

---

[ClickUp Task](https://app.clickup.com/t/1rgy34)

[1]: https://github.com/FlorianRhiem/pyGLFW
[2]: https://github.com/pupil-labs/pupil/blob/8a85c534a170d945b9fe2cdb30260d71415b9679/pupil_src/shared_modules/glfw.py


pfaion:
@romanroibu can you add `pip install glfw` to https://github.com/pupil-labs/pupil/blob/master/.travis/run_tests.sh
This should fix the Travis errors hopefully.

romanroibu:
@pfaion Thanks for reminding me 👍🏻 I added your suggestion, and also updated the dependency documentation.

The only thing I haven't touched is the bundle deployment scripts, but there seem to be [useful functions](https://github.com/FlorianRhiem/pyGLFW/blob/master/glfw/library.py) to get the path(s) to the library files. @papr Please let me know if you want me to try to update these; I don't have a lot of experience with `PyInstaller`, but can take a stab at it.

pupil_src/shared_modules/gl_utils/utils.py
```diff
@@ -211,3 +222,89 @@ def __exit__(self, *exc):
...
+                y=in_min_y,
+                width=in_max_x - in_min_x,
+                height=in_max_y - in_min_y,
+            )
+        else:
+            return None
+
+
+def get_content_scale(window) -> float:
+    # TODO: Explicitly rename to `get_content_scale_x`
```
pfaion, line 259: @romanroibu here's a TODO left, did you plan to change something?

romanroibu: Discussed this with @papr; We decided the name is good enough to hide the implementation details of how the scale value is calculated, and descriptive enough for the callers.

papr:
Bundling works. Only issue: There was a case where the clipboard could not be accessed on Windows. Might be a race condition.

papr:
[`pyGLFW 2.0.0` was just released](https://github.com/FlorianRhiem/pyGLFW/releases/tag/v2.0.0). Most notably, the default error reporting changed from `raise` to `warn`.

pfaion:
@romanroibu is there not a better version to change `ERROR_REPORTING` than to change this after every import? This seems somehow very verbose and easy to forget.

romanroibu:
@pfaion for most use-cases it's _probably_ enough to change the `ERROR_REPORTING` after the imports in the `launchables`, since the rest of the code runs in the context of a launchable process. However, I'm not sure if no other process is calling any `glfw` method, so I just added this statement after each import for safety.

I couldn't think of a better way to globally change the error reporting behaviour.

papr (approved):

PR #2021, opened 2020-10-01
Fix eye overlay plugin video file filter

romanroibu:
This PR fixes issues related to the eye overlay plugin that caused crashes when the confidence value was not valid (`inf` or `nan`). Additionally, it fixes the way this plugin was searching for eye videos in the recordings and was causing it to mistakenly use `eye1.instrinsics` as a video source.

---
[Click-Up Task](https://app.clickup.com/t/8jxww7)

### DIFF
--- a/pupil_src/shared_modules/video_overlay/plugins/eye_overlay.py
+++ b/pupil_src/shared_modules/video_overlay/plugins/eye_overlay.py
@@ -97,15 +97,9 @@ class Eye_Overlay(Observable, Plugin):
         return overlay
 
     def _video_path_for_eye(self, eye_id: int) -> str:
-        rec_dir = self.g_pool.rec_dir
-
-        # Create eye video file filter for eye_id
-        file_filter = PupilRecording.FileFilter(rec_dir)
-        file_filter = file_filter.eye_id(eye_id)
-        file_filter = file_filter.videos()
-
         # Get all eye videos for eye_id
-        eye_videos = list(file_filter)
+        recording = PupilRecording(self.g_pool.rec_dir)
+        eye_videos = list(recording.files().videos().eye_id(eye_id))
 
         if eye_videos:
             return str(eye_videos[0])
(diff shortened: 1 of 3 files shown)


papr:
@romanroibu have you tested other eye related functionality, e.g. Offline pupil detection, eye export, etc

pupil_src/shared_modules/video_overlay/plugins/eye_overlay.py
```diff
@@ -94,14 +96,20 @@ def _setup_eye(self, eye_id, prefilled_config):
...
-            return next(video_path_candidates)
-        except StopIteration:
+
+        # Create eye video file filter for eye_id
+        file_filter = PupilRecording.FileFilter(rec_dir)
+        file_filter = file_filter.eye_id(eye_id)
+        file_filter = file_filter.videos()
+
+        # Get all eye videos for eye_id
+        eye_videos = list(file_filter)
```
pfaion, line 108: The intended usage for the FileFilter was more something like:
```python
recording = PupilRecording(rec_dir)
eye_videos = list(recording.files().videos().eye_id(eye_id))
```
with the `.files()` function there's no need to expose the FileFilter explicitly.
Not sure if this makes it clearer what's happening.

pupil_src/shared_modules/video_overlay/utils/image_manipulation.py
```diff
@@ -112,7 +118,8 @@ def render_ellipse(self, image, ellipse, color):
         outline = [np.asarray(outline, dtype="i")]
         cv2.polylines(image, outline, True, color, thickness=1)
 
-        center = (int(ellipse["center"][0]), int(ellipse["center"][1]))
+        center = ellipse["center"]
+        center = (float_to_int(center[0]), float_to_int(center[1]))
```
papr, line 122: Converting all fields from NaN to 0 was not my intention. Rather, I would prefer a behavior where the whole datum to be treated as if the _confidence_ was 0 if any of the relevant fields (confidence, ellipse center) is NaN-valued.

pupil_src/shared_modules/video_overlay/plugins/eye_overlay.py
papr: @romanroibu Please adjust this accordingly in order to be consistent with other uses of the filefilter across the code base.

papr (changes requested):

papr (approved):

papr:
Tested using the recording that caused the initial issue (eye overlay works, but without visualization, as expected) and a separate recording (eye overlay works with visualization, as expected.).

PR #2027, opened 2020-10-14
Fix FPS drop on macOS for occluded windows

romanroibu:
This PR fixes an issue on macOS that caused FPS drops when the window becomes occluded.

---

[ClickUp Task](https://app.clickup.com/t/7cnn4h)

### DIFF
--- a/pupil_src/launchables/eye.py
+++ b/pupil_src/launchables/eye.py
@@ -577,6 +577,16 @@ def eye(
 
         frame = None
 
+        if platform.system() == "Darwin":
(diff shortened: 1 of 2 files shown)


papr (approved):
I like that this fix was only applied minimally. There is no need to add it to Player or the Service UI. 👍

Also, there does not seem to be any visible tearing as one would expect based on the documentation:
> Using a swap interval of zero can be useful for benchmarking purposes, when it is not desirable to measure the time it takes to wait for the vertical retrace. However, a swap interval of one lets you avoid tearing.
Source: https://www.glfw.org/docs/latest/window_guide.html#buffer_swap

PR #2040, opened 2020-11-04
fix cpu graph refresh, improve performance and UI look

bpinsard:
I noticed that the CPU graph is being updated too often on my setup, resulting in non-readable UI, not sure if it is the general case.
I also noticed that this too frequent refresh is using a good chunk of CPU in the eye process.

This small PR makes the CPU graph to be updated at the screen refresh rate, but I am not sure this is the optimal choice.
WDYT?

### DIFF
--- a/pupil_src/launchables/eye.py
+++ b/pupil_src/launchables/eye.py
@@ -733,10 +733,9 @@ def eye(
                 for result in event.get(EVENT_KEY, ()):
                     pupil_socket.send(result)
 
-            cpu_graph.update()
-
             # GL drawing
             if window_should_update():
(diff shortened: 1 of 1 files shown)


papr:
Hi, thanks for coming up with a solution. @romanroibu Could you please try reproducing the issue?

@bpinsard Meanwhile, could you elaborate on your setup?

bpinsard:
Our setup uses a custom video_capture module for a single eye and no world video, apart from that it is the latest commit on the develop branch.
Maybe the difference is that the NDSI backend is not used, so that loop might run faster.

papr (approved):

PR #2068, opened 2020-12-08
Update msgpack dependency

### LINKED ISSUE #1830: ValueError: int is not allowed for map key when using msgpack 1.0.0

romanroibu:
Fixes #1830 and #1419


pupil_src/shared_modules/raw_data_exporter.py
```diff
@@ -421,14 +421,18 @@ def dict_export(
...
+                    "1", [None, None, None]
+                )
                 #
                 gaze_normals0_3d = raw_value["gaze_normals_3d"].get(
-                    0, [None, None, None]
+                    "0", [None, None, None]
                 )
                 gaze_normals1_3d = raw_value["gaze_normals_3d"].get(
-                    1, [None, None, None]
+                    "1", [None, None, None]
```
papr, line 435: Fallback to int-keys if string-keys are not available.

requirements.txt
```diff
@@ -2,7 +2,7 @@
 ### Third-party
 ###
 cython
-msgpack==0.5.6
+msgpack
```
papr, line 5:
```suggestion
msgpack>=1.0.0
```

papr (changes requested):

pupil_src/shared_modules/file_methods.py
```diff
@@ -23,10 +23,6 @@
 import msgpack
 import numpy as np
 
-assert (
-    msgpack.version[1] == 5
-), "msgpack out of date, please upgrade to version (0, 5, 6 ) or later."
-
```
papr, line 29: Please reintroduce this test with `msgpack.version[0] == 1`.

papr (changes requested):

papr:
Please make a PR to adjust https://docs.pupil-labs.com/developer/core/overview/#gaze-datum-format accordingly.

papr (approved):

PR #2082, opened 2021-01-11
Support 200hz gaze data in PI recording

romanroibu:
This PR adds support for 200Hz gaze data in Pupil Invisible recordings. The recording upgrade procedure will prioritise 200Hz data, if it's available, and fall back to the old behaviour in case it's not available.

### DIFF
--- a/pupil_src/shared_modules/video_capture/utils.py
+++ b/pupil_src/shared_modules/video_capture/utils.py
@@ -599,10 +599,6 @@ def pi_gaze_items(root_dir):
             conf_data = load_worn_data(find_worn_path(timestamps_path))
         # Otherwise, don't load confidence data
         else:
-            # TODO: Should we handle the case were:
-            #       - not is_200hz_data_available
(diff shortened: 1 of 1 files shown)


pupil_src/shared_modules/video_capture/utils.py
```diff
@@ -525,9 +556,55 @@ def load_worn_data(path):
...
+        elif (
+            not is_200hz_data_available and find_worn_path(timestamps_path) is not None
+        ):
+            conf_data = load_worn_data(find_worn_path(timestamps_path))
+        # Otherwise, don't load confidence data
+        else:
+            # TODO: Should we handle the case were:
+            #       - not is_200hz_data_available
+            #       - find_worn_200hz_path(timestamps_path) is not None
+            # meaning: there is no 200hz data, but there is (only) 200hz confidence?
```
papr, line 605:
```suggestion
```
No, I do not think we need to handle it.

papr (approved):
Works as expected 👍

PR #2101, opened 2021-02-23
Support recordings with decoding issues

papr:
Ignores audio and video decoding issues in the middle of the stream. (Decoding issues at the start of the stream disables the audio/video stream as it assumes the complete stream to be broken.)

Previously, the software crashed when encountering such decoding issues. The new implementation can lead to audio desync or silence when encountering any audio decoding issues. This is assumed to be an acceptable trade-off given the previous behavior.

### DIFF
--- a/pupil_src/shared_modules/av_writer.py
+++ b/pupil_src/shared_modules/av_writer.py
@@ -24,7 +24,7 @@ from av.packet import Packet
 
 import audio_utils
 from video_capture.utils import Video, InvalidContainerError
-from player_methods import iter_catch
+from methods import iter_catch
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


pupil_src/shared_modules/av_writer.py
```diff
@@ -24,6 +24,7 @@
 
 import audio_utils
 from video_capture.utils import Video, InvalidContainerError
+from player_methods import iter_catch
```
romanroibu, line 27: I get an `ImportError` unless I change it to `from methods import iter_catch`

papr, line 27:
```suggestion
from methods import iter_catch
```

romanroibu (approved):

PR #2108, opened 2021-03-03
Fixes laggy playback of PI recordings after seeking.

papr:
TODO:
- [x] Fix new frame drop every few seconds

----
[ClickUp Task](https://app.clickup.com/t/ch1ued)

### DIFF
--- a/pupil_src/shared_modules/audio_utils.py
+++ b/pupil_src/shared_modules/audio_utils.py
@@ -79,8 +79,9 @@ def _load_audio_single(file_path, return_pts_based_timestamps=False):
     if return_pts_based_timestamps:
         timestamps = start + packet_pts * stream.time_base
 
-    # pts seeking requires Python ints; convert after `packet_pts * stream.time_base`
-    # to leverage numpy element-wise function application
+    # pts seeking requires primitive Python integers and does not accept numpy int types;
+    # `.tolist()` converts numpy integers to primitive Python integers; do conversion after
+    # `packet_pts * stream.time_base` to leverage numpy element-wise function application
     packet_pts = packet_pts.tolist()
 
     try:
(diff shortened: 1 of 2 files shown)


pupil_src/shared_modules/audio_utils.py
```diff
@@ -72,23 +72,25 @@ def _load_audio_single(file_path, return_pts_based_timestamps=False):
...
-        )
-        try:
-            container.seek(0)
-        except av.AVError as err:
-            logger.debug(f"{err}")
-            return None
+        timestamps = start + packet_pts * stream.time_base
+
+    # pts seeking requires Python ints; convert after `packet_pts * stream.time_base`
+    # to leverage numpy element-wise function application
```
papr, line 84:
```suggestion
    # pts seeking requires primitive Python integers and does not accept numpy int types;
    # `.tolist()` converts numpy integers to primitive Python integers; do conversion after
    # `packet_pts * stream.time_base` to leverage numpy element-wise function application
```

pupil_src/shared_modules/av_writer.py
```diff
@@ -502,11 +501,7 @@ def _iterate_audio_frames_and_audio_gaps(audio_parts):
         for part_idx, audio_part in enumerate(audio_parts):
 
             frames = audio_part.container.decode(audio=0)
-            frames = iter_catch(frames, av.AVError)
             for frame, timestamp in zip(frames, audio_part.timestamps):
-                if frame is None:
-                    continue  # ignore audio decoding errors
```
romanroibu, line 508: Is this intended? wasn't this introduced for handling decoding issues #2107 ?

romanroibu (changes requested):
- There are many changes that have nothing to do with the issue itself. I think it's better if there is a separate PR for `black`-related changes, and a separate PR for the README change (or direct commit on `develop` is also fine in this case IMO)
- There are some changes that I'm not sure are intended, that undo some of the fixes from previous PRs (see comments)

papr:
@romanroibu looks like Github got confused after I merged #2110 into master and this previous branch, but not into develop. Merging master into develop did not fix the situation. Therefore, I rebased this branch on top of develop and force-pushed. Please review again.

PR #2112, opened 2021-03-09
Implement accuracy visualisation and calculation for HMD calibrations

papr:
@romanroibu Please test with a screen marker calibration:
1. Calibrate and check if visualization looks correct (Linux, Mac)
2. Consistency with previous version (one OS is sufficient)
    a. Make recording in latest release
    b. Open and setup post-hoc calibration and validation in latest Player release
    c. Check if validation result is the same as when running this branch
    
I tested (1) with screenmarker and hmd calibration on Windows already.


romanroibu (approved):
Tested; everything works as expected 👍🏻

PR #2125, opened 2021-03-31
Disable "Pupil from Recording" for Pupil Invisible recordings

romanroibu:
This PR removes the `"Pupil from Recording"` pupil data option from Player for Pupil Invisible recordings. These recordings don't contain any pupil data, so this PR is only meant to provide consistency in the way Player works. This PR also prevents raw data exports from generating an empty `pupil_positions.csv` when no pupil data is available.

---
[ClickUp Task](https://app.clickup.com/t/fh1cqd)

### DIFF
--- a/pupil_src/shared_modules/pupil_producers.py
+++ b/pupil_src/shared_modules/pupil_producers.py
@@ -291,6 +291,53 @@ class Pupil_Producer_Base(Observable, System_Plugin_Base):
             self.glfont.pop_state()
 
 
+class DisabledPupilProducer(Pupil_Producer_Base):
+    """
+    This is a stub implementation of a pupil producer,
+    intended to be used when no (other) pupil producer is available.
+    """
+
+    @classmethod
+    def is_available_within_context(cls, g_pool) -> bool:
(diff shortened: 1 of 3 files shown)


pupil_src/shared_modules/pupil_producers.py
```diff
@@ -57,6 +57,24 @@ def plugin_menu_label(cls) -> str:
     def pupil_data_source_selection_label(cls) -> str:
         return cls.plugin_menu_label()
 
+    @staticmethod
+    def available_pupil_producer_plugins(g_pool) -> list:
+        def is_plugin_included(p, g_pool) -> bool:
```
papr, line 62: Since this is a closure, you can think about simplifying the call signature.
```suggestion
        def is_plugin_included(p) -> bool:
```

romanroibu: true, but in general I try to not have references to outer scope inside function definitions. makes it easier to extract into a separate method later if needed.

papr (approved):

PR #2126, opened 2021-04-01
Fix toggling pupil detection on/off

romanroibu:
This PR fixes a bug that prevented toggling pupil detection on/off in Capture's general settings.

---

[ClickUp Task](https://app.clickup.com/t/d733te)

### DIFF
--- a/pupil_src/launchables/world.py
+++ b/pupil_src/launchables/world.py
@@ -105,7 +105,7 @@ def world(
 
     def detection_enabled_setter(is_on: bool):
         g_pool.pupil_detection_enabled = is_on
-        n = {"subject": "set_pupil_detection_enabled", "value": is_on}
+        n = {"subject": "pupil_detector.set_enabled", "value": is_on}
         ipc_pub.notify(n)
 
     try:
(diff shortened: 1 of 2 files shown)


pupil_src/launchables/world.py
```diff
@@ -100,12 +100,13 @@ def start_stop_eye(eye_id, make_alive):
         else:
             stop_eye_process(eye_id)
 
-    def detection_enabled_getter() -> bool:
-        return g_pool.pupil_detection_enabled
+    def detection_enabled_getter() -> int:
+        return int(g_pool.pupil_detection_enabled)
```
papr, line 104: Why not convert to bool directly? That would fit better to the `detection_enabled_setter` call signature
```suggestion
    def detection_enabled_getter() -> bool:
        return bool(g_pool.pupil_detection_enabled)
```

romanroibu: The `selected` property of `ui.Switch` is of type `bint` ([see here](https://github.com/pupil-labs/pyglui/blob/3d52f8ed9ad7d6cdc7dab08a2b7bd5aa3f454df5/pyglui/ui_elements.pxi#L238)). These type annotations are more for documentation purposes, and to keep in mind to do the `int` to `bool` conversion in the setter before sending the notification.

papr (approved):

PR #2137, opened 2021-04-20
Fix validation crash if no reference data was collected

romanroibu:
---
[ClickUp Task](https://app.clickup.com/t/h71f2u)

### DIFF
--- a/pupil_src/shared_modules/accuracy_visualizer.py
+++ b/pupil_src/shared_modules/accuracy_visualizer.py
@@ -368,10 +368,12 @@ class Accuracy_Visualizer(Plugin):
         return True
 
     def recalculate(self):
+        NOT_ENOUGH_DATA_COLLECTED_ERR_MSG = (
+            "Did not collect enough data to estimate gaze mapping accuracy."
+        )
+
         if not self.recent_input.is_complete:
-            logger.info(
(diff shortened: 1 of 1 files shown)


pupil_src/shared_modules/accuracy_visualizer.py
```diff
@@ -371,6 +385,9 @@ def recalculate(self):
             succession_threshold=self.succession_threshold,
         )
 
+        if not results.is_valid:
+            return
```
papr, line 389: I suggest adding user feedback at this point. The suggestion below is just a copy from the warning above. A warning instead of an info message might be more appropriate. What do you think?
```suggestion
            logger.info(
                "Did not collect enough data to estimate gaze mapping accuracy."
            )
            return
```

pupil_src/shared_modules/accuracy_visualizer.py
```diff
@@ -386,6 +386,9 @@ def recalculate(self):
         )
 
         if not results.is_valid:
+            logger.info(
+                "Did not collect enough data to estimate gaze mapping accuracy."
+            )
```
papr, line 391: Please change to a warning in both cases and remove duplicated string definition.

papr (approved):

PR #2140, opened 2021-05-05
Export surface tracker marker detections

romanroibu:


papr:
Works for me. Waiting for the verification that performance is not impacted by this change.

papr (approved):

PR #2144, opened 2021-05-18
Exit fullscreen window before destroying

### LINKED ISSUE #2143: Calibration window does not close correctly when displayed fullscreen on different screen
Reported [here ](https://discord.com/channels/285728493612957698/285728493612957698/841349920140951592) and [here](https://discord.com/channels/285728493612957698/285728493612957698/841362816244908042). Only reproducible on Windows.

Edit: Actually, it is not reproducible on all Windows machines. Since #2144 comes with a tradeoff (fullscreen window flashes white when closing) and one can workaround the original issue by using window mode, we have decided to not fix this issue.

papr:
Fixes #2143

Known issue:
- When exiting fullscreen mode the screen seems to flicker shortly

### DIFF
--- a/pupil_src/shared_modules/calibration_choreography/controller/gui_window.py
+++ b/pupil_src/shared_modules/calibration_choreography/controller/gui_window.py
@@ -131,6 +131,9 @@ class GUIWindow(Observable):
         if not self.is_open:
             return
         with self._switch_to_current_context():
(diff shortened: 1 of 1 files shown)


papr:
See https://github.com/pupil-labs/pupil/issues/2143#issue-894510320 for reasoning about closing this PR.

PR #2151, opened 2021-06-21
Added imu_timeline.py to shared_modules

N-M-T:
@romanroibu My original method for determining whether or not a PI recording was loaded was to identify the presence of IMU data (line 269). Then I used bool 'is_invisible_rec' to disable interaction with the Plugin (e.g. line 288) when it wasn't. 

I think these checks can safely be removed if the plugin is only loaded with PI recordings. However, there have been a few cases where the IMU was missing due to hardware failure. In which case it might be worth leaving these conditions in?


romanroibu:
@N-M-T @papr I've updated the plugin to prevent it from being loaded by default unless:
- Player is running
- Recording was made using Pupil Invisible
- Recording contains IMU data

Also removed the checks that Neil mentioned above. Tested this behaviour with a Pupil Invisible recording and a Pupil Capture recording.

romanroibu (approved):

pupil_src/shared_modules/imu_timeline.py
```diff
@@ -0,0 +1,718 @@
...
+        "accel_x": cygl_utils.RGBA(0.83921, 0.15294, 0.15686, 1.0),
+        "accel_y": cygl_utils.RGBA(0.58039, 0.40392, 0.74117, 1.0),
+        "accel_z": cygl_utils.RGBA(0.54901, 0.33725, 0.29411, 1.0),
+        "pitch": cygl_utils.RGBA(0.12156, 0.46666, 0.70588, 1.0),
+        "roll": cygl_utils.RGBA(1.0, 0.49803, 0.05490, 1.0),
+    }
+    NUMBER_SAMPLES_TIMELINE = 4000
+    TIMELINE_LINE_HEIGHT = 16
+    icon_chr = chr(0xE922)
+    icon_font = "pupil_icons"
```
papr, line 291: This currently corresponds to the `Vis Polyline` plugin icon. I suggest finding an alternative icon.

papr (changes requested):

papr (approved):

PR #2167, opened 2021-07-21
Export-in-progress file-name suffixes

papr:
Exporting video can take a long time. Pupil Player writes the result incrementally to disk. Specifically, it writes the result directly to the final export location. While the export is ongoing, the video file is invalid but still available to the user. This can lead to users opening the video before it has finished exporting and then encountering playback issues.

This PR changes the video exports to include a `.writing` file name suffix during the export. This prevents other applications from recognizing the partially exported video file as complete. Once the export is done, the suffix is removed and other applications are able to recognize the file type as video.

### DIFF
--- a/pupil_src/shared_modules/av_writer.py
+++ b/pupil_src/shared_modules/av_writer.py
@@ -127,7 +127,9 @@ class AV_Writer(abc.ABC):
                 "Using a different container is risky!"
             )
 
-        self.container = av.open(output_file_path, "w")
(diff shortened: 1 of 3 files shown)


N-M-T (approved):
The intended functionality works as expected on Windows 10. Clicking on the .writing file just gives you this: 
image
I'd be surprised if anyone manages to break the export.

romanroibu (approved):

PR #2170, opened 2021-07-27
Recording-specific hotkeys

papr:
Starting with this PR, annotation definitions are stored in `<recording dir>/offline_data/annotation_definitions.json` with the format
```json
{
    "version": 1,
    "definitions": {
        "<label>": "<hotkey>"
    }
}
```
This file can be shared between recordings.

On load, Pupil Player attempts to load the annotation definitions from the recording-specific file. If it is not found or invalid, it will fallback to the session settings, i.e. the last known annotation definitions.

### DIFF
--- a/pupil_src/shared_modules/annotations.py
+++ b/pupil_src/shared_modules/annotations.py
@@ -11,8 +11,11 @@ See COPYING and COPYING.LESSER for license details.
 
 import abc
 import csv
+import json
 import logging
 import os
+import traceback
+import typing as T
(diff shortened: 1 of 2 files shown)


N-M-T (commented):
This works a treat, the functionality is just what I would expect. 
One condition where a user could lose their annotations hotkeys is if Player is closed via the terminal or crashes prior to a normal shutdown or new file load. I'm not sure that is such an issue

PR #2175, opened 2021-07-30
Add Euler angles to headpose tracker result

N-M-T:


pupil_src/shared_modules/imu_timeline.py
```diff
@@ -204,7 +204,7 @@ def update(self, accel, gyro):  # 2-tuples (x, y, z) for accel, gyro
 
         # These are modified to account for Invisible IMU coordinate system and positioning of
         # the IMU within the invisible headset
-        roll = (
+        roll = -(
```
N-M-T, line 207: @papr this change was to bring the IMU roll result in line with the headpose tracker, i.e. a negative IMU roll rotation should correspond to a negative headpose roll rotation, and vice versa for positive roll. 
Since the headpose tracker has been around longer, it made sense to change the IMU direction.

papr: Actually, Cloud has started exporting orientation based on your code already. I think we need to stay consistent with that. Is there a clear definition for what positive or negative roll means? If yes, we should use a reference for all three components calculating orientation. If not, please use the existing orientation components as reference.

N-M-T: The IMU orientation uses the right-hand convention: Positive roll is rightward roll (i.e. right ear lower than left). I have reverted the IMU calc to its original form and modified the headpose estimation to use the same right-hand convention

pupil_src/shared_modules/head_pose_tracker/function/utils.py
```diff
@@ -59,6 +59,35 @@ def convert_matrix_to_extrinsic(extrinsic_matrix):
...
+        z = -np.arctan2(rot[1, 0], rot[0, 0])
+    else:
+        x = np.arctan2(-rot[1, 2], rot[1, 1])
+        y = -np.arctan2(-rot[2, 0], sin_y)
+        z = 0
+
+    return np.rad2deg([x, y, z])
+
+
 def get_camera_pose(camera_extrinsics):
```
papr, line 91:
```suggestion
def get_camera_pose_and_orientation(camera_extrinsics):
```
I suggest reflecting the change in the function name (and its references)

papr (changes requested):

pupil_src/shared_modules/head_pose_tracker/function/utils.py
N-M-T: 'Pose' refers to position and orientation. Therefore, I think adding 'orientation' would be superfluous/incorrect, as Euler is merely another representation of orientation.

papr: ok, in this case, I suggest renaming the `camera_pose` and `camera_orientation` variables to better reflect their content.

N-M-T: I have left camera_pose unchanged, since this contains orientation and translation, but renamed camera_orientation to euler_orientation. I think these variable names are now representative

papr (approved):

PR #2176, opened 2021-07-30
Post hoc HMD gazer

papr:


papr:
Thanks @kbarkevich and @perForm-Lab-RIT for your input on this feature!

PR #2191, opened 2021-09-21
Fix confidence graphs in Pupil Player for pre-2.0 recordings

papr:
Previously, the confidence graphs in Pupil Player had no data for pre-2.0 recordings. This PR fixes this by falling back to the legacy pupil topic.

---
https://app.clickup.com/t/nz7r5x


N-M-T (approved):
The confidence graphs are displayed as expected.

PR #2210, opened 2021-11-18
Raw data exporter: Add option to not export low confidence values

papr:
Pupil Core software assigns "confidence" values to its pupil  detections and gaze estimations. They indicate the quality of the measurement. Until now, the raw data exporter plugin exported all data, independently of its confidence. New users are often not aware of this and wonder why the exported data looks noisier than what they have seen in Pupil Player.

For visualization, Pupil Player hides gaze data below the "Minimum data confidence" threshold (adjustable in the general settings menu). Often, we recommend users to filter their data by a confidence threshold after exporting.

Starting with this PR, users will have the option to only including data  above the "Minimum data confidence" threshold in the export. The option can be enabled in the raw data exporter's menu.

(credits to @N-M-T for raising awareness of this issue)

Screenshot 2021-11-18 at 15 48 00

### DIFF
--- a/pupil_src/shared_modules/raw_data_exporter.py
+++ b/pupil_src/shared_modules/raw_data_exporter.py
@@ -117,7 +117,7 @@ class Raw_Data_Exporter(Plugin):
         should_export_pupil_positions=True,
         should_export_field_info=True,
         should_export_gaze_positions=True,
-        should_include_low_confidence_data=False,
+        should_include_low_confidence_data=True,
     ):
         super().__init__(g_pool)
 
@@ -175,10 +175,9 @@ class Raw_Data_Exporter(Plugin):
             ui.Info_Text(
                 'Pupil Core software assigns "confidence" values to its pupil '
                 "detections and gaze estimations. They indicate the quality of the "
-                "measurement. By default, the raw data export will only include data "
-                'above the "Minimum data confidence" threshold. It can be adjusted '
-                "in the general settings menu. To export all data, turn on the option "
-                "below."
+                "measurement. Disable the option below to only export data above the"
+                '"Minimum data confidence" threshold. It can be adjusted in the '
+                "general settings menu."
             )
         )
         self.menu.append(
--- a/pupil_src/shared_modules/raw_data_exporter.py
+++ b/pupil_src/shared_modules/raw_data_exporter.py
@@ -130,6 +130,14 @@ class Raw_Data_Exporter(Plugin):
         self.should_export_gaze_positions = should_export_gaze_positions
         self.should_include_low_confidence_data = should_include_low_confidence_data
 
+    def get_init_dict(self):
+        return {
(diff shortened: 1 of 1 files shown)


N-M-T (commented):
The export functionality works as expected. However, I think the user should have to 'opt in' to filter out low-confidence data. My reasons are as follows:
1. A lot of existing users are familiar with the raw export containing low-confidence data
2. By defaulting to filter out low-confidence data, we _might_ need to revise tutorials that filter blinks, or at least clarify for the user in each tutorial which option they should have toggled
3. Filtering low-confidence data means there are missing samples in the export. This might lead to confusion for users who are calculating, e.g. eye rotation velocity using the intersample duration, or just when investigating sampling rates in general

papr:
1. Fair enough. But they should also understand what is happening if these are gone. But I get the sentiment, that changing the default "include low conf data = True" is more backwards compatible than the current implementation.
2. That is a very good point
3. eye rotation velocity should not be calculated with low confidence values as they introduce high velocity outliers. The intersample duration can be easily calculated even if you have gaps in the data

N-M-T:
3a. Indeed, users should not calculate eye rotation velocity using low-confidence data points
3b. More generally though, I think having the low-confidence data removed by default would have thrown users off when they were calculating, e.g the average FPS of their gaze data – if they weren't aware of the gaps, `diff(timestamps)` would contain some erroneously large values. In anycase, `include low conf data = True` will prevent this!

pupil_src/shared_modules/raw_data_exporter.py
```diff
@@ -161,6 +171,22 @@ def init_ui(self):
                 "should_export_gaze_positions", self, label="Export Gaze Positions"
             )
         )
+        self.menu.append(
+            ui.Info_Text(
+                'Pupil Core software assigns "confidence" values to its pupil '
+                "detections and gaze estimations. They indicate the quality of the "
+                "measurement. Disable the option below to only export data above the"
+                '"Minimum data confidence" threshold. It can be adjusted in the '
```
N-M-T, line 179:
```suggestion
                '"Minimum data confidence" threshold. This threshold can be adjusted in the '
```

N-M-T (changes requested):

papr:
3b. Agreed!

peteratBHVI:
> 3a. Indeed, users should not calculate eye rotation velocity using low-confidence data points 3b. More generally though, I think having the low-confidence data removed by default would have thrown users off when they were calculating, e.g the average FPS of their gaze data – if they weren't aware of the gaps, `diff(timestamps)` would contain some erroneously large values. In anycase, `include low conf data = True` will prevent this!

Good points, 
also the amount of low confidence data is a good quality measure. e.g., indicate if the eye cameras need to be in a better position for particular visual tasks. I like the option to be on.

N-M-T (approved):

PR #2220, opened 2022-01-20
avoid crashing when no frame are present if displaying undistorted frame

bpinsard:


papr (approved):
Great catch! Thank you for fixing this issue!

PR #2284, opened 2022-11-17
Refactor bundling code

papr:


papr:
```py
[573396] Error loading Python lib '/opt/pupil_capture/libpython3.11.so.1.0': dlopen: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.35' not found (required by /opt/pupil_capture/libpython3.11.so.1.0)
```
on Ubuntu 20.04.5 LTS

PR #2305, opened 2023-02-02
Expose canny threshold parameter in the 2D detector menu

papr:


pupil_src/shared_modules/pupil_detector_plugins/detector_2d_plugin.py
```diff
@@ -138,6 +138,22 @@ def init_ui(self):
                 step=1,
             )
         )
+        info = ui.Info_Text(
+            "When using Neon in IR-light-heavy environments, it can help to increase "
+            "the Canny Threshold to reduce the effect of reflections on the pupil "
+            "detection result. The default value is 160."
```
N-M-T, line 144:
```suggestion
            "When using Neon in bright light, increasing the Canny Threshold can "
            "help reduce the effect of reflections in the eye image and improve pupil "
            "detection. The default value is 160.
```

N-M-T (commented):
I think the initial text was fine. Nevertheless, I added an alternative. No preference which gets used!

Issue #2312, opened 2023-02-16
glfwGetClipboardString blocks, causing a process to hang until there is mouse movement within the corresponding window


papr:
**Affected plattforms**
- Linux, bundle only

**Reproducibility**
Only on some computers, and only happens after the window has been inactive/not focused for a while.

**Resulting issues**
- The blocking process will drop data until the user unblocks it by moving the mouse in the blocked window

**Workarounds**
- Run from source

bpinsard:
You mentioned that running from source was a workaround, but we had that same issue while running from source too. 
See #2334 for a potential fix.

PR #2334, opened 2023-09-19
move clipboard updates to a focus callback, perfs enh + stability fix

bpinsard:
We had a curious issue for years where eye or world window would freeze on our production setup only. However moving the mouse over the frozen window would usually unfreeze it. The main issue was data loss, because the software was not processing camera frames and the network buffers would overflow.

After inspecting the stack when that freezing occurs, it seems that calls to `glfw.get_clipboard_string` sometimes hangs.
This might be specific interactions with our software / hardware, or that in our config (FPS/ screen framerate) the clipboard was concurrently written/read by eye and world windows, causing the hangs.

In the meantime, profiling would show that a small but non-negligible portion of the CPU cycles would be spent querying the clipboard.

I here moved that call to the focus callback: the logic being, if the OS clipboard (not the pyglui one that seems separate IIUC)  was modified, it is only happening if the focus was given to another window, so we can retrieve it only when the pupil window get focus again. 
That way, the clipboard is not probed at the screen framerate and likelyhood of hangs is largely dimished. It also separates the processing loop from call to OS clipboard that seems likely to disrupt RT reqs of pupil.  

Let me know if that makes any sense and if you want me to make changes.

### DIFF
--- a/pupil_src/launchables/eye.py
+++ b/pupil_src/launchables/eye.py
@@ -12,8 +12,8 @@ import os
 import platform
 import signal
 import time
-from types import SimpleNamespace
 from functools import partial
+from types import SimpleNamespace
 
 
 class Is_Alive_Manager:
--- a/pupil_src/launchables/eye.py
+++ b/pupil_src/launchables/eye.py
@@ -13,6 +13,7 @@ import platform
 import signal
 import time
 from types import SimpleNamespace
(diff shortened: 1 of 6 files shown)


bpinsard:
Just realized that it had been documented in #2312 recently. 
So this PR should fix #2312

Issue #2337, opened 2023-11-14
camera intrinsics error


cpicanco:
I am getting the follwing error when pressing I for camera intrinsics:

```txt
world - [ERROR] launchables.world: Process Capture crashed with trace:
Traceback (most recent call last):
  File "launchables\world.py", line 755, in world
  File "camera_intrinsics_estimation.py", line 331, in recent_events
cv2.error: OpenCV(3.4.13) C:\Users\appveyor\AppData\Local\Temp\1\pip-req-build-cg3xbgmk\opencv\modules\core\src\matmul.dispatch.cpp:439: error: (-215:Assertion failed) scn == m.cols || scn + 1 == m.cols in function 'cv::transform'
```

nky001:
Verify the number of channels in the input matrices because the error seems to indicate there might be a mismatch between channels.

cpicanco:
Hi @nky001 , do you know if this type of mismatch is supposed to occur in the released bundle for non pupil cameras?

cpicanco:
So, I found the problem. It is related to bad detection of the pattern.
![image](https://github.com/pupil-labs/pupil/assets/6511849/8fff2c22-958d-48c2-8e33-4483fae1376d)

@papr , I don't see a need to throw an exception here. Please, could you kindly confirm it?
https://github.com/pupil-labs/pupil/blob/master/pupil_src/shared_modules/camera_intrinsics_estimation.py#L327-L335

What about a warning "The pattern couldn't be found. You may adjust world video source settings and try again."

cpicanco:
Take a look here:
https://answers.opencv.org/question/215356/findcirclesgrid-throws-exception/


> Mystery resolved: opencv throws and catches an exception internally. Enabling the flag <All C++ Exceptions not in this list> (in "C++ Exceptions" in Visual Studio 2019) makes VS catch it even though it's within try-catch block. I don't know if it's by design, but changing that flag (it is disabled by default) seems to be cause and the solution of the problem.

Not sure if it may apply to Python try except too.

cpicanco:
So, I think there is no need to raise an expection. For example:
```
            try:
                status, grid_points = cv2.findCirclesGrid(
                    img, (4, 11), flags=cv2.CALIB_CB_ASYMMETRIC_GRID
                )
            except cv2.error:
                return
```

should do the work. As Pupil now do not have any plans to fix this, I am sharing a plugin with the proposed fix:

```python
import cv2
import gl_utils
import glfw
[...]

PR #2340, opened 2024-02-03
avoid crash for bad intrinsics sampling

cpicanco:
I don't see a real need to raise an exception and crashing the application in production here. Please, make me known if I am wrong. Note that this error may also happen for Pupil cameras, not only for third party cameras as originally documented by me.


cpicanco:
#2337

